PR #7, opened 2024-03-07
implements inputs 

kmer2016:


ui-core/src/styles/inputs/input.css
```diff
@@ -0,0 +1,329 @@
...
+}
+
+@keyframes spin {
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
+}
```
Math-R, line 329: need to fix eof

ui-core/tailwind.config.js
```diff
@@ -89,4 +101,4 @@ module.exports = {
     }
   },
   plugins: [],
-};
+};
```
Math-R, line 104: same for eof

Math-R (changes requested):

PR #11, opened 2024-03-20
review project architecture

sim51:
- monorepo project with sub-packages which allow us to share dep, scripts, eslint, ts config ...
- rollup configuration (remove commonjs, babel, and adding postcss)
- package.json with export
- remove CRA trace
- adding a storybook package that read stories across project
- adding lerna for release


sim51:
merged in `dev`

PR #42, opened 2024-04-17
UI core: radiobuttons

achrafmohye:
- Adds a new radio button component.
- Implements a radio button group to manage multiple radio buttons.


ui-core/src/components/inputs/RadioButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export interface RadioButtonProps {
+  label: string;
+  value: string;
+  hint?: string;
+  checked: boolean;
+  disabled?: boolean;
+  readonly?: boolean;
+  small?: boolean;
+  onChange: (value: string) => void;
```
kmer2016: onChange don't need to be mandatory

kmer2016:
Some changes : 
- **Replaced `useKeyPress` with `useTabFocus` Hook**: This change allows for better handling of focus when using the Tab key.
- **Fixed event handling in `RadioButton` component**: Previously, the events bound to the HTML input within the RadioButton component were not being triggered because the element was obscured by a span (the checkmark). I resolved this by adjusting the z-index, placing the input above the span. This fix now allows events to be triggered as expected. Consequently, I removed the onClick event handler from the parent div wrapping the radio input, and deleted the custom onChange prop override.
- **Updated radioButton Storybook layout**: I modified the Storybook design to align with the provided sketch.
- **Code Cleanup**: I standardized the implementation of the `InputStatusIcon` component to use a single consistent approach.

PR #45, opened 2024-04-18
ui-core: add ui-icons to ui-core

kmer2016:


kmer2016:
taken into account in #51

PR #51, opened 2024-04-25
Cnh/add tailwind css

### LINKED ISSUE #50: Centralize Common Tailwind CSS Styling Parameters for All Projects in the Monorepo
- Created a tailwind.preset.js file at the root of the monorepo.
- Updated ui-core: imported the preset in tailwind.config.js.

kmer2016:
closes #50 
Each project need to import the theme.css from ui-core and tailwind-preset.js from the monorepo root.


ui-core/package.json
```diff
@@ -34,7 +34,9 @@
     "prepublishOnly": "npm run clean && npm run build"
   },
   "peerDependencies": {
-    "react": ">=18.0"
+    "react": ">=18.0",
+    "@osrd-project/ui-core": ">=0.0.18",
```
Math-R: self dependencies ?

sim51:
IMHO making a sub-project that export the tailwind config is a little overkill.
Why not just put it at the root of the project ? This file has no deps and is pretty much a simple json.

An other point, the `font.css` is not a real CSS, the `inline` function doesn't exist in CSS, it's a tailwind plugin function (from `postcss-assets`). For me it's weird to have a lib that export a CSS, which can't be used like that.

sim51 (approved):

PR #53, opened 2024-04-26
gh-pages: make it work

nicolaswurtz:


emersion:
This needs more work and will not get updated because Nicolas has moved on to greener pastures. Let's track this in https://github.com/OpenRailAssociation/osrd-ui/issues/1037.

PR #56, opened 2024-04-26
ui-icons: adding story

sim51:
Adding a story for storybook that allows you to search an icon and check its render.

image


sim51:
FYI, the error boundary is here because some icons are missing (depending of the size and the variant), and throw an error. The error boundary is here to avoid to have an error page on storybook.

PR #66, opened 2024-05-21
core: add tolerancePicker component and styling

### LINKED ISSUE #284: ui-core: add Tolerance Picker component
Add a component to pick tolerances at a waypoint

achrafmohye:
closes https://github.com/OpenRailAssociation/osrd-ui/issues/284


ui-core/src/stories/TolerancePicker.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import '@osrd-project/ui-core/dist/theme.css';
+
+const meta: Meta<typeof TolerancePicker> = {
+  component: TolerancePicker,
+  args: {
+    label: 'TolerancePicker',
+    id: 'time-picker',
+  },
+  title: 'TolerancePicker',
```
clarani:
```suggestion
  title: 'core/TolerancePicker',
```

ui-core/src/components/inputs/TolerancePicker.tsx
```diff
@@ -0,0 +1,97 @@
+import React, { useState, useRef } from 'react';
+import Input, { InputProps } from './Input';
+import Modal from '../Modal';
+
+const TolerancePicker: React.FC<InputProps> = (inputProps) => {
+  const [showPicker, setShowPicker] = useState(false);
+  const [minusTolerance, setMinusTolerance] = useState('00');
+  const [plusTolerance, setPlusTolerance] = useState('00');
+  const [inputValue, setInputValue] = useState('');
```
clarani: The inputValue and setInputValue should come from the parent component since this component is only a ui-component (no data manipulation here)

ui-core/src/components/inputs/TolerancePicker.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const handleTimeChange = (newMinusTolerance: string, newPlusTolerance: string) => {
+    setMinusTolerance(newMinusTolerance);
+    setPlusTolerance(newPlusTolerance);
+    const formattedValue = `-${newMinusTolerance}/+${newPlusTolerance}`;
+    setInputValue(formattedValue);
```
clarani: The parent component should give a onChange function, which should be call here. It should take a MinusTolerance and a PlusTolerance which should be both in seconds:
```
onChange: (minusTolerance: number, plusTolerance: number ) => void
```

clarani (changes requested):

ui-core/src/components/inputs/TolerancePicker.tsx
```diff
@@ -0,0 +1,115 @@
...
+      <Modal inputRef={inputRef} isOpen={showPicker} onClose={() => setShowPicker(false)}>
+        <div className="time-tolerance">
+          <div className="time-picker-section">
+            <div className="tolerance-grid">
+              {toleranceRanges.map((tolerance, index) => {
+                const { label, value } = tolerance;
+                return (
+                  <button
+                    key={`${label}-${index}`}
+                    className={`minute ${minusTolerance === value ? 'selected' : ''}`}
```
clarani: Can you use cx here ? :) Same line 97

ui-core/src/components/inputs/TolerancePicker.tsx
```diff
@@ -0,0 +1,115 @@
...
+import Modal from '../Modal';
+
+export type ToleranceValues = {
+  minusTolerance: number;
+  plusTolerance: number;
+};
+
+type ToleranceRange = { label: string; value: number };
+
+export type TolerancePickerProps = InputProps & {
```
clarani: Do you think we can Omit 'value' in InputProps ? Since we handle the value with toleranceValues
```suggestion
export type TolerancePickerProps = Omit<InputProps, 'value'> & {
```

ui-core/src/components/inputs/TolerancePicker.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const plusToleranceIndex = toleranceRanges.findIndex((range) => range.value === plusTolerance);
+    return { minusToleranceIndex, plusToleranceIndex };
+  }, [minusTolerance, plusTolerance, toleranceRanges]);
+
+  const handleTimeChange = (
+    { label: newMinusLabel, value: newMinusValue }: ToleranceRange,
+    { label: newPlusLabel, value: newPlusValue }: ToleranceRange
+  ) => {
+    const formattedValue = `-${newMinusLabel}/+${newPlusLabel}`;
+    setInputValue(formattedValue);
```
clarani: We should update the inputValue when the toleranceValues changes. Can you add a useEffect which update the inputValue each time the toleranceValues changes ?
```suggestion
```

ui-core/src/components/inputs/TolerancePicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+    <div className="tolerance-picker">
+      <div>
+        <Input
+          {...inputProps}
+          value={inputValue}
+          // Event not used for now
+          onChange={(e) => {
+            setInputValue(e.target.value);
+            setShowPicker(true);
+          }}
```
clarani: You can removed onChange for now
```suggestion
```

ui-core/src/components/inputs/TolerancePicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const { minusToleranceIndex, plusToleranceIndex } = useMemo(() => {
+    const minusToleranceIndex = toleranceRanges.findIndex(
+      (range) => range.value === minusTolerance
+    );
+    const plusToleranceIndex = toleranceRanges.findIndex((range) => range.value === plusTolerance);
+    if (minusToleranceIndex < 0 || plusToleranceIndex < 0)
+      setWarningStatus({
+        status: 'warning',
+        message: "One of the values isn't included in the ranges provided.",
+      });
```
clarani: You never remove the warning if we select valid values

ui-core/src/components/inputs/TolerancePicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  const { minusToleranceIndex, plusToleranceIndex } = useMemo(() => {
+    const minusToleranceIndex = toleranceRanges.findIndex(
+      (range) => range.value === minusTolerance
+    );
+    const plusToleranceIndex = toleranceRanges.findIndex((range) => range.value === plusTolerance);
+    if (minusToleranceIndex < 0 || plusToleranceIndex < 0)
+      setWarningStatus({
+        status: 'warning',
+        message: "One of the values isn't included in the ranges provided.",
```
clarani: Maybe this message is more explicit ?
```suggestion
        `${minusToleranceIndex === -1 ? minusTolerance : plusTolerance} is not a valid tolerance value.`
```

ui-core/src/components/inputs/TolerancePicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+              {toleranceRanges.map((tolerance, index) => {
+                const { label, value } = tolerance;
+                return (
+                  <button
+                    key={`${label}-${index}`}
+                    className={cx('minute', { selected: plusTolerance === value })}
+                    style={{ gridArea: `a${index + 1}` }}
+                    onClick={() =>
+                      onToleranceChange({
+                        minusTolerance: toleranceRanges[minusToleranceIndex].value,
```
clarani: You can keep the value from the input
```suggestion
                        minusTolerance,
```

ui-core/src/components/inputs/TolerancePicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+                const { label, value } = tolerance;
+                return (
+                  <button
+                    key={`${label}-${index}`}
+                    className={cx('minute', { selected: minusTolerance === value })}
+                    style={{ gridArea: `a${index + 1}` }}
+                    onClick={() =>
+                      onToleranceChange({
+                        minusTolerance: value,
+                        plusTolerance: toleranceRanges[plusToleranceIndex].value,
```
clarani: Same here
```suggestion
                        plusTolerance,
```

ui-core/src/components/inputs/TolerancePicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+}) => {
+  const formatToleranceValue = (minusIndex: number, plusIndex: number) =>
+    `-${toleranceRanges[minusIndex]?.label || ''}/+${toleranceRanges[plusIndex]?.label || ''}`;
+
+  const [showPicker, setShowPicker] = useState(false);
+  const [inputValue, setInputValue] = useState(formatToleranceValue(0, 0));
+  const [warningStatus, setWarningStatus] = useState<StatusWithMessage | undefined>(undefined);
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const { minusToleranceIndex, plusToleranceIndex } = useMemo(() => {
```
clarani: Actually, you don't need these index. You can move the code in the useEffect below with the setInputValue (for the warningStatus) and remove the useMemo (also add toleranceRanges in the dependencies)

ui-core/src/components/inputs/TolerancePicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+export type TolerancePickerProps = Omit<InputProps, 'value'> & {
+  onToleranceChange: (toleranceValues: ToleranceValues) => void;
+  toleranceRanges: ToleranceRange[];
+  toleranceValues?: ToleranceValues;
+};
+
+const TolerancePicker: React.FC<TolerancePickerProps> = ({
+  onToleranceChange,
+  toleranceRanges,
```
clarani: I think it would be better to completely remove the toleranceRanges from the props (as we have discussed it, it will break the css etc).
```suggestion
```
You can move the toleranceRanges const from the story in the main file TolerancePicker. You can also rename it with screaming case TOLERANCE_RANGES

ui-core/src/components/inputs/TolerancePicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+          ref={inputRef}
+        />
+      </div>
+
+      <Modal inputRef={inputRef} isOpen={showPicker} onClose={() => setShowPicker(false)}>
+        <div className="time-tolerance">
+          <div className="time-picker-section">
+            <div className="tolerance-grid">
+              {toleranceRanges.map((tolerance, index) => {
+                const { label, value } = tolerance;
```
clarani: nit: You can even destructure the tolerance (but I'm not sure that it will improve the readibility though, as you wish)
```suggestion
              {toleranceRanges.map(({ label, value } , index) => (
```

ui-core/src/stories/TolerancePicker.stories.tsx
clarani, line 1: Can you add a story with invalid values at the first render ? :)

ui-core/src/styles/main.css
clarani, line 1: If you rebase, I think you might need to update this file

ui-core/src/hooks/useModalPosition.tsx
clarani, line 1: I think that this hook has been already added in another folder, can you check after your rebase ?

ui-core/src/components/inputs/TolerancePicker.tsx
SharglutDev: Shouldn't this be in a i18n key for trad ?

ui-core/src/components/inputs/TolerancePicker.tsx
```diff
@@ -0,0 +1,145 @@
...
+                        plusTolerance,
+                      })
+                    }
+                  >
+                    {`-${label}`}
+                  </button>
+                );
+              })}
+            </div>
+          </div>
```
kmer2016: Here’s a suggestion: since these lines are used to render a range button, we can extract this into a component. This way, we can reuse it for both minus and plus ranges. What do you think?

ui-core/src/styles/main.css
```diff
@@ -2,9 +2,13 @@
 @import 'tailwindcss/components';
 @import 'tailwindcss/utilities';
 @import './animations.css';
+@import './button.css';
 @import './inputs/index.css';
+@import './inputs/radioButton.css';
+@import './inputs/textArea.css';
+@import './inputs/timePicker.css';
+@import './inputs/tolerancePicker.css';
```
clarani: You can move this in `styles/inputs/index.css`

clarani (approved):
LGTM & tested :heavy_check_mark: 

Very nice job :) I only left 1 last comment, you can resolve it youself

ui-core/src/components/inputs/TolerancePicker.tsx
clarani: Ah I don't know how it was done in the speedSpaceChart. Here it is ok for me. We have a props function which takes the invalid tolerance and returns the translated string.
In osrd, we can give this function as a prop:
```
translateWarningMessage={(invalidTolerance: number) => t('toleranceIsInvalid', { invalidTolerance })}
```
where the translation 'toleranceIsInvalid' is translated in our json files, what do you think ? 

(invalidTolerance maybe should be a number and not a string)

ui-core/src/styles/inputs/tolerancePicker.css
```diff
@@ -0,0 +1,152 @@
...
+    box-shadow:
+      0px 15px 12px -7px rgba(0, 0, 0, 0.04),
+      0px 2px 6px -3px rgba(0, 0, 0, 0.26),
+      0px 1px 1px rgba(0, 0, 0, 0.12),
+      0px 4px 10px -4px rgba(60, 138, 255, 0.5),
+      0px 8px 16px -6px rgba(114, 168, 247, 0.29);
+  }
+
+  input {
+    cursor: pointer;
```
kmer2016, line 150:
```suggestion
    cursor: pointer;
    caret-color: transparent;
```

ui-core/src/components/inputs/tolerancePicker/TolerancePicker.tsx
```diff
@@ -0,0 +1,113 @@
...
+  }, [minusTolerance, plusTolerance]);
+
+  return (
+    <div className="tolerance-picker">
+      <div>
+        <Input
+          {...inputProps}
+          value={inputValue}
+          statusWithMessage={warningStatus}
+          readOnly
```
kmer2016:
```suggestion
```

kmer2016 (commented):
I proposed an improvement to make it clearer that the component is not editable using the input. Using the readOnly attribute on the input can be confusing and might give the impression that the component is disabled.

kmer2016 (approved):
LGTM and tested ! Good job

PR #68, opened 2024-05-23
ui-spacetimechart: implements various features

jacomyal:
The related features from ticket [#502](https://github.com/osrd-project/osrd-confidential/issues/502) are the following:
- Inverser les axes du graphe (échanger les axes X et Y)
- Ajouter un snap aux PRs du sillon à l’outil de mesure
- Pouvoir masquer le quadrillage
- Différencier l’espace occupé par les voies uniques sur toute la largeur du graphe
- Afficher les temps de fermeture de postes
- Disposer d’une image PNG de la vue courante
- Permettre d’accéder à différents thèmes de couleur (fond, axes, lignes…)
- Garder un sillon sélectionné dans la vue lorsqu’on se déplace dans le temps à l’aide de la molette

I kept quite atomic commits, to make them easier to review.


ui-spacetimechart/src/stories/options.stories.tsx
```diff
@@ -0,0 +1,197 @@
...
+      style={{
+        position: 'absolute',
+        right: 10,
+        bottom: 10,
+      }}
+    >
+      <button
+        onClick={() =>
+          captureCanvases().then((blob) => FileSaver.saveAs(blob, 'space-time-chart.png'))
+        }
```
Yohh, line 38: is there a reason to use `FileSaver` instead of the canvas method `toDataURL()` ?

ui-spacetimechart/src/utils/canvas.ts
```diff
@@ -75,11 +77,96 @@ export function drawAliasedLine(
...
+ * getAliasedDiscShape(1);
+ * // [0, 1, 0,
+ * //  1, 1, 1,
+ * //  0, 1, 0]
+ * getAliasedDiscShape(2);
+ * // [0, 0, 1, 0, 0,
+ * //  0, 1, 1, 1, 0,
+ * //  1, 1, 1, 1, 1,
+ * //  0, 1, 1, 1, 0,
+ * //  0, 0, 1, 0, 0]
```
Yohh:
```suggestion
 * getAliasedDiscShape(0);
 * [1]
 * getAliasedDiscShape(1);
 * [0, 1, 0,
 *  1, 1, 1,
 *  0, 1, 0]
 * getAliasedDiscShape(2);
 * [0, 0, 1, 0, 0,
 *  0, 1, 1, 1, 0,
 *  1, 1, 1, 1, 1,
 *  0, 1, 1, 1, 0,
 *  0, 0, 1, 0, 0]
```

ui-spacetimechart/src/stories/options.stories.tsx
jacomyal: Using `toDataURL` allows opening the image in a new tab, but it's not enough to download the image as a local file (which was what I was aiming at).

And historically, saving a local file from a browser is a bit tricky (using an hidden `a` tag in the DOM, and differently in some browsers). So, I used the most used JS lib to handle that, [file-saver](https://www.npmjs.com/package/file-saver).

Yohh: ok that was what I expected to see (add an hidden `a`)
thanks

ui-spacetimechart/src/lib/types.ts
```diff
@@ -126,6 +133,37 @@ export type MouseState = {
...
+export type SpaceTimeChartTheme = {
+  background: string;
+  breakpoints: number[];
+  timeRanges: number[];
+  pathsStyles: {
+    font: string;
+  };
+  spaceGraduationsStyles: Record<
+    number,
+    { width: number; color: string; opacity?: number; dashArray?: number[] }
```
Math-R: Will spaceGradutationsStyles and timeGraduationsStyles  Records always be similar ?
If so we can encapsulate those properties in some interface ?

ui-spacetimechart/src/stories/options.stories.tsx
jacomyal: If you prefer that we cut the dependency, it's possible. But it's quite a light one, and it relieves us from maintaining some tricky (short, but still tricky) code that handles something `file-saver` does better.

Yohh: no it's fine to me, I was just asking for "personal culture"

jacomyal: ok :)

ui-spacetimechart/src/lib/types.ts
jacomyal: Good point. I don't see what could change - and if it does, we can handle it then.

ui-spacetimechart/src/utils/snapping.ts
```diff
@@ -0,0 +1,58 @@
+import { type HoveredItem, type Point } from '../lib/types';
+
+/**
+ * This function takes a point and a segment, and returns the point of the segment that is the
+ * closest to the input point.
+ */
+export function getClosestPointOnSegment(
```
Math-R, line 7: I see there is no tolerance setting in your snap function, but you are using disks to perform the snap. 
But in some cases the cursor will move from a point on the curve to the end of the segment. And we cannot therefore visualize the data of the curves between this point and the end of the segment.

Yohh: the context:
[Screencast from 2024-08-29 15-47-10.webm](https://github.com/user-attachments/assets/5e13f7a0-ba55-4591-9623-090847a134b7)

Yohh: [Screencast from 2024-08-29 15-47-10.webm](https://github.com/user-attachments/assets/c7cee236-85f6-4b7e-95cf-fbf224a0c859)

jacomyal: I get your point. Indeed, when the mouse is too close to a point, the mouse will never snap to the curve. I'm not sure it's an issue, though, since points might be more important than curves (in which case, they must be easier to catch).

What do you think about this issue @thibautsailly?

thibautsailly: The maximum zoom level displays 1 minute over 96px.
The snapping engages when the cursor is at 12~13px from a point, or 7,8 seconds at this zoom level.
I don't think missing contextual informations for this amount of time could be a problem, but I'll ask engineers on Monday if it's acceptable or not from their point of view.

I'm anticipating that they will be interested with snapping at path intersections as well as points.

ui-spacetimechart/src/hooks/useMouseInteractions.ts
```diff
@@ -184,7 +125,5 @@ export function useMouse(
         initialData: panningState.initialData,
       });
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [position]);
-
-  return mouseState;
+  }, [position.x, position.y]);
```
clarani, line 140: Do you have a special reason to use these dependencies instead of `[position]` ?

ui-spacetimechart/src/hooks/useMouseInteractions.ts
```diff
@@ -173,7 +105,16 @@ export function useMouse(
 
   // Listen to "move" updates:
   useEffect(() => {
-    const { onPan } = handlersRef.current;
+    const { onPan, onMouseMove } = handlersRef.current;
+
+    if (onMouseMove) {
+      onMouseMove({
+        position,
+        isHover: isHover,
```
clarani:
```suggestion
        isHover,
```

ui-spacetimechart/src/hooks/useMouseTracking.ts
```diff
@@ -0,0 +1,73 @@
...
+        isHover,
+      }));
+    },
+    [dom]
+  );
+
+  // Bind event handlers:
+  useEffect(() => {
+    if (!dom) return;
+    dom.addEventListener('mousedown', downHandler);
```
clarani, line 54: Why do you use the dom here and document below ?

ui-spacetimechart/src/stories/utils.ts
```diff
@@ -40,3 +40,28 @@ export function zoom(
     yOffset: y - ((y - state.yOffset) / state.yZoomLevel) * yZoomLevel,
   };
 }
+
+/**
+ * This function helps to pretty-print a Date as a length:
```
clarani:
```suggestion
 * This function helps to pretty-print a Date as a length.
```

ui-spacetimechart/src/stories/options.stories.tsx
```diff
@@ -15,16 +16,18 @@ import {
 import { SpaceTimeChart, PathLayer } from '../';
 import { type Point } from '../lib/types';
 import { getDiff } from '../utils/vectors';
-
 import './tailwind-mockup.css';
 
 /**
- * This story aims at showcasing how to swap time and space axis in a SpaceTimeChart.
+ * This story aims at showcasing how to handle panning and zooming in a SpaceTimeChart.
```
clarani: This description is not correct I think

ui-spacetimechart/src/stories/components.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+import { formatTimeLength } from './utils';
+import { WHITE_75 } from '../lib/consts';
+import { MouseContext, SpaceTimeChartContext } from '../lib/context';
+import { type DataPoint, type Point } from '../lib/types';
+
+/**
+ * This component draws a dashed line from p1 to p2, using SVG:
+ */
+export const Line: FC<{
```
clarani: No need to export it for the moment

ui-spacetimechart/src/stories/components.tsx
```diff
@@ -0,0 +1,112 @@
...
+    <div
+      style={{
+        position: 'absolute',
+        width: 1,
+        height: CROSS_SIZE,
+        left: -0.5,
+        top: -CROSS_SIZE / 2,
+        background: 'black',
+      }}
+    />
```
clarani: nit: you can even create a small component `Cross` with these 2 divs

ui-spacetimechart/src/stories/components.tsx
```diff
@@ -0,0 +1,112 @@
...
+      strokeDasharray="5, 5"
+    />
+  </svg>
+);
+
+/**
+ * This component renders a data label, to help bring context to the SpaceTimeChart:
+ */
+const CROSS_SIZE = 11;
+export const DataLabel: FC<{
```
clarani: nit: no need to export DataLabel for the moment

ui-spacetimechart/src/stories/additional-data.stories.tsx
```diff
@@ -0,0 +1,291 @@
...
+          to: OPERATIONAL_POINTS[i + 1].position,
+          ...(spaceScaleType === 'linear'
+            ? { size: 50 * state.yZoomLevel }
+            : { coefficient: 150 / state.yZoomLevel }),
+        }))}
+        timeOrigin={+new Date('2024/04/02')}
+        timeScale={60000 / state.xZoomLevel}
+        xOffset={state.xOffset}
+        yOffset={state.yOffset}
+        onPan={({ initialPosition, position, isPanning }) => {
```
clarani, line 201: Maybe it would be nice to extract this function in a helper since it is used in many stories

ui-spacetimechart/src/utils/png.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (!layers.length) throw Error('There must be at least one layer to capture.');
+
+  const firstCanvas = canvases[layers[0]];
+  const width = firstCanvas.width;
+  const height = firstCanvas.height;
+
+  // Create a new canvas, on which the different layers will be drawn:
+  const canvas = document.createElement('CANVAS') as HTMLCanvasElement;
+  canvas.setAttribute('width', width + '');
+  canvas.setAttribute('height', height + '');
```
clarani, line 19: Use real casts 
```suggestion
  canvas.setAttribute('width', String(width));
  canvas.setAttribute('height', String(height));
```

ui-spacetimechart/src/components/TimeCaptions.tsx
```diff
@@ -88,9 +62,9 @@ const TimeCaptions: FC = () => {
       const pixelsPerMinute = (1 / timeScale) * MINUTE;
       let labelLevels: number[] = [];
 
-      BREAKPOINTS.some((breakpoint, i) => {
+      breakpoints.some((breakpoint, i) => {
         if (pixelsPerMinute < breakpoint) {
-          labelLevels = LABELS_PRIORITIES[i];
+          labelLevels = timeCaptionsPriorities[i];
```
clarani, line 67: Should we add a test here if the user doesn't provide a timeCpationsPriorities long enough ?

ui-spacetimechart/src/components/TimeCaptions.tsx
```diff
@@ -112,16 +86,16 @@ const TimeCaptions: FC = () => {
...
         ctx.fillRect(0, spaceAxisSize - CAPTION_SIZE, timeAxisSize, CAPTION_SIZE);
       } else {
         ctx.fillRect(0, 0, CAPTION_SIZE, timeAxisSize);
       }
 
       // Render caption top border:
-      ctx.strokeStyle = '#979797';
-      ctx.lineWidth = 2;
+      ctx.strokeStyle = timeGraduationsStyles[1].color;
+      ctx.lineWidth = timeGraduationsStyles[1].width;
```
clarani, line 98: What if `timeGraduationsStyles.length === 1` ? Should we throw an error or use the default theme ?

ui-spacetimechart/src/components/TimeGraduations.tsx
```diff
@@ -1,63 +1,10 @@
...
-  6: {
-    width: 1,
-    color: BLUE,
-    opacity: 0.5,
-    dashArray: [1, 12],
-  },
-};
-
-const TimeGraduations: FC = () => {
+export const TimeGraduations: FC = () => {
```
clarani: Remove the export since you already default export it at the end of the file

ui-spacetimechart/src/components/PathLayer.tsx
```diff
@@ -237,18 +245,20 @@ export const PathLayer: FC<PathLayerProps> = ({
...
       ctx.textAlign = 'start';
 
       const dx = 5;
       const dy = angle >= 0 ? -5 : 15;
       const padding = 2;
       const measure = ctx.measureText(label);
       const w = measure.width + 2 * padding;
-      const h = FONT_SIZE + 2 * padding;
-      ctx.fillStyle = WHITE_75;
+      const h = measure.actualBoundingBoxAscent + measure.actualBoundingBoxDescent + 2 * padding;
```
clarani: Can you add a intermediate const here ? I'm not used to see `measure.actualBoundingBoxAscent` and `measure.actualBoundingBoxDescent`
```suggestion
      const fontSize = measure.actualBoundingBoxAscent + measure.actualBoundingBoxDescent;
      const h = fontSize + 2 * padding;
```

ui-spacetimechart/src/components/PathLayer.tsx
```diff
@@ -237,18 +245,20 @@ export const PathLayer: FC<PathLayerProps> = ({
...
+      ctx.globalAlpha = 0.75;
+      ctx.fillStyle = background;
       ctx.fillRect(dx - padding, dy - h + padding, w, h);
       ctx.fillStyle = labelColor;
+      ctx.globalAlpha = 1;
       ctx.fillText(label, dx, dy - FONT_SIZE / 4);
```
clarani: Shouldn't we use the fontSize computed above here ? (`measure.actualBoundingBoxAscent + measure.actualBoundingBoxDescent`)

ui-spacetimechart/src/utils/geometry.ts
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * This function helps to find if two segments do intersect.
+ */
+export function intersectsSegments(s1From: Point, s1To: Point, s2From: Point, s2To: Point) {
+  const crossProduct = (a: Point, b: Point) => a.x * b.y - a.y * b.x;
+  const subtract = (a: Point, b: Point) => ({ x: a.x - b.x, y: a.y - b.y });
+
+  const d1 = subtract(s1To, s1From);
+  const d2 = subtract(s2To, s2From);
+  const delta = subtract(s2From, s1From);
```
clarani: You can compute the delta after the if l17

ui-spacetimechart/src/utils/geometry.ts
```diff
@@ -0,0 +1,64 @@
+import { inRange } from 'lodash';
+
+import { type Point } from '../lib/types';
+
+/**
+ * This function helps to find if two segments do intersect.
+ */
+export function intersectsSegments(s1From: Point, s1To: Point, s2From: Point, s2To: Point) {
+  const crossProduct = (a: Point, b: Point) => a.x * b.y - a.y * b.x;
+  const subtract = (a: Point, b: Point) => ({ x: a.x - b.x, y: a.y - b.y });
```
clarani: The function `getDiff` in `utils/vector.ts` does the same thing than subtract

ui-spacetimechart/src/components/PathLayer.tsx
```diff
@@ -89,7 +89,6 @@ export type PathLayerProps = {
  * - The "picking" shape (to handle interactions)
  */
 export const PathLayer: FC<PathLayerProps> = ({
-  index,
```
clarani, line 92: Removing this will break Manchette, can you add a commit to fix it ?

ui-spacetimechart/src/stories/scroll-navigation.stories.tsx
```diff
@@ -0,0 +1,225 @@
...
+                const path = PATHS_DICT[s.selected];
+                const xyPoints = path.points.map(getPoint);
+
+                if (isPathOnScreen(width, height, xyPoints)) {
+                  // Find if, for the time in the middle of the screen, the path is lower or higher:
+                  const centerTime = getTime(width / 2);
+                  const pathPositionAtTime = getSpaceAtTime(path, centerTime);
+                  const pathYAtCenter = getSpacePixel(pathPositionAtTime);
+
+                  newState.yOffset -= (pathYAtCenter - height / 2) / 4;
```
clarani: Why divide by 4 here ? Is it only to make the animation smoother ?

ui-spacetimechart/src/lib/types.ts
```diff
@@ -89,13 +91,17 @@ export type PointToData = (point: Point) => DataPoint;
...
 export const PICKING_LAYERS = ['paths'] as const;
 export type PickingLayerType = (typeof PICKING_LAYERS)[number];
-export const LAYERS = ['graduations', 'paths', 'captions', 'overlay'] as const;
+export const LAYERS = ['background', 'graduations', 'paths', 'captions', 'overlay'] as const;
 export type LayerType = (typeof LAYERS)[number];
 
+// PICKING SPECIFIC TYPES:
+export type PickingElement =
+  | { type: 'point'; path: string; point: Point }
+  | { type: 'segment'; path: string; from: Point; to: Point };
```
clarani:
```suggestion
  | { type: 'point'; pathId: string; point: Point }
  | { type: 'segment'; pathId: string; from: Point; to: Point };
```

ui-spacetimechart/src/hooks/useMouseInteractions.ts
jacomyal: If a new `position` object with the same `x` and `y` values is received (with something like `const newPosition = { ...position };` for instance), the old `[position]` dependencies version triggers the effect, while the new `[position.x, position.y]` does not.

ui-spacetimechart/src/hooks/useMouseTracking.ts
jacomyal: We only want to trigger the `"mousedown"` event when it's done on the `dom` element. But once the user starts dragging / panning, we want to keep interacting even when the mouse leaves the `dom` element. That's why, for `"mousemove"` and `"mouseup"`, we bind the events to `document` instead.

ui-spacetimechart/src/utils/png.ts
jacomyal: I personnally don't agree with that one. `String`, as well as `Boolean` and `Number`, are very misleading in JavaScript, because they are class constructors, initially designed to handle mutable objects-like reference-based strings, booleans and numbers - while we almost always handle those values as immutable value-based strings, booleans and numbers. The fact that we can use them as casting functions is weird, in terms of design.

I fear that some JavaScript beginner, seeing `const w = String(width);` might think is it a typo and should be `const w = new String(width);` instead, while `const w = width + '';` is unique and enough commonly used. I have the same opinion on `const isEmpty = !!array.length;` and `const now = +(new Date());`.

Now, if you want to enforce the use of these types of casting, we can add the [`no-implicit-coercion` ESLint rule](https://eslint.org/docs/latest/rules/no-implicit-coercion).

ui-spacetimechart/src/components/PathLayer.tsx
jacomyal: (I agree, but it won't be named `fontSize`, since it does not compute a font size, but the actual text bounding box height)

ui-spacetimechart/src/components/PathLayer.tsx
jacomyal: No, since the above is not a font size - but indeed, since a custom theme can override the actual font size, we should not use `FONT-SIZE`. I'll fix that :+1:

ui-spacetimechart/src/utils/png.ts
Math-R: You can also do ``` `${width}` ``` ?

ui-spacetimechart/src/stories/scroll-navigation.stories.tsx
jacomyal: Yes, exactly. I added a `frictionCoefficient` to make the code easier to understand.

ui-spacetimechart/src/stories/additional-data.stories.tsx
jacomyal: Agreed.

ui-spacetimechart/src/utils/png.ts
jacomyal: Yes, it does the exact same as `width + ''` I think.

ui-spacetimechart/src/stories/additional-data.stories.tsx
jacomyal: It will be done later.

Math-R (approved):
lgtm =)

clarani (approved):

Yohh (approved):

PR #70, opened 2024-05-27
core: make Select more generic

### LINKED ISSUE #69: [UI-CORE], select: Enhancing the flexibility of the Select component
With the aim of improving the code flexibility, I propose an enhancement to the Select component.

Currently, the Select component requires input data to be formatted in a certain way to be used as options. This can be restrictive for users who have to format their data before passing it to the component.

I propose to make the Select component more generic, allowing it to work with any type of input data. This could be achieved by passing an additional function as a prop to the Select component. This function would be used to determine how to extract the value and label from each object.

This would allow users to pass any array of objects to the Select component, without having to worry about how the data is formatted. This would make the Select component easier to use and more flexible.

kmer2016:
closes #69

### DIFF
--- a/ui-core/src/components/Select.tsx
+++ b/ui-core/src/components/Select.tsx
@@ -2,88 +2,76 @@ import React, { useState } from 'react';
 import FieldWrapper, { FieldWrapperProps } from './inputs/FieldWrapper';
 import cx from 'classnames';
 
-export type SelectOption = { value: string; label: string };
-
-/**
- * Type for the props of the Select component.
- *
- * We use Omit to create a new type based on React.InputHTMLAttributes<HTMLSelectElement>,
- * but without the 'value' property. This is because we want to enforce that the 'value'
- * prop for our Select component is always a string. In React, the 'value' prop for a
- * select element can be a string, a number, or an array of strings (for multiple selects),
- * but for our Select component, we want to ensure that it's always a string.
- */
-export type SelectProps = Omit<React.InputHTMLAttributes<HTMLSelectElement>, 'value'> &
+export type SelectProps<T> = Omit<
+  React.InputHTMLAttributes<HTMLSelectElement>,
+  'value' | 'onChange'
+> &
   Omit<FieldWrapperProps, 'children'> & {
-    value?: string;
-    placeholder?: string;
-    options: SelectOption[];
+    options: Array<T>;
+    value: T;
+    getOptionLabel: (option: T) => string;
+    getOptionValue: (option: T) => string;
+    onChange: (option?: T) => void;
   };
 
-const Select = React.forwardRef<HTMLSelectElement, SelectProps>(
-  (
-    {
-      id,
-      label,
-      hint,
-      value,
-      options,
-      placeholder,
-      statusWithMessage,
-      required,
-      disabled,
-      readOnly,
-      small,
-      onChange,
-    },
-    ref
-  ) => {
-    if (value && value != '' && !options.some((option) => option.value === value)) {
-      console.warn(`option "${value}" does not match any option`);
-    }
+const PLACEHOLDER_VALUE = '__PLACEHOLDER__';
 
-    const [selectedOption, setSelectedOption] = useState<string>(value || '');
+const Select = <T,>({
+  id,
+  label,
+  hint,
+  value,
+  options,
+  placeholder,
+  statusWithMessage,
+  required,
+  disabled,
+  readOnly,
+  small,
+  getOptionLabel,
+  getOptionValue,
+  onChange,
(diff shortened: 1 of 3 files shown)


ui-core/src/stories/Select.stories.tsx
clarani, line 1: Can you change the title of the story to 'core/Select' (l36) ?

ui-core/src/stories/Select.stories.tsx
```diff
@@ -1,20 +1,27 @@
...
-    options: [
-      { value: 'blue', label: 'Blue' },
-      { value: 'red', label: 'Red' },
-      { value: 'green', label: 'Green' },
-    ],
+    value: options[0],
+    options,
+    getOptionLabel: (option: unknown) => (option as Option).label,
+    getOptionValue: (option: unknown) => (option as Option).value,
+    onChange: (option?: unknown) => console.log(option),
```
clarani: I don't really like this, can you rather define a wrapper like this ? 
```
const SelectWrapper = (props: SelectProps<Option>) => {
  return <Select {...props} />;
};
```
You can use it instead of the Select l16 and then change your functions:
```suggestion
    getOptionLabel: (option: Option) => option.label,
    getOptionValue: (option: Option) => option.value,
    onChange: (option?: Option) => console.log(option),
```

ui-core/src/components/Select.tsx
```diff
@@ -2,88 +2,72 @@ import React, { useState } from 'react';
...
+export type SelectProps<T> = Omit<
+  React.InputHTMLAttributes<HTMLSelectElement>,
+  'value' | 'onChange'
+> &
   Omit<FieldWrapperProps, 'children'> & {
-    value?: string;
-    placeholder?: string;
-    options: SelectOption[];
+    options: Array<T>;
+    value?: T;
```
clarani: If value can be undefined, it should be handle in T
```suggestion
    value: T;
```

ui-core/src/components/Select.tsx
```diff
@@ -2,88 +2,72 @@ import React, { useState } from 'react';
...
+    getOptionLabel: (option: T) => string;
+    getOptionValue: (option: T) => string;
+    onChange: (option?: T) => void;
```
clarani, line 14:
```suggestion
    onChange: (option: T) => void;
```

ui-core/src/components/Select.tsx
```diff
@@ -2,88 +2,72 @@ import React, { useState } from 'react';
...
+  statusWithMessage,
+  required,
+  disabled,
+  readOnly,
+  small,
+  getOptionLabel,
+  getOptionValue,
+  onChange,
+}: SelectProps<T>) => {
+  const [selectedOption, setSelectedOption] = useState<T | undefined>(value);
```
clarani, line 35:
```suggestion
  const [selectedOption, setSelectedOption] = useState<T>(value);
```

ui-core/src/components/Select.tsx
```diff
@@ -2,88 +2,72 @@ import React, { useState } from 'react';
...
+      disabled={disabled}
+      small={small}
+    >
+      <select
+        className={cx('custom-select', {
+          'placeholder-selected': !selectedOption,
+          small,
+          'read-only': readOnly,
+          [statusWithMessage?.status || '']: !!statusWithMessage,
+        })}
```
clarani:
```suggestion
        className={cx('custom-select', `${statusWithMessage}`, {
          'placeholder-selected': !selectedOption,
          small,
          'read-only': readOnly,
        })}
```

clarani (changes requested):

ui-core/src/components/Select.tsx
kmer2016: I keep it like this, because the user can select the placeholder, in that case, the option can be undefined

ui-core/src/components/Select.tsx
kmer2016: I have kept it that way for the same reason I mentioned in the comment above.

ui-core/src/components/Select.tsx
```diff
@@ -48,18 +53,17 @@ const Select = <T,>({
       small={small}
     >
       <select
-        className={cx('custom-select', {
-          'placeholder-selected': !selectedOption,
+        className={cx('custom-select', `${statusWithMessage?.status || ''}`, {
```
clarani:
```suggestion
        className={cx('custom-select', statusWithMessage?.status, {
```

clarani (commented):
Can you squash your commits and rephrase it ? Here is an example: "core: make select more generic"

clarani (approved):
LGTM

PR #73, opened 2024-06-20
ui-speedspacechart: add power restrictions layer

### LINKED ISSUE #74: osrd-ui: add power restrictions layer in GEV V2
### Description and goal

the GEV V2 should be able to display a linear representation of the power restrictions under the curve layer

> [!CAUTION]
> the real power restrictions can't be seen in OSRD-UI
>**we have to use fake data**

### Acceptance criteria

- [ ] the power restrictions layer follows the design system mockup
- [ ] this layer is displayed or not in accordance to the settings panel
- [ ] this layer is synchronized with the user interactions (zoom, pan...)
- [ ] the power restrictions are displayed in the tooltip when hovering the curve layer

SharglutDev:
close #74


ui-speedspacechart/src/components/helpers/drawElements/electricalProfile.ts
```diff
@@ -74,7 +74,6 @@ export const drawElectricalProfile = (
         for (let i = 0; i < 9; i++) {
           ctx.fillRect(x, topRect + 15 + i * 3, profileWidth, 1);
         }
-        ctx.stroke();
```
emersion, line 77: Since we're getting rid of the various `stroke()` calls here in this file: I think we can get rid of `beginPath()` calls as well.

ui-speedspacechart/src/types/chartTypes.ts
```diff
@@ -44,6 +47,13 @@ export type TrainDetails = {
   marecoSpeedText: string;
   effortText: string;
   electricalProfileText: string;
+  powerRestrictionText: string;
   previousGradientText: number;
   modeText: string;
 };
+
+export enum LinearLayerHeights {
```
emersion: Hm, I find it a bit confusing to make this an `enum`, should this be an object instead (like `layerHeights` was)?

ui-speedspacechart/src/components/utils.ts
```diff
@@ -63,26 +63,34 @@ export const getAdaptiveHeight = (
...
+  let adjustment = 0;
+  const { electricalProfiles, powerRestrictions, speedLimitTags } = layersDisplay;
+
+  if (electricalProfiles && (powerRestrictions || speedLimitTags))
+    adjustment = LinearLayerHeights.electricalProfiles;
+
+  if (powerRestrictions && speedLimitTags) adjustment += LinearLayerHeights.powerRestrictions;
+
+  return height + adjustment - MARGINS.MARGIN_BOTTOM;
+};
```
emersion: Hmm, I'm not sure I'm following what this function is doing…

It seems like this function is computing the top margin specifically for the power restrictions layer? I don't think we'd be able to use it for other layers, since the top margin depends on the layer stacking order. If so, maybe it would be worth indicating in the function name that it's specific to power restrictions?

Second, I'm not sure I understand the conditions. `powerRestrictions` will always be true for the result of the function to be actually used by the caller. And ultimately I think we only care about which layer stacked on top of the power restrictions row is enabled?

It seems like the following code works fine in my (limited) testing:

```js
export const getLinearLayerMarginTop = (height: number, layersDisplay: Store['layersDisplay']) => {
  let adjustment = 0;
  const { electricalProfiles, powerRestrictions, speedLimitTags } = layersDisplay;

  if (electricalProfiles)
    adjustment += LinearLayerHeights.electricalProfiles;

  if (speedLimitTags) adjustment += LinearLayerHeights.speedLimitTags;

  return height + adjustment - MARGINS.MARGIN_BOTTOM;
};
```

Am I missing something here?

ui-speedspacechart/src/components/helpers/drawElements/reticle.ts
```diff
@@ -127,6 +129,13 @@ export const drawCursor = (
         electricalModeText = `${isElectrified ? electrificationUsage.mode : electrificationUsage.object_type}`;
       }
 
+      const previousPowerRestriction = powerRestrictions.findLast(
+        (range) => range.start <= previousCurvePosition.position
```
emersion: Shouldn't we also check `range.end` here? Or do we want to display text for the closest power restriction on the left hand side even if we're past it?

ui-speedspacechart/src/components/helpers/drawElements/powerRestrictions.ts
```diff
@@ -0,0 +1,138 @@
...
+  ctx.font = 'normal 12px IBM Plex Sans';
+  ctx.fillStyle = 'rgb(121, 118, 113)';
+  ctx.strokeStyle = 'rgb(121, 118, 113)';
+  ctx.lineWidth = 1;
+  ctx.textAlign = 'center';
+
+  let powerRestrictionsBackgroundColor = LINEAR_LAYERS_BACKGROUND_COLOR.FIRST;
+
+  // Power restrictions linear can either be the first layer
+  // if electrical profiles aren't displayed or second if they are
```
emersion, line 56: … Or the third layer if both electrical profiles and speed limit tags are displayed? (Does this matter? Do we need to adjust the background color for the speed limit tags layer?)

Note, the difference between FIRST and SECOND is very subtle - I used the browser color picker to tell them apart.

ui-speedspacechart/src/components/helpers/drawElements/powerRestrictions.ts
```diff
@@ -0,0 +1,138 @@
...
+      const textStartX = (startX + horizontalEndX) / 2 - textWidth / 2 - TEXT_PADDING_HORIZONTAL;
+      const textEndX = (startX + horizontalEndX) / 2 + textWidth / 2 + TEXT_PADDING_HORIZONTAL;
+
+      ctx.moveTo(startX, horizontalY);
+      ctx.lineTo(textStartX - TEXT_PADDING_HORIZONTAL, horizontalY);
+      ctx.moveTo(textEndX + TEXT_PADDING_HORIZONTAL, horizontalY);
+      ctx.lineTo(horizontalEndX - 2, horizontalY);
+
+      const textX = (startX + horizontalEndX) / 2;
+      ctx.fillText(range.code, textX, horizontalY + 4); // text centering adjustment
```
emersion: Would we be able to leverage [`textBaseline`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/textBaseline) to drop this constant?

ui-speedspacechart/src/components/helpers/drawElements/electricalProfile.ts
SharglutDev: Good point ! Done.

ui-speedspacechart/src/components/helpers/drawElements/powerRestrictions.ts
SharglutDev: Done

ui-speedspacechart/src/types/chartTypes.ts
SharglutDev: Done

ui-speedspacechart/src/components/utils.ts
SharglutDev: Good catch ! Done by adding a third boolean parameter for the speed limit tags linear layer

ui-speedspacechart/src/components/helpers/drawElements/reticle.ts
SharglutDev: Nice catch, rectified the condition and change shortened one of the power restriction range so we can test this edge case easier.

ui-speedspacechart/src/components/helpers/drawElements/powerRestrictions.ts
SharglutDev: Power restrictions layer can't be at third position according the sketch

ui-speedspacechart/src/components/SpeedSpaceChart.tsx
```diff
@@ -111,6 +114,7 @@ const SpeedSpaceChart = ({
       electrification: data.simulation.present.trains[0].electrification_ranges || [],
       slopes: data.simulation.present.trains[0].slopes || [],
       electricalProfiles: data.electricalProfiles,
+      powerRestrictions: data.powerRestrictions || [],
```
jacomyal: Why are there these `|| []` when the data type enforces the presence of data?

ui-speedspacechart/src/components/common/DetailsBox.tsx
```diff
@@ -13,6 +13,7 @@ type DetailsBoxProps = {
   effortText: string;
   electricalModeText: string;
   electricalProfileText: string;
+  powerRestrictionText: string;
```
jacomyal, line 16: You added this required prop, but you did not update the places where `DetailsBox` is instantiated. So, you need to handle `<DetailsBox width={width} height={height} store={store} {...debouncedDetailsBox} />` in `ReticleLayer.tsx`:
https://github.com/OpenRailAssociation/osrd-ui/blob/8034a6290d0c6716338925230103663b7804c7c4/ui-speedspacechart/src/components/layers/ReticleLayer.tsx#L52

ui-speedspacechart/src/types/chartTypes.ts
```diff
@@ -1,8 +1,10 @@
+import { LINEAR_LAYERS_HEIGHTS } from '../components/const';
```
jacomyal: When I build the project, this line throws an error:
```
 [plugin typescript] src/types/chartTypes.ts (1:1): @rollup/plugin-typescript TS6133: 'LINEAR_LAYERS_HEIGHTS' is declared but its value is never read.
/home/jacomyal/code/osrd-ui/ui-speedspacechart/src/types/chartTypes.ts:1:1

1 import { LINEAR_LAYERS_HEIGHTS } from '../components/const';
```

I think it is because it creates a circular dependency between `chartTypes.ts` and `const.ts`.

ui-speedspacechart/src/components/const.ts
```diff
@@ -10,12 +10,20 @@ export const MARGINS = {
...
 };
 
-export const LAYERS_HEIGHTS = {
+export const LINEAR_LAYERS_HEIGHTS = {
   ELECTRICAL_PROFILES_HEIGHT: 56,
   POWER_RESTRICTIONS_HEIGHT: 40,
   SPEED_LIMIT_TAGS_HEIGHT: 40,
 };
 
+export const LinearLayerHeightsByName = {
```
jacomyal: This variable name should be UPPER_CASE_SNAKE_CASE instead of PascalCase for consistency (it actually bothered me when I was reading your modifications to `components/utils.ts`, it looked like you were iterating on the keys of a type).

jacomyal (changes requested):

ui-speedspacechart/src/components/common/DetailsBox.tsx
SharglutDev: I don't see what I should add regarding power restriction layer.
The error in this file came from the missing property `electricalModeText` in the `TrainDetails` type which seems to be an oversight from the electrical profiles PR.

ui-speedspacechart/src/types/chartTypes.ts
SharglutDev: Actually this wasn't used anymore so I removed it.

ui-speedspacechart/src/components/const.ts
SharglutDev: Done

ui-speedspacechart/src/components/utils.ts
```diff
@@ -63,26 +63,39 @@ export const getAdaptiveHeight = (
...
+  height: number,
+  layersDisplay: Store['layersDisplay'],
+  isSpeedLimitByTag: boolean = false
+) => {
+  let adjustment = 0;
+  const { electricalProfiles, powerRestrictions, speedLimitTags } = layersDisplay;
+
+  if (electricalProfiles) adjustment = LinearLayerHeightsByName.electricalProfiles;
+
+  if (powerRestrictions && speedLimitTags && isSpeedLimitByTag)
```
emersion: Nit: I believe the `speedLimitTags` condition is unnecessary here. If this is called with `isSpeedLimitByTag === true`, then `speedLimitTags` is necessarily true as well.

ui-speedspacechart/src/components/SpeedSpaceChart.tsx
emersion: Indeed, `OsrdSimulationState.powerRestrictions` cannot be null.

ui-speedspacechart/src/components/helpers/drawElements/powerRestrictions.ts
emersion: It seems like [`textBaseline`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/textBaseline) isn't used in the last push? Did this end up not working?

ui-speedspacechart/src/components/helpers/drawElements/powerRestrictions.ts
emersion: Indeed!

ui-speedspacechart/src/components/helpers/drawElements/powerRestrictions.ts
emersion: To be clear, my suggestion here is to add something like `ctx.textBaseline = 'middle'`.

ui-speedspacechart/src/components/SpeedSpaceChart.tsx
SharglutDev: Good point ! Actually powerRestrictions can be undefined (like electricalProfiles). So I changed the logic to match this case.
And I added a logic in the settings panel to disable the layer checkbox when their data is undefined.

ui-speedspacechart/src/components/utils.ts
SharglutDev: Good point, done.

ui-speedspacechart/src/components/helpers/drawElements/powerRestrictions.ts
SharglutDev: Oh got it, I'll check that

SharglutDev: Here is a comparative of your solution (bottom view), the hard coded one (top left) and the sketch (top right).

Capture d’écran 2024-07-09 à 15 44 05

It seems like with textBaseLine middle, it's 1 or 2 pixel to high. I'll check with @thibautsailly if this needs to be pixel perfect (for the letter horizontal lines aligned with the main horizontal line) or not.

ui-speedspacechart/src/components/common/DetailsBox.tsx
jacomyal: OK, my bad

jacomyal:
LGTM - and great work, by the way :+1:

jacomyal (approved):

ui-speedspacechart/src/components/helpers/drawElements/powerRestrictions.ts
emersion: The `textBaseline` solution might look better when some of the characters' shapes are under the baseline, such as "p" or "g". The "F" and "P" characters' shapes are all above the baseline (they take up a lot more space on the top).

emersion: Hm, actually it seems like it's the other way around. Maybe using `horizontalY + 0.5` instead of `horizontalY` would help? (half the `lineWidth`)

emersion: Better yet, `horizontalY + ctx.lineWidth / 2` to better reflect the intent

emersion (approved):
LGTM, nice work! The `textBaseline` thing is optional :)

ui-speedspacechart/src/components/helpers/drawElements/powerRestrictions.ts
SharglutDev: Oh yep, it seems to be perfect like that !

SharglutDev: Done

kmer2016 (approved):
LGTM !

PR #86, opened 2024-07-05
ui-manchette : integration get manchette

Math-R:


ui-manchette/src/components/Manchette.tsx
```diff
@@ -1,51 +1,217 @@
...
 const Manchette: React.FC<ManchetteProps> = ({
   operationalPoints,
-  children,
+  projectPathTrainResult,
   isProportionnal = false,
 }) => {
-  const [zoom, setZoom] = React.useState(1);
+  const manchette = useRef<HTMLDivElement>(null);
+
+  const [zoomY, setZoomY] = React.useState(1);
```
Yohh: import useState ?

ui-manchette/src/components/helpers.ts
```diff
@@ -1,3 +1,72 @@
+import { OperationalPointType, StyledOperationalPointType } from '../types';
+import { BASE_OP_HEIGHT } from './consts';
+
 export const positionMmtoKm = (position: number) => {
```
Yohh:
```suggestion
export const positionMmToKm = (position: number) => {
```

Yohh (commented):
:muscle:

ui-manchette/src/components/Manchette.tsx
```diff
@@ -1,51 +1,217 @@
...
+import {
+  OperationalPoint,
+  PathData,
+  SpaceScale,
+} from '@osrd-project/ui-spacetimechart/dist/lib/types';
+import {
+  BASE_KM_HEIGHT,
+  BASE_OP_HEIGHT,
+  INITIAL_OP_LIST_HEIGHT,
+  INITIAL_SPACE_TIME_CHART_HEIGHT,
```
kmer2016: `INITIAL_OP_LIST_HEIGHT` is not used

```suggestion
  BASE_KM_HEIGHT,
  BASE_OP_HEIGHT,
  INITIAL_SPACE_TIME_CHART_HEIGHT,
```

ui-manchette/src/hooks/isOverFlowed.ts
```diff
@@ -0,0 +1,39 @@
+import React, { useState, useLayoutEffect, useRef, RefObject } from 'react';
```
kmer2016: `React` and `useRef` are not used
```suggestion
import { useState, useLayoutEffect, RefObject } from 'react';
```

ui-manchette/src/hooks/isOverFlowed.ts
kmer2016, line 1: rename the filename wbyith the name of the hook ?

ui-manchette/src/hooks/isOverFlowed.ts
```diff
@@ -0,0 +1,39 @@
...
+
+interface CallbackFunction {
+  (hasOverflow: boolean): void;
+}
+
+interface CallbackFunction {
+  (hasOverflow: boolean): void;
+}
```
kmer2016: As in osrd, use type instead of interface and delete duplications
```suggestion
type CallbackFunction = (hasOverflow: boolean) => void;
```

ui-manchette/src/components/Manchette.tsx
```diff
@@ -1,51 +1,217 @@
...
+    operationalPoints.map((point) => {
+      operationalPointswithPosition.push({
+        id: point.id,
+        label: point.id,
+        position: point.position,
+        importanceLevel: 1,
+      });
+    });
+    return operationalPointswithPosition;
+  };
```
kmer2016:
```suggestion
  const getOperationalPointsWithPosition = (operationalPoints: StyledOperationalPointType[]) => {
    return operationalPoints.map(
      (point) =>
        ({
          id: point.id,
          label: point.id,
          position: point.position,
          importanceLevel: 1,
        }) as OperationalPoint
    );
  };
```

ui-manchette/src/components/Manchette.tsx
```diff
@@ -1,51 +1,217 @@
...
+      const dataPoints = [];
+      for (let i = 0; i < curve.positions.length; i++) {
+        dataPoints.push({
+          time: curve.times[i] + +new Date(path.departure_time),
+          position: curve.positions[i],
+        });
+      }
+      return dataPoints;
+    }),
+  }));
```
kmer2016: I suggest here that to refactore the points mapping to directly return the result and wrapped it in `useMemo` for performance optimization, improving code readability and conciseness.
```suggestion
const paths = useMemo(() => projectPathTrainResult.map((path, index) => ({
    id: path.departure_time,
    label: `Train ${index + 1}`,
    color: '#000000',
    points: path.space_time_curves.flatMap((curve) => {
      return curve.positions.map((position, i) => ({
        time: curve.times[i] + +new Date(path.departure_time),
        position,
      }));
    }),
  })), [projectPathTrainResult]);
```

ui-manchette/src/components/Manchette.tsx
kmer2016, line 1: May be move `getScales`, `getOperationalPointsWithPosition` and the code which compute `paths` at line 70 in helper.ts file ?

kmer2016 (commented):
Great work ! Add some small refacto. suggestions. I noticed that there are no unit tests included for helpers. This can be done in a separate PR if needed

ui-manchette/src/components/Manchette.tsx
Math-R: yes it is

ui-manchette/src/components/Manchette.tsx
Math-R: done

ui-manchette/src/components/Manchette.tsx
Math-R: done i created a new hook

ui-manchette/src/components/Manchette.tsx
Math-R: done

ui-manchette/src/components/helpers.ts
Math-R: i don't get your comment

ui-manchette/src/hooks/isOverFlowed.ts
Math-R: done

ui-manchette/src/hooks/isOverFlowed.ts
Math-R: done

ui-manchette/src/hooks/isOverFlowed.ts
Math-R: done

ui-manchette/src/components/Manchette.tsx
Math-R: done

kmer2016 (approved):
LGTM and tested ! Well done !

Yohh (approved):
nice! don't forget to add `type` prefix when you import a Type

ui-manchette/src/components/helpers.ts
Yohh: a lowercase issue for `to` => `To`

ui-manchette/package.json
```diff
@@ -36,7 +36,8 @@
   "dependencies": {
     "classnames": "^2.5.1",
     "lodash.isequal": "^4.5.0",
-    "tailwindcss": "^3.4.1"
+    "tailwindcss": "^3.4.1",
```
jacomyal, line 39: Since this `@osrd-project/ui-manchette` module exports its own built CSS code, why is TailWind in the dependencies? Can I use it without Tailwind? And if I use Tailwind in my project that depends on `@osrd-project/ui-manchette`, what is the benefit?

ui-manchette/src/components/Manchette.tsx
```diff
@@ -1,51 +1,202 @@
...
-  operationalPoints,
-  children,
-  isProportionnal = false,
-}) => {
-  const [zoom, setZoom] = React.useState(1);
+  const paths = usePaths(projectPathTrainResult);
 
-  const zoomIn = () => {
-    if (zoom < 10) setZoom(zoom + 0.5);
+  const zoomYIn = () => {
```
jacomyal, line 58: It would be nice to have those three constants (min, max, step) stored and named somewhere, like in a `consts.ts` file somewhere. I would even argue to make them overridable through props.

ui-manchette/src/components/helpers.ts
```diff
@@ -1,3 +1,97 @@
...
+import type { OperationalPointType, StyledOperationalPointType } from '../types';
+import { BASE_KM_HEIGHT, BASE_OP_HEIGHT } from './consts';
+
+export const positionMmToKm = (position: number) => {
   return Math.round((position / 1000000) * 10) / 10;
 };
+
+export const calcOperationalPointsToDisplay = (
+  operationalPoints: OperationalPointType[],
+  isProportionnal: boolean,
```
jacomyal, line 11: I only point it here, but there are a lot of typos with `proportionnal` that should be `proportional` with one single **N** in the manchette codebase.

ui-manchette/src/components/helpers.ts
```diff
@@ -1,3 +1,97 @@
...
+      }
+    }
+    result.shift();
+    return result;
+  } else {
+    return operationalPoints as StyledOperationalPointType[];
+  }
+};
+
+export const operationalPointsHeight = (
```
jacomyal, line 52: I see that in this function, you mutate existing `styles` attributes for the input `operationalPoints`.

I think this pattern is dangerous: When I see this function that takes a `StyledOperationalPointType[]` and returns another `StyledOperationalPointType[]`, I would expect it not to mutate what's inside the input value:
- Either you return something totally new
- Either you mutate, and name and type the function accordingly (like `const setOperationalPointsHeight: (...) => void` for instance)

I know JavaScript itself does not respect this convention (like with `Array#reverse` for instance), but it has been widely criticized, and it is now quite standard to have no side effects on functions that return something new.

Anyway, a common way to avoid mutating the input would be to replace the `op.styles = {...}; return op;` by `return {...op, styles: {...}};`.

ui-manchette/src/types.ts
```diff
@@ -1,3 +1,5 @@
+import React from 'react';
```
jacomyal, line 1: Only import `{ CSSProperties }` instead would be better I think

ui-manchette/src/types.ts
```diff
@@ -80,3 +82,43 @@ export type ArrayElement<ArrayType extends readonly unknown[] | null | undefined
...
+    time_end: number;
+    /** The aspects start being displayed at this time (number of mseconds since `departure_time`) */
+    time_start: number;
+  }[];
+  /** List of space-time curves sections along the path */
+  space_time_curves: {
+    positions: number[];
+    times: number[];
+  }[];
+} & {
```
jacomyal, line 117: Why are there two types that are combined with `{ ... } & { ... }` instead of one single type?

ui-manchette/src/components/Manchette.tsx
```diff
@@ -1,51 +1,202 @@
...
-  children?: React.ReactNode;
-  isProportionnal?: boolean;
+  projectPathTrainResult: ProjectPathTrainResult[];
 };
+import usePaths from '../hooks/usePaths';
+
+const Manchette: React.FC<ManchetteProps> = ({ operationalPoints, projectPathTrainResult }) => {
+  const manchette = useRef<HTMLDivElement>(null);
+
+  const [zoomY, setZoomY] = useState(1);
```
jacomyal, line 30: Why do you use so many separate `setState` instead of one single "larger" `setState` such as something like this for instance?

```typescript
const [state, setState] = useState({
  zoomX: 1,
  zoomY: 1,
  offsetX: 0,
  offsetY: 0,
  // ...
});
```

The "more compact" version seems easier to read for me (it's easy to have a full overview of what's in the state). Do you see any pros on your "split" version?

Issue #93, opened 2024-07-07
Add normal & primary btn stories


ghost:
No `Normal` or `Primary` btn stories were in storybook. Adding these stories is helpful, so programmers & designers can see what each btn type looks like

PR #94, opened 2024-07-07
ui-core: add normal & primary btn stories

ghost:
Issue #93

### DIFF
--- a/ui-core/src/stories/Button.stories.tsx
+++ b/ui-core/src/stories/Button.stories.tsx
@@ -33,6 +33,41 @@ const meta: Meta<typeof Button> = {
 export default meta;
 type Story = StoryObj<typeof Button>;
 
+export const Normal: Story = {
+  args: {
+    label: 'Normal',
+    variant: 'Normal',
+  },
+};
+
+export const Primary: Story = {
+  args: {
+    label: 'Primary',
+    variant: 'Primary',
+  },
+};
+
+export const Quiet: Story = {
+  args: {
+    label: 'Quiet',
+    variant: 'Quiet',
+  },
+};
+
+export const Destructive: Story = {
+  args: {
+    label: 'Destructive',
+    variant: 'Destructive',
+  },
+};
+
+export const Cancel: Story = {
+  args: {
+    label: 'Cancel',
+    variant: 'Cancel',
+  },
+};
+
 export const Default: Story = {
   args: {
     label: 'Click me',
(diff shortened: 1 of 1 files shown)


ghost:
Are others okay with brainstorming other `ButtonVariant` types than `Normal`, `Cancel`, `Quiet`, `Destructive`, `Primary`. I find some of these names confusing. 

For example, both `Normal` & `Primary` imply default. The `Normal` variant appears grey. Could I rename the `Normal` variant to `secondary` or `grey`?

`Destructive` & `Cancel` is also a little confusing. Maybe `Destructive` could be renamed to `warning` or `critical`? `Cancel` could be called grayed outline?

`Quiet` is a sound. Maybe we could rename to `Subtle`?

ghost:
@sim51 @emersion I seem not able to request a review. Would you please review this pr? :)

sim51 (approved):

emersion:
> I find some of these names confusing.

I'd prefer to keep semantic names rather than visual/color names. In other words, I think it's better to use a name that explains _when_ a button should be used, rather than describe how it actually looks like. The reason is that the visual appearance of the button might change, e.g. because the UI designer updates the mockups, or because a dark theme is added. Additionally, buttons should be used consistently in the whole app - we're more likely to end up with button variants used for the wrong use-case if they're named after colors "just because it looked better" to a developer.

ghost:
> > I find some of these names confusing.
> 
> I'd prefer to keep semantic names rather than visual/color names. In other words, I think it's better to use a name that explains _when_ a button should be used, rather than describe how it actually looks like. The reason is that the visual appearance of the button might change, e.g. because the UI designer updates the mockups, or because a dark theme is added. Additionally, buttons should be used consistently in the whole app - we're more likely to end up with button variants used for the wrong use-case if they're named after colors "just because it looked better" to a developer.

Thanks for the feedback & explanation. I'm happy to stick with the existing options :)

SharglutDev:
Hi @turnerian2004, it seems like on each story for buttons (default, loading, disabled...) that you can change the variant of the button to see what it look likes.

Capture d’écran 2024-07-16 à 09 47 47

I feel like it's not necessary to add more stories (there already a lot of them) to display just a different variant when you can already switch that prop pretty easily.

Is that ok for you ?

ghost:
Yes, that's a very good point @SharglutDev! Thanks for the insights. I learnt something new. We can close this pr without merging from my end :)

PR #99, opened 2024-07-09
ui-speedspacechart: fix detailsBox display

### LINKED ISSUE #98: ui-speedspacechart : hide reticle and detailsbox when hovering settings panel
Issue reported by @thibautsailly 

## Description

When the cursor pass on the settings panel, we can see the blured details box behind the panel. This is not a desired behavior.

https://github.com/OpenRailAssociation/osrd-ui/assets/113104137/ecd1ca08-d54d-4384-8927-0538cf7e14d0

## Acceptance Criteria

- [ ] When the cursor hovers the panel, we want to hide the reticle and details box

Yohh:
- add showReticleLayer state in SpeedSpaceChart.tsx
- display or not ReticleLayer while onMouseDown or onMouseLeave SettingsPanel

close #98

### DIFF
--- a/ui-speedspacechart/src/components/SpeedSpaceChart.tsx
+++ b/ui-speedspacechart/src/components/SpeedSpaceChart.tsx
@@ -101,6 +101,7 @@ const SpeedSpaceChart = ({
     : width;
 
   const [showDetailsBox, setShowDetailsBox] = useState(false);
+  const [isMouseHoveringSettingsPanel, setIsMouseHoveringSettingsPanel] = useState(false);
 
   const reset = () => {
     setStore((prev) => ({
@@ -162,6 +163,7 @@ const SpeedSpaceChart = ({
             color={backgroundColor}
             store={store}
             setStore={setStore}
+            setIsMouseHoveringSettingsPanel={setIsMouseHoveringSettingsPanel}
             translations={translations}
           />
         </div>
@@ -187,14 +189,14 @@ const SpeedSpaceChart = ({
       <TickLayerY width={width} height={height} store={store} />
       {store.layersDisplay.electricalProfiles && (
         <ElectricalProfileLayer
-          width={width}
+          width={adjustedWidthRightAxis}
           height={height + LINEAR_LAYERS_HEIGHTS.ELECTRICAL_PROFILES_HEIGHT}
           store={store}
         />
       )}
       {store.layersDisplay.powerRestrictions && (
         <PowerRestrictionsLayer
-          width={width}
+          width={adjustedWidthRightAxis}
           marginTop={getLinearLayerMarginTop(height, store.layersDisplay)}
           store={store}
         />
@@ -204,15 +206,16 @@ const SpeedSpaceChart = ({
       {store.layersDisplay.declivities && (
         <TickLayerYRight width={width} height={height} store={store} />
       )}
-      <ReticleLayer
-        width={adjustedWidthRightAxis}
-        height={dynamicHeight}
-        heightOffset={dynamicHeightOffset}
-        store={store}
-        showDetailsBox={showDetailsBox}
-        setShowDetailsBox={setShowDetailsBox}
-      />
-
+      {!isMouseHoveringSettingsPanel && (
+        <ReticleLayer
+          width={adjustedWidthRightAxis}
+          height={dynamicHeight}
+          heightOffset={dynamicHeightOffset}
+          store={store}
+          showDetailsBox={showDetailsBox}
+          setShowDetailsBox={setShowDetailsBox}
+        />
+      )}
       <FrontInteractivityLayer
         width={WIDTH_OFFSET}
         height={dynamicHeightOffset}
--- a/ui-speedspacechart/src/components/common/SettingsPanel.tsx
+++ b/ui-speedspacechart/src/components/common/SettingsPanel.tsx
@@ -10,11 +10,19 @@ type SettingsPanelProps = {
   color: string;
   store: Store;
   setStore: React.Dispatch<React.SetStateAction<Store>>;
+  setIsMouseHoveringSettingsPanel: React.Dispatch<React.SetStateAction<boolean>>;
   translations?: SpeedSpaceChartProps['translations'];
 };
 
-const SettingsPanel = ({ color, store, setStore, translations }: SettingsPanelProps) => {
+const SettingsPanel = ({
+  color,
+  store,
+  setStore,
+  setIsMouseHoveringSettingsPanel,
+  translations,
+}: SettingsPanelProps) => {
   const closeSettingsPanel = () => {
+    setIsMouseHoveringSettingsPanel(false);
     setStore((prev) => ({
       ...prev,
       isSettingsPanelOpened: false,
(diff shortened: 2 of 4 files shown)


ui-speedspacechart/src/components/SpeedSpaceChart.tsx
```diff
@@ -145,6 +146,7 @@ const SpeedSpaceChart = ({
             color={backgroundColor}
             store={store}
             setStore={setStore}
+            setShowReticleLayer={setShowReticleLayer}
```
anisometropie: From a design standpoint, I don’t think SettingsPanel needs to know or mutate the state of something the belongs to reticleLayer.

The parent should know if the settingsPanel is displayed and use that.

can’t you just use `store.isSettingsPanelOpened` as the condition to display ReticleLayer ?

anisometropie (changes requested):

ui-speedspacechart/src/components/SpeedSpaceChart.tsx
Yohh: that's an interesting point, the fact is I thought that we should inspect the graph even if settings panel is open, but not if we hover it, but if @thibautsailly agree with your proposition I m not against it

Akctarus (approved):
Tested, LGTM

ui-speedspacechart/src/components/SpeedSpaceChart.tsx
anisometropie: ah if the panel is open it would hide the reticle even while we’re above the graph, I didn’t see that.

the current behavior is good, no need to change that.

in that case you can just rename `showReticleLayer` into `mouseIsOverSettingsPanel` (that’s really what it’s testing) to avoid having what looks like reticle logic inside the settings component

to make the intention of the code really clear.

ui-speedspacechart/src/components/common/SettingsPanel.tsx
```diff
@@ -26,6 +34,8 @@ const SettingsPanel = ({ color, store, setStore, translations }: SettingsPanelPr
       id="settings-panel"
       style={{ background: `rgba(${color.substring(4, color.length - 1)}, 0.4)` }}
       className="font-sans"
+      onMouseMove={() => setShowReticleLayer(false)}
```
clarani: Why not use onMouseEnter ?

clarani (changes requested):
image
You still have an issue if the cursor is near the panel.

I also have the issue if I hover a PR.
image

ui-speedspacechart/src/components/common/SettingsPanel.tsx
Akctarus: For the first case (the cursor is near the panel), it has been validated by @thibautsailly  to leave it like this

Yohh: for the first comment: the two cases are ok for me, I can adapt the position of the box if you want, let me know if you want me to modify it 
for the PR layers: my bad, I just missed a pixel

ui-speedspacechart/src/components/SpeedSpaceChart.tsx
Yohh: done

ui-speedspacechart/src/components/common/SettingsPanel.tsx
Yohh: done

anisometropie (approved):

ui-speedspacechart/src/components/SpeedSpaceChart.tsx
```diff
@@ -90,6 +90,7 @@ const SpeedSpaceChart = ({
   const dynamicHeightOffset = getAdaptiveHeight(HEIGHT_OFFSET, store.layersDisplay);
 
   const [showDetailsBox, setShowDetailsBox] = useState(false);
+  const [mouseIsOverSettingsPanel, setMouseIsOverSettingsPanel] = useState(true);
```
SharglutDev: To match the usual css vocabulary, shouldn't we use "hover" for this state ?

```suggestion
  const [mouseHoveringSettingsPanel, setMouseHoveringSettingsPanel] = useState(true);
```

ui-speedspacechart/src/components/SpeedSpaceChart.tsx
```diff
@@ -180,14 +182,16 @@ const SpeedSpaceChart = ({
...
-      <ReticleLayer
-        width={width}
-        height={dynamicHeight}
-        heightOffset={dynamicHeightOffset}
-        store={store}
-        showDetailsBox={showDetailsBox}
-        setShowDetailsBox={setShowDetailsBox}
-      />
+      {mouseIsOverSettingsPanel && (
+        <ReticleLayer
```
SharglutDev: Shouldn't this be the opposite ? When the mouse doesn't hover the settings panel, we want to see the reticle I think (so we need to invert the setState : mouseEnter > true and mouseLeave > false)

```suggestion
      {!mouseIsOverSettingsPanel && (
        <ReticleLayer
```

SharglutDev (commented):
Thank you for this fix, left some suggestions.

ui-speedspacechart/src/components/SpeedSpaceChart.tsx
Yohh: you're right if we talk about css, but here it's JSX rendering, anyway if you want this change I can do it

ui-speedspacechart/src/components/SpeedSpaceChart.tsx
Yohh: if we agree with your previous comment you're right, and in that case `mouseHoveringSettingsPanel` should be initially set to `false`

ui-speedspacechart/src/components/SpeedSpaceChart.tsx
Yohh: to be more explicit: if this state triggers a class I can accept it's about css, but it isn't what it does

I can also change the name to avoid the possible missunderstanding over/hover, @anisometropie what do you think about it?

ui-speedspacechart/src/components/SpeedSpaceChart.tsx
clarani: I agree with Paul on this, if I read only this line, I understand that the reticle layer is displayed only if the mouse is over the settings panel.
It's quite confusing

ui-speedspacechart/src/components/common/SettingsPanel.tsx
clarani: Yes it would be nice to adapt the position of the box if we are close to the settings panel :+1:

ui-speedspacechart/src/components/SpeedSpaceChart.tsx
Yohh: sure, so what do I do, do I use his suggestion `mouseHoveringSettingsPanel` even if it's not about css and apply this second suggestion?

I begin to figure out that my first proposal `[showReticleLayer, setShowReticleLayer]` was better...

ui-speedspacechart/src/components/common/SettingsPanel.tsx
Yohh: done

ui-speedspacechart/src/components/SpeedSpaceChart.tsx
anisometropie: I think if the mouse is hovering over the panel, the mouse being **over** the panel is the important part to express in that name.

ui-speedspacechart/src/components/SpeedSpaceChart.tsx
Yohh: anyway I don't care, it's done

clarani (approved):
LGTM  & tested :white_check_mark:

Math-R (approved):
lgtm and tested to

PR #109, opened 2024-07-10
Add check commit title action

flomonster:


younesschrifi (approved):
LGTM. 
Perhaps you could add an example commit in the error message if you think this is necessary ?

flomonster:
> LGTM. Perhaps you could add an example commit in the error message if you think this is necessary ?

I just copied the script from the osrd repository.

bloussou (approved):
Thanks ✨

Yohh (approved):
:muscle:

PR #168, opened 2024-07-10
ui-icons: add class and color in the icons props

### LINKED ISSUE #9: UI-icons: update the icon component to add classname props
The Icon component can not receive classname for the moment, so we can not style it from OSRD. It's painful, especially for the color.

Goal: allow the component to receive classname prop so we can style it from OSRD.

# Acceptance criteria
- [ ] the component can be styled from OSRD through a classname prop

SarahBellaha:
Closes #9

### DIFF
--- a/ui-icons/scripts/build.js
+++ b/ui-icons/scripts/build.js
@@ -152,10 +152,10 @@ for (const [name, currentData] of Object.entries(representation)) {
 
     return [
       `IconReplaceName${variant}Props`,
-      `interface IconReplaceName${variant}Props { size?: ${sizeStr}; variant?: "${variant}"; title?: string; };`,
+      `interface IconReplaceName${variant}Props { size?: ${sizeStr}; variant?: "${variant}"; title?: string; iconColor?: string; className?: string; };`,
     ];
   });
-  const iconPropsContent = definitions.map(([name, content]) => `${content}\n`).join('\n');
+  const iconPropsContent = definitions.map(([_, content]) => `${content}\n`).join('\n');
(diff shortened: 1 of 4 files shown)


ui-icons/scripts/build.js
```diff
@@ -152,10 +152,10 @@ for (const [name, currentData] of Object.entries(representation)) {
 
     return [
       `IconReplaceName${variant}Props`,
-      `interface IconReplaceName${variant}Props { size?: ${sizeStr}; variant?: "${variant}"; title?: string; };`,
+      `interface IconReplaceName${variant}Props { size?: ${sizeStr}; variant?: "${variant}"; title?: string; iconColor?: string; className?: string; };`,
```
ghost, line 155: A friendly suggestion, maybe the variable name `iconFillColor` is better, since the variable applies a colour that `fills` the interior of the icon? :)

Yohh: I have a doubt... fill is usually used as an SVG attribute, here it's not exactly the same concept

Yohh (approved):
LGTM :muscle:

ui-icons/scripts/build.js
SharglutDev: For most of the icons, it doesn't fill the "interior" of the icons but only changes the stroke color of it. As we don't have 2 distinct props to change to color of the stroke or the "interior", I think it's fine to keep this prop named "iconColor" (maybe just `color` would have been fine too)

SharglutDev (approved):
Lgtm and tested, thank you !

PR #225, opened 2024-07-19
ui-warpedmap: fix useeffect warnings

### LINKED ISSUE #224: Fix warning ui warped map
We have two warnings in osrd-ui PRs related to ui warped map missing dependencies in the useEffect. 
image

We should add a rule for those specifics warnings if this dependency isn't convenient.

RomainValls:
close #224


jacomyal (approved):

SharglutDev (approved):
Lgtm but how come we don't have these warnings in osrd-project ?
I don't see any global rule for that in eslintrc or package.json, do you know why @jacomyal ?

PR #271, opened 2024-07-28
ui-core: fix text-area story background

ghost:
Issue [273](https://github.com/OpenRailAssociation/osrd-ui/issues/273)

The story background is restricted, so the background does not 'wrap' the component fully.

After
fix - textarea

Before
problem - textarea


clarani (approved):
Nice fix, thank you very much :pray:

PR #272, opened 2024-07-28
ui-core: add label stories

ghost:
Issue [274](https://github.com/OpenRailAssociation/osrd-ui/issues/274)

`Comment:` The `Label` component has a `hint` prop. I do not see the value of this prop, since there is no `star` like in the `required` story or additional `text` like in the stories of the components that have a `hint` prop. 

Should the `hint` prop be removed? Should the `Hint` component be passed to the `Label` component as an optional prop?

### DIFF
--- a/ui-core/src/stories/Label.stories.tsx
+++ b/ui-core/src/stories/Label.stories.tsx
@@ -1,0 +1,55 @@
+import React from 'react';
+import type { Meta, StoryObj } from '@storybook/react';
+import '@osrd-project/ui-core/dist/theme.css';
+
+import Label from '../components/inputs/Label';
+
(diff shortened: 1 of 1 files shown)


clarani (approved):
LGTM and tested, thanks :pray: :white_check_mark: 

From what I see, the hasHint props only removes the margin bottom of the label. If you want, you can indeed remove the story Hint (but we should keep the prop in the component since it is usefull, see the component `Input` for instance).

kmer2016:
Thank you for this PR. 
As mentionned by @clarani [here](https://github.com/OpenRailAssociation/osrd-ui/pull/272#pullrequestreview-2227123771), If you're okay with that, we'll let you remove the story Hint

PR #276, opened 2024-07-29
ui-icons: fix gauge and signal icons for color

sim51:
Actually `gauge` & `signal` icon doesn't support color changed.
It is due to the SVG that defined the filled color, so it can't be overridden.


SharglutDev (approved):
Lgtm and tested

PR #277, opened 2024-07-29
ui-icons: adding route icon

sim51:
Adding route icon (given by thibaut).
This icon is needed for the  "release constraints on path finding" feature


kmer2016 (approved):
LGTM and tested

PR #323, opened 2024-08-07
core: fix value update on select

SarahBellaha:
Add a useEffect() to update the `Select` component on value change


kmer2016 (approved):
LGTM !

PR #363, opened 2024-08-13
core: fix calendar reference date

### LINKED ISSUE #360: [UI Core] Fix referenceDate Handling in useCalendar Hook to Prevent Past Slot Graying
## Description
Currently, the `useCalendar` hook receives an optional `referenceDate` prop, which defaults to the current date if not provided.
However, this prop is never passed, resulting in `referenceDate` always being set to today's date.
This causes an issue when the `selectedSlot` prop includes a past time window, as the component is entirely grayed out due to the `past` class being applied.

## Acceptance Criteria
- [ ] The referenceDate prop is no longer passed but rather initialized within the useCalendar hook.
 By default, it is set to today's date, but if selectedSlot is provided, the referenceDate is the start date.
- [ ] Adapt useCalendar unit tests.

SarahBellaha:
Closes #360

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -91,30 +91,30 @@
       }
     },
     "node_modules/@babel/compat-data": {
-      "version": "7.25.0",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.25.0.tgz",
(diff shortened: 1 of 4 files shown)


kmer2016 (approved):
LGTM and tested

PR #384, opened 2024-08-21
core: update inputWithSuggestions component and index exports

SarahBellaha:
The inPutWithSuggestions component was missing an export statement.


kmer2016 (approved):
LGTM !

PR #389, opened 2024-08-22
ui-manchette: make the space time chart scale fit vertically

### LINKED ISSUE #313: GET+Manchette: can not see the whole path on proportional mode
In proportional mode, when the component is rendered for the first time, the whole path should be displayed.

The initial zoom level is not correct.

clarani:
closes #313

### DIFF
--- a/ui-manchette/src/components/Manchette.tsx
+++ b/ui-manchette/src/components/Manchette.tsx
@@ -20,11 +20,6 @@ import {
 } from './helpers';
 import OperationalPointList from './OperationalPointList';
 import { useIsOverflow } from '../hooks/useIsOverFlow';
-type ManchetteProps = {
-  operationalPoints: OperationalPointType[];
-  projectPathTrainResult: ProjectPathTrainResult[];
-  selectedProjection?: number;
-};
 import usePaths from '../hooks/usePaths';
 import type {
   OperationalPointType,
@@ -33,10 +28,20 @@ import type {
 } from '../types';
 import { getDiff } from '../utils/vector';
 
+const DEFAULT_HEIGHT = 561;
+
+type ManchetteProps = {
+  operationalPoints: OperationalPointType[];
+  projectPathTrainResult: ProjectPathTrainResult[];
+  selectedProjection?: number;
+  height?: number;
+};
+
 const Manchette: FC<ManchetteProps> = ({
   operationalPoints,
   projectPathTrainResult,
   selectedProjection,
+  height = DEFAULT_HEIGHT,
 }) => {
   const manchette = useRef<HTMLDivElement>(null);
 
@@ -59,7 +64,7 @@ const Manchette: FC<ManchetteProps> = ({
     yOffset: 0,
     panning: null,
     scrollPosition: 0,
-    isProportional: false,
+    isProportional: true,
     operationalPointsChart: [],
     operationalPointsToDisplay: [],
     panY: false,
@@ -115,6 +120,7 @@ const Manchette: FC<ManchetteProps> = ({
   useEffect(() => {
     const computedOperationalPoints = calcOperationalPointsToDisplay(
       operationalPoints,
+      height,
       isProportional,
       yZoom
     );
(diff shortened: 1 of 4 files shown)


ui-manchette/src/components/Manchette.tsx
```diff
@@ -59,7 +64,7 @@ const Manchette: FC<ManchetteProps> = ({
     yOffset: 0,
     panning: null,
     scrollPosition: 0,
-    isProportional: false,
+    isProportional: true,
```
SharglutDev, line 67: We want to be by default in proportional ? Or only the story should be like that ?

ui-manchette/src/components/helpers.ts
```diff
@@ -55,23 +55,33 @@ export const calcOperationalPointsToDisplay = (
...
   operationalPoints: StyledOperationalPointType[],
+  height: number,
   zoom: number,
   isProportional: boolean
-) =>
-  operationalPoints.map((op, index) => {
+) => {
+  if (operationalPoints.length < 2) return [];
+  const from = operationalPoints[0].position;
+  const to = operationalPoints[operationalPoints.length - 1].position;
```
SharglutDev: How would do you feel to start using this syntax ? I think it's more concise. And it forces us to check if the value is undefined so maybe it's more safe ?

```suggestion
  const to = operationalPoints.at(-1)!.position;
```

ui-manchette/src/components/helpers.ts
SharglutDev, line 1: If I test with a lot of ops, I get this in proportional. Maybe something should also be changed in `calcOperationalPointsToDisplay` to prevent overlapping ?

Capture d’écran 2024-08-29 à 10 28 27

ui-manchette/src/components/helpers.ts
```diff
@@ -83,13 +93,28 @@ export const getOperationalPointsWithPosition = (
...
-  return scales;
+export const getScales = (
+  ops: OperationalPoint[],
+  height: number,
+  isProportional: boolean,
+  zoomY: number
+) => {
+  if (ops.length < 2) return [];
+  const from = ops[0].position;
+  const to = ops[ops.length - 1].position;
```
SharglutDev: Same here for `.at(-1)` if you decide to use it.

ui-manchette/src/components/helpers.ts
```diff
@@ -83,13 +93,28 @@ export const getOperationalPointsWithPosition = (
...
+
+  const totalDistance = to - from;
+  const heightWithoutFinalOp = height - FOOTER_HEIGHT - OP_LINE_HEIGHT / 2;
```
SharglutDev, line 114: We use this two times in this file. Should we extract this logic in a function with a param for the last part ? OR do you think it's overkill ?

```suggestion
  const heightWithoutFinalOp = getHeightWithoutFinalOp(OP_LINE_HEIGHT / 2);
```

ui-manchette/src/components/Manchette.tsx
SharglutDev, line 1: Capture d’écran 2024-08-29 à 10 28 27

Doesn't seem to be related to your PR, but if I understand correctly, all the horizontal blue lines we see on the space time chart correspond to ops ? If the correspond ops are hidden, should we still see these lines ?

Also the alignment between an op and the blue line doesn't seem to be very good.

SharglutDev (changes requested):
Thank you for this, left some comments.

ui-manchette/src/components/Manchette.tsx
clarani: We want it by default in proportional (asked by Loïc)

ui-manchette/src/components/helpers.ts
clarani: Yes you're right, it might be better :+1:

ui-manchette/src/components/helpers.ts
clarani: Done

ui-manchette/src/components/Manchette.tsx
clarani: Yes it's not related to this PR, so I leave it for another PR :)

ui-manchette/src/components/helpers.ts
clarani: Good catch, I fixed !

SharglutDev (approved):
Lgtm and tested, good job ! 

We should just create an issue for the alignment problem.

clarani:
Here is the ticket for the not aligned PR: https://github.com/OpenRailAssociation/osrd-ui/issues/396

PR #390, opened 2024-08-22
manchette: fit horizontal scale at first render

### LINKED ISSUE #388: GET+Manchette: the horizontal scale is not fit at the first render

clarani:
closes #388 

See the commits:
1. fit the horizontal scale at first render
2. add the horizontal zoom on scroll + shift

### DIFF
--- a/ui-manchette/src/components/Manchette.tsx
+++ b/ui-manchette/src/components/Manchette.tsx
@@ -1,4 +1,4 @@
-import React, { type FC, useCallback, useEffect, useRef, useState } from 'react';
+import React, { type FC, useCallback, useEffect, useRef, useState, useMemo } from 'react';
 
 import { ZoomIn, ZoomOut } from '@osrd-project/ui-icons';
 import { SpaceTimeChart, PathLayer } from '@osrd-project/ui-spacetimechart';
@@ -14,9 +14,11 @@ import {
 } from './consts';
 import {
   calcOperationalPointsToDisplay,
+  computeTimeWindow,
   getOperationalPointsWithPosition,
   getScales,
   operationalPointsHeight,
+  zoomX,
 } from './helpers';
 import OperationalPointList from './OperationalPointList';
 import { useIsOverflow } from '../hooks/useIsOverFlow';
@@ -44,6 +46,8 @@ const Manchette: FC<ManchetteProps> = ({
   height = DEFAULT_HEIGHT,
 }) => {
   const manchette = useRef<HTMLDivElement>(null);
+
+  const [isShiftPressed, setIsShiftPressed] = useState(false);
 
   const [state, setState] = useState<{
     xZoom: number;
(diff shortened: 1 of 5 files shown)


ui-manchette/src/components/helpers.ts
```diff
@@ -73,6 +82,29 @@ export const operationalPointsHeight = (
...
 
+export const computeTimeWindow = (trains: ProjectPathTrainResult[]) => {
+  const { minTime, maxTime } = trains.reduce(
+    (times, train) => {
+      if (train.space_time_curves.length === 0) return times;
+
+      const lastCurve = train.space_time_curves[train.space_time_curves.length - 1];
+      if (lastCurve.times.length === 0) return times;
+
+      const firstPoint = +new Date(train.departure_time);
```
SharglutDev: If I remember correctly, elsewhere in the app you had a comment between using `+` or `number` and it had been decided `Number` was better ?

```suggestion
      const firstPoint = Number(new Date(train.departure_time));
```

ui-manchette/src/components/helpers.ts
```diff
@@ -93,3 +125,18 @@ export const getScales = (ops: OperationalPoint[], isProportional: boolean, zoom
...
+  { delta, position: { x } }: Parameters<NonNullable<SpaceTimeChartProps['onZoom']>>[0]
+) => {
+  const xZoom = Math.min(Math.max(currentXZoom * (1 + delta / 10), MIN_ZOOM_X), MAX_ZOOM_X);
+  return {
+    xZoom,
+    // Adjust zoom level relatively to the input delta value:
+    // These lines are here to center the zoom on the mouse position:
+    xOffset: x - ((x - currentXOffset) / currentXZoom) * xZoom,
+  };
+};
```
SharglutDev, line 175: Wasn't it possible to use the `zoom` function from spaceTimeChart ? (maybe with a little refacto to use only one of the two axis) ?

ui-manchette/src/components/Manchette.tsx
```diff
@@ -81,19 +85,34 @@ const Manchette: FC<ManchetteProps> = ({
...
+
   const zoomYIn = useCallback(() => {
     if (yZoom < MAX_ZOOM_Y) setState((prev) => ({ ...prev, yZoom: yZoom + ZOOM_Y_DELTA }));
   }, [yZoom]);
   const zoomYOut = useCallback(() => {
     if (yZoom > MIN_ZOOM_Y) setState((prev) => ({ ...prev, yZoom: yZoom - ZOOM_Y_DELTA }));
   }, [yZoom]);
   const handleScroll = useCallback(() => {
-    if (manchette.current) {
+    if (!isShiftPressed && manchette.current) {
```
SharglutDev, line 107: Could you add some spacing between all these `useCallback` ?

SharglutDev, line 107: I don't use a mouse. What happens if you don't put the `!isShiftPressed` ?

ui-manchette/src/components/helpers.ts
```diff
@@ -73,6 +82,29 @@ export const operationalPointsHeight = (
...
+  const { minTime, maxTime } = trains.reduce(
+    (times, train) => {
+      if (train.space_time_curves.length === 0) return times;
+
+      const lastCurve = train.space_time_curves[train.space_time_curves.length - 1];
+      if (lastCurve.times.length === 0) return times;
+
+      const firstPoint = +new Date(train.departure_time);
+      const lastPoint =
+        +new Date(train.departure_time) + lastCurve.times[lastCurve.times.length - 1];
```
SharglutDev: We can use the `.at(-1)` if you don't think it's bothersome with the typing

SharglutDev (changes requested):
Thank you for the PR, left some comments

ui-manchette/src/components/helpers.ts
clarani: Yes, done :+1:

ui-manchette/src/components/helpers.ts
clarani: No unfortunately, we cannot since we have separated modules and this function zoomX is not exported from ui-space-time-chart (on purpose)

ui-manchette/src/components/helpers.ts
```diff
@@ -1,10 +1,19 @@
...
+import { BASE_KM_HEIGHT, BASE_OP_HEIGHT, MAX_TIME_WINDOW, MAX_ZOOM_X, MIN_ZOOM_X } from './consts';
+import type {
+  OperationalPointType,
+  ProjectPathTrainResult,
+  StyledOperationalPointType,
+} from '../types';
 
 export const positionMmToKm = (position: number) => Math.round((position / 1000000) * 10) / 10;
 
+export const msToS = (time: number) => time / 1000;
```
Yohh: shouldn't it be in a `utils.ts` ? (same for `positionMmToKm`)

SharglutDev (approved):
Lgtm and tested

ui-manchette/src/components/helpers.ts
clarani: Yes you're right, I move them there :+1:

Yohh:
if I hover the chart without clicking on it then shift+scroll previewKeyDown isn't triggered, you could maybe add a `mouseover` listener to enable it

clarani:
I suspect that the problem comes from the iframe and that we should not encounter it when integrating. Can we test like this and open a bug ticket if the problem is not solved ?

PR #397, opened 2024-08-29
manchette: drop extraneous scroll event listener

emersion:
We already listen for scroll events on the manchette div. The listener performs actions which depend on the manchette div scroll position only. No need to listen for window scroll.


theocrsb (approved):
LGTM

clarani (approved):
Thanks :pray: 
LGTM :white_check_mark:

PR #398, opened 2024-08-29
ui-manchette: fix space-time-chart proportional scale

clarani:
The proportional scale was too long (see the bottom of the GET).
image

After the fix, the proportional scale fits the the whole path
image


SharglutDev (approved):
Lgtm and tested, thank you for the fix !

Yohh (approved):

PR #420, opened 2024-09-02
setup: ban FC type in eslintrc

### LINKED ISSUE #411: setup: add an eslint rule to ban the React.FC
### Description and goal

To harmonize the code base, we ban the React.FC

### Acceptance criteria

No React.FC in the code base anymore

Yohh:
- add rules to ban FC, react.FC, react.FunctionComponent & react.FunctionalComponent

close #411

### DIFF
--- a/ui-core/src/components/inputs/datePicker/Calendar.tsx
+++ b/ui-core/src/components/inputs/datePicker/Calendar.tsx
@@ -20,13 +20,7 @@ type DayProps = {
   onClick: (date: Date) => void;
 };
 
-const Day: React.FC<DayProps> = ({
-  date,
-  isReferenceDate,
-  isSelectable,
-  dayWrapperClassName,
-  onClick,
-}) => (
+const Day = ({ date, isReferenceDate, isSelectable, dayWrapperClassName, onClick }: DayProps) => (
   <div
     onClick={() => {
       if (isSelectable) onClick(date);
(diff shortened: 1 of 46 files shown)


SharglutDev (approved):
Lgtm, thank you for this PR.

I know it's a pain but don't you think we should extract the props in a type where there is more than one ?

Yohh:
> Lgtm, thank you for this PR.
> 
> I know it's a pain but don't you think we should extract the props in a type where there is more than one ?

I was wondering myself, it's related to the stories right? We should know why it has been written this way I think.

SharglutDev:
> > Lgtm, thank you for this PR.
> > I know it's a pain but don't you think we should extract the props in a type where there is more than one ?
> 
> I was wondering myself, it's related to the stories right? We should know why it has been written this way I think.

Yep and for the Warped Map as well

ui-core/src/components/inputs/datePicker/Calendar.tsx
```diff
@@ -40,7 +34,7 @@ const Day: React.FC<DayProps> = ({
   </div>
 );
 
-const Calendar: React.FC<CalendarProps> = (props) => {
+const Calendar = (props: CalendarProps) => {
   const { displayedMonthStartDate, onDayClick } = props;
```
Uriel-Sautron:
```suggestion
const Calendar = ({ displayedMonthStartDate, onDayClick } : CalendarProps) => {
```

Uriel-Sautron (approved):
LGTM !! Thanks

ui-core/src/components/inputs/datePicker/Calendar.tsx
Yohh: done

Yohh:
> > > Lgtm, thank you for this PR.
> > > I know it's a pain but don't you think we should extract the props in a type where there is more than one ?
> > 
> > 
> > I was wondering myself, it's related to the stories right? We should know why it has been written this way I think.
> 
> Yep and for the Warped Map as well

done

SharglutDev (approved):
Lgtm, thank you

PR #430, opened 2024-09-04
Fix speed space chart round value to decimal

### LINKED ISSUE #8297: GEV: distance of ticks always display useless `.0` and the cursor distance doesn't show 100 meters
### What happened?

There is a useless 100m precision on the distance ticks because they are rounded at 1km.
However the cursor distance indicator is also rounded at 1km but it should not be.
image

The `.0` on the ticks should be removed according to the mockup.
And the cursor distance must be able to display a precision of 100m.
image

.

Recette (SNCF)

Google Chrome

0a34296

flomonster:
Fix the rounding to decimal instead of unit.

image

close https://github.com/OpenRailAssociation/osrd/issues/8297

### DIFF
--- a/ui-speedspacechart/src/components/helpers/drawElements/reticle.ts
+++ b/ui-speedspacechart/src/components/helpers/drawElements/reticle.ts
@@ -261,11 +261,11 @@ export const drawCursor = ({ ctx, width, height, store }: DrawFunctionParams) =>
       ctx.lineTo(curveX + MARGIN_LEFT, height - MARGIN_BOTTOM + 6);
 
       // calculate the position value between x.a and x.b
-      const position = Math.round(
+      const position =
         ((curveX - x.a) / (x.b - x.a)) *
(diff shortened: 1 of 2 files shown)


Yohh (commented):
thanks for this PR

ok for the rounded adjustment

https://github.com/OpenRailAssociation/osrd/issues/8297 can't be closed, `.0` is still possible, we should add a condition to avoid `toFixed()` method if `roundedPosition` doesn't have a decimal

flomonster:
> [OpenRailAssociation/osrd#8297](https://github.com/OpenRailAssociation/osrd/issues/8297) can't be closed, `.0` is still possible, we should add a condition to avoid `toFixed()` method if `roundedPosition` doesn't have a decimal

Good point, I also fixed it (in a second commit).

> [!IMPORTANT]
> This fix expose a new bug. The reticle is slightly offset from the graduations.

image

Yohh (approved):
LGTM

Math-R (approved):
lgtm 👍🏻

PR #436, opened 2024-09-06
Code refactoring speed space chart tick X axis

flomonster:
Main changes :
1. MaxPositionValues function rationalized.
2. Iterate on ticks instead of position in `speeds`.
3. Modify the way we handle scaling.

This corrects a few problems:
- X axis of ticks no longer disappears when zoomed in.
- Tick position is more accurate
- Having a simplified version of `speeds` doesn't break everything.

### DIFF
--- a/ui-speedspacechart/src/components/helpers/drawElements/tickX.ts
+++ b/ui-speedspacechart/src/components/helpers/drawElements/tickX.ts
@@ -5,7 +5,7 @@ import { clearCanvas, maxPositionValue } from '../../utils';
 const { MARGIN_LEFT, MARGIN_RIGHT, MARGIN_BOTTOM, CURVE_MARGIN_SIDES } = MARGINS;
 
 export const drawTickX = ({ ctx, width, height, store }: DrawFunctionParams) => {
-  const { speeds, ratioX, leftOffset, cursor } = store;
+  const { ratioX, leftOffset, cursor } = store;
 
   clearCanvas(ctx, width, height);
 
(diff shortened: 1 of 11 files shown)


ui-speedspacechart/src/components/helpers/drawElements/tickX.ts
```diff
@@ -17,65 +17,68 @@ export const drawTickX = ({ ctx, width, height, store }: DrawFunctionParams) =>
...
 
-  const { maxPosition, RoundMaxPosition, intermediateTicksPosition } = maxPositionValues(store);
+  const maxPosition = maxPositionValue(store);
+  const scaleRatio = (maxPosition * 10) / (width * store.ratioX);
+  // Tick scale is the distance between two ticks.
+  // When the scale ratio is important we want to have ticks scale that are multiple of 5.
+  let tickScale = Math.floor(scaleRatio) * 5;
+  // When the scale ratio is small we want one of the following values: 4, 2, 1, 0.5, 0.25.
+  if (scaleRatio < 1) {
+    tickScale = 4 / Math.pow(2, Math.min(Math.floor(0.5 / scaleRatio), 4));
```
Yohh: could you divide this line with explicit variables instead of nesting methods?

ui-speedspacechart/src/components/helpers/drawElements/tickX.ts
```diff
@@ -17,65 +17,68 @@ export const drawTickX = ({ ctx, width, height, store }: DrawFunctionParams) =>
...
-  speeds.forEach((_, i) => {
-    ctx.moveTo(MARGIN_LEFT + xPosition * i + CURVE_MARGIN_SIDES / 2, height - MARGIN_BOTTOM);
-    ctx.lineTo(
-      MARGIN_LEFT + xPosition * i + CURVE_MARGIN_SIDES / 2,
-      height - MARGIN_BOTTOM + CURVE_MARGIN_SIDES / 2
-    );
+      const distanceLeftBorder = Math.abs(
+        MARGIN_LEFT + CURVE_MARGIN_SIDES / 2 - leftOffset - positionX
+      );
+      let opacityLeftBorder = Math.max(Math.min(distanceLeftBorder / 45 - 0.1, 1.0), 0.0);
```
Yohh: needs some explanations, 45 can be set in a variable, `0.1, 1.0), 0.0)` should be clarified

Yohh (commented):
nice! thanks for this great refacto!

a few comments to clarify some nebulous parts

ui-speedspacechart/src/components/helpers/drawElements/tickX.ts
flomonster: This code needs to be rewritten. @thibautsailly will provide a matrix to explain how graduation should work depending on the zoom level.

flomonster: I have highly simplified this part. It will be easy to adapt this part with a real mockup later.

ui-speedspacechart/src/components/helpers/drawElements/tickX.ts
flomonster: Done

ui-speedspacechart/src/components/helpers/drawElements/tickX.ts
```diff
@@ -17,65 +17,83 @@ export const drawTickX = ({ ctx, width, height, store }: DrawFunctionParams) =>
...
+  let principleTickFrequency = 5;
+  if (windowLength >= 200) {
+    tickScale = Math.floor(windowLength / 200) * 5;
+    principleTickFrequency = 2;
+  } else if (windowLength >= 50) {
+    tickScale = 2;
+    principleTickFrequency = 5;
+  } else if (windowLength >= 20) {
+    tickScale = 1;
+    principleTickFrequency = 5;
```
SharglutDev, line 34: Do we need to re-set `principleTickFrequency` to 5 even though it's initialized with this value on l.25 ?

flomonster: This is not required. However, this value is very much tied to `tickScale`, more than to the condition. That's why I like them next to each other. If you modify one of the values, it's very likely that the other will also be modified.

In rust I would have written it like that:

```rust
let (tickScale, principleTickFrequency) = if windowLength >= 200 {
      ((windowLenght / 200.0).floor() * 5.0, 2)
  } else if windowLenght >= 50 {
      (2.0, 5)
  } else if windowLenght >= 20 {
      (1.0, 5)
  } else {
      (0.1, 5)
  };
```

Do you think there's a way to write this in TS that matches the code above?

Yohh: exactly like you did, you could also declare an object or a tupple but it seems over-engineered to me

in this specific case, the repetition seems useless for the logic itself, and even if your point is understandable it's not a concept usually used in TS

Yohh: to get the best of two worlds:

```js
let tickScale: number;
let principleTickFrequency: number;
if (windowLength >= 200) {
    tickScale = Math.floor(windowLength / 200) * 5;
    principleTickFrequency = 2;
} else if (windowLength >= 50) {
    tickScale = 2;
    principleTickFrequency = 5;
} else if (windowLength >= 20) {
    tickScale = 1;
    principleTickFrequency = 5;
} else {
    tickScale = 0.1;
    principleTickFrequency = 5;
}
```

Yohh (approved):
LGTM

SharglutDev (approved):
Lgtm and tested, great job !

PR #449, opened 2024-09-10
use the wright icon for the dropdown Select

### LINKED ISSUE #448: Update Select Dropdown Icon to Match Mockup Design
### Description and goal

Fix the icon used for the Select Dropdown. In the mockup, the icon displays two arrows, whereas the current implementation uses a single arrow.

Current implementation
image

Mockup
image

### Acceptance criteria

The icon for Select Dropdown should match the one in the mockup.

kmer2016:
close #448


ui-core/src/styles/select.css
```diff
@@ -5,11 +5,15 @@
...
-  padding: 0.438rem 0.813rem 0.562rem 0.813rem;
+  padding: 0.438rem 2rem 0.563rem 0.813rem;
   letter-spacing: 0;
   text-align: left;
   line-height: 1.5rem;
   width: 100%;
+  -webkit-appearance: none;
+  -moz-appearance: none;
+  appearance: none;
+  background: inline('../assets/triangle-up-down.svg') no-repeat right .75rem center;
```
SharglutDev, line 16: Related to this [comment](https://github.com/OpenRailAssociation/osrd/pull/8804#discussion_r1751450325), maybe there is more to add to have the icon also when focused or else.

kmer2016: To avoid any other kind of conflict, I preferred to rename the class.

ui-core/src/styles/select.css
```diff
@@ -1,15 +1,19 @@
...
```
SharglutDev, line 16: This works fine but it's not what we use in the rest of the app. I'm afraid it might be difficult if we need to sometime hide it, focus it for accessibility etc. As discussed, I would like to discuss it together to understand what are the advantages :)

ui-core/src/styles/select.css
```diff
@@ -1,15 +1,19 @@
-.custom-select {
+.osrd-ui-custom-select {
   border-radius: 4px;
   border: 1px solid rgba(148, 145, 142, 1);
   box-shadow:
     0px 1px 1px rgba(0, 0, 0, 0.2),
     0px 2px 4px rgba(0, 0, 0, 0.1);
   height: 2.5rem;
-  padding: 0.438rem 0.813rem 0.562rem 0.813rem;
+  padding: 0.438rem 2rem 0.563rem 0.813rem;
```
SharglutDev: If I remove everything for just a `padding-left: 0.813rem;`, the position stays unchanged. Do we need all this configuration (I agree it matches the mockup, but it matches it also without it)

SharglutDev (approved):
Lgtm and tested :)

PR #450, opened 2024-09-10
Code refactoring speed space chart tick Y axis

flomonster:
Main changes :
1. Iterate on ticks instead of position in `speeds`.
2. The scale is adapting with the max speed of the chart

This corrects a few problems:
- Axis Y is not visually overloaded even with a top speed of 300km/h
- The Y axis is not visually empty even with a maximum speed of 30km/h
- Having a simplified version of `speeds` doesn't break everything.

### Before

image

### After

image


Yohh (commented):
I'm skeptical about removing major lines, the mockup clearly shows them, the only difference is about the max speed to display
Capture d’écran 2024-09-11 à 06 37 29
did you talk about it with Thibaut?

and if we keep this change, `drawMajorGridY()` should become `drawGridY()`

flomonster:
> I'm skeptical about removing major lines, the mockup clearly shows them, the only difference is about the max speed to display

I misinterpreted the mock-up. I'll put the major lines back.

ui-speedspacechart/src/components/helpers/drawElements/tickY.ts
```diff
@@ -4,44 +4,49 @@ import { clearCanvas, speedRangeValues } from '../../utils';
...
-      if (textPositionY > height - MARGIN_BOTTOM - 30)
-        opacity = (10 - (height - MARGIN_BOTTOM - 30 - textPositionY) / 3) / 10;
-
-      ctx.fillStyle = `rgba(182, 179, 175, ${opacity})`;
-      ctx.fillText(text, textOffsetX, textPositionY);
-    }
-  });
+  for (let i = 0; i <= nbTicks; i++) {
+    const positionY = height - MARGIN_BOTTOM - ticksOffset * i;
+    ctx.moveTo(42, positionY);
```
SharglutDev: To avoid the magic number, maybe we can add a `const TICK_WIDTH = 6;` and then 

```suggestion
    ctx.moveTo(MARGIN_LEFT - TICK_WIDTH, positionY);
```

ui-speedspacechart/src/components/helpers/drawElements/tickY.ts
```diff
@@ -4,44 +4,49 @@ import { clearCanvas, speedRangeValues } from '../../utils';
...
+    ctx.lineTo(MARGIN_LEFT, positionY);
+    ctx.textAlign = 'right';
+    const text = (i * tickScale).toString();
+    const textPositionY = positionY + 4;
+
+    ctx.fillStyle = `rgb(182, 179, 175 )`;
```
SharglutDev: Won't change much but the color on mockup seems to be : 

Capture d’écran 2024-09-12 à 12 01 52

ui-speedspacechart/src/components/helpers/drawElements/tickY.ts
```diff
@@ -4,44 +4,49 @@ import { clearCanvas, speedRangeValues } from '../../utils';
...
 
   clearCanvas(ctx, width, height);
 
-  const { maxSpeed } = speedRangeValues(store);
-
   ctx.strokeStyle = 'rgb(121, 118, 113)';
   ctx.lineWidth = 0.5;
   ctx.font = 'normal 12px IBM Plex Sans';
   ctx.fillStyle = 'rgb(182, 179, 175)';
   const textOffsetX = 36;
```
SharglutDev: This can be moved outside of the function

ui-speedspacechart/src/components/helpers/drawElements/gridY.ts
```diff
@@ -1,60 +1,31 @@
...
-
   clearCanvas(ctx, width, height);
 
   const { maxSpeed } = speedRangeValues(store);
+  const tickScale = tickScaleY(maxSpeed);
+  const nbTicks = Math.ceil(maxSpeed / tickScale);
+  const maxTickSpeed = nbTicks * tickScale;
+  const ratioRoundPositions = maxTickSpeed / maxSpeed;
+  const ticksOffset =
+    ((height - MARGIN_BOTTOM - MARGIN_TOP - CURVE_MARGIN_TOP) * ratioRoundPositions) / nbTicks;
```
SharglutDev: This block of code (and l.12) is used in both `gridY.ts` and `tickY.ts`. We could create a function in `utils.ts` that returns `tickScale`, `nbTicks` and `ticksOffset`.

ui-speedspacechart/src/components/layers/AxisLayerY.tsx
```diff
@@ -4,18 +4,18 @@ import type { Store } from '../../types/chartTypes';
 import { drawGridY } from '../helpers/drawElements/gridY';
 import { useCanvas } from '../hooks';
 
-type AxisLayerYProps = {
+type GridYProps = {
```
SharglutDev: We try to keep the props type name the same as the component. We should either rename the component `GridY` to match the type and the id or keep these as they were before.

SharglutDev (commented):
Thank you for the PR, left a few comments

ui-speedspacechart/src/components/helpers/drawElements/gridY.ts
flomonster: I was thinking about merging these files into one. What do you think?

SharglutDev: As both are highly related, I think it should be ok. But as it will become one layer (is that what you meant ?), maybe we should check with @thibautsailly that he doesn't plan later to make the "horizontal lines" an optional option of the speedspace chart.

flomonster: Actually, the plan is to make vertical line optionnal if OPs (steps) are disable. So it makes a lot of sense to have it in the same layer.

ui-speedspacechart/src/components/helpers/drawElements/tickY.ts
flomonster: Good catch it was a typo I think :laughing:

ui-speedspacechart/src/components/helpers/drawElements/tickY.ts
clarani: Move this const outside the function and use screaming case

ui-speedspacechart/src/components/helpers/drawElements/gridY.ts
flomonster: I merged the two layers. We can always adjust if necessary in the future.

clarani (commented):
LGTM :white_check_mark:

ui-speedspacechart/src/components/helpers/drawElements/tickY.ts
clarani: Not relevant anymore

ui-speedspacechart/src/components/helpers/drawElements/tickY.ts
```diff
@@ -1,52 +1,90 @@
...
+  }
   ctx.stroke();
 
-  // prevent overlapping with margin top
+  // Prevent overlapping with margin top
   ctx.clearRect(0, 0, width, MARGIN_TOP);
   ctx.clearRect(MARGIN_LEFT - 6, height - MARGIN_BOTTOM, width, MARGIN_BOTTOM);
   ctx.clearRect(0, height - MARGIN_BOTTOM + 6, MARGIN_LEFT, MARGIN_BOTTOM);
+
+  ctx.fillStyle = 'rgb(182, 179, 175)';
```
SharglutDev:
```suggestion
  ctx.fillStyle = 'rgb(182, 178, 175)';
```

SharglutDev (approved):
Lgtm and tested, nice improvement !

ui-speedspacechart/src/components/helpers/drawElements/gridY.ts
Yohh: the purpose of having two isolated layers was to facilitate the debug (as for each other layer), but yes it seems ok.

also, this new layer should be named `AxisLayerY.tsx` instead of `TickLayerY.tsx`

Yohh: and `tickY.ts` -> `gridY.ts` 

edit: `axisY.ts` seems better, so we could maybe group every `<bla bla>X.ts` in a `axisX` folder and `axisY.ts` & `tickYRight.ts` in a `axisY` folder to improve visibility and navigation.

flomonster: I agree, we have to rename some files to improve visibility and consistency.


> the purpose of having two isolated layers was to facilitate the debug (as for each other layer), but yes it seems ok.

Both layers need the same information. Separating them has the disadvantage of duplicating the code, or requiring a large number of help functions that reduce the readability of the drawing functions.

ui-speedspacechart/src/components/helpers/drawElements/tickY.ts
flomonster: Done

PR #455, opened 2024-09-10
Speed space chart: several improvements of the reticle, steps and crosshair

flomonster:
Main changes :
1. Gather steps rectangle, lines and title together
2. Match style with mock-up (color, size, opacity)
3. Vertical line can be visible without step name
4. Step snapping matches the mock-up
5. Add vetical and horizontal lines from crosshair to axis

Fixes:
- Steps with duplicate positions are now draw only once.
- Fix reticle shown position
- Fix snapped step position
- Snapped steps have their text hidden

image


ui-speedspacechart/src/components/utils.ts
```diff
@@ -266,24 +266,49 @@ export const createSvgBlobUrl = (svgString: string): string => {
...
 ) =>
-  position.start * ((width - MARGINS.CURVE_MARGIN_SIDES) / maxPosition) * ratioX +
-  MARGINS.CURVE_MARGIN_SIDES / 2;
+  (position / maxPosition) *
+    ((width - MARGINS.MARGIN_LEFT - MARGINS.MARGIN_RIGHT - MARGINS.CURVE_MARGIN_SIDES) * ratioX) +
+  MARGINS.CURVE_MARGIN_SIDES / 2 +
+  MARGINS.MARGIN_LEFT +
+  leftOffset;
+
+/** Interpole on a straight line giving two points and the x-axis value */
```
clarani:
```suggestion
/** Interpole on a straight line given two points and the x-axis value */
```

ui-speedspacechart/src/components/utils.ts
```diff
@@ -266,24 +266,49 @@ export const createSvgBlobUrl = (svgString: string): string => {
...
-  maxPosition: number
+  leftOffset: number = 0
 ) =>
-  position.start * ((width - MARGINS.CURVE_MARGIN_SIDES) / maxPosition) * ratioX +
-  MARGINS.CURVE_MARGIN_SIDES / 2;
+  (position / maxPosition) *
+    ((width - MARGINS.MARGIN_LEFT - MARGINS.MARGIN_RIGHT - MARGINS.CURVE_MARGIN_SIDES) * ratioX) +
+  MARGINS.CURVE_MARGIN_SIDES / 2 +
+  MARGINS.MARGIN_LEFT +
+  leftOffset;
```
clarani: Maybe it would be easier to read like this
```suggestion
  const xWidth = width - MARGINS.MARGIN_LEFT - MARGINS.MARGIN_RIGHT - MARGINS.CURVE_MARGIN_SIDES;
  const leftMargin = MARGINS.CURVE_MARGIN_SIDES / 2 + MARGINS.MARGIN_LEFT + leftOffset;
  return (position / maxPosition) * (xWidth * ratioX) + leftMargin;
```

ui-speedspacechart/src/components/utils.ts
```diff
@@ -266,24 +266,49 @@ export const createSvgBlobUrl = (svgString: string): string => {
...
-export const getStopPosition = (
-  position: LayerData<string>['position'],
+/** Transform a position in km into a position on the x-axis in pixels */
+export const positionToPosX = (
+  position: number,
+  maxPosition: number,
   width: number,
   ratioX: number,
-  maxPosition: number
+  leftOffset: number = 0
```
Yohh:
```suggestion
  leftOffset = 0
```

ui-speedspacechart/src/components/helpers/drawElements/gridX.ts
```diff
@@ -1,63 +1,114 @@
...
+const STEP_WIDTH = 1;
+const STEP_RADIUS = 1.0;
+
+/** Filter stops to avoid overlapping draw */
+export const filterStops = (
+  stops: LayerData<string>[],
+  ratioX: number,
+  width: number,
+  maxPosition: number,
+  offset: number = 8
```
Yohh:
```suggestion
  offset = 8
```

ui-speedspacechart/src/components/layers/GridX.tsx
```diff
@@ -4,13 +4,13 @@ import type { Store } from '../../types/chartTypes';
...
 
-type AxisLayerXProps = {
+type GridXProps = {
   width: number;
   height: number;
   store: Store;
 };
 
-const AxisLayerX = ({ width, height, store }: AxisLayerXProps) => {
+const GridX = ({ width, height, store }: GridXProps) => {
```
Yohh: `GridX.tsx` contains lines and steps, we can keep `AxisLayerX.tsx`

ui-speedspacechart/src/components/helpers/drawElements/reticle.ts
```diff
@@ -18,6 +21,49 @@ const {
...
+  const { ratioX, leftOffset } = store;
+  const maxPosition = maxPositionValue(store);
+  const cursorPosition =
+    ((cursorX - leftOffset - CURVE_MARGIN_SIDES / 2) * maxPosition) /
+    ((width - MARGIN_LEFT - MARGIN_RIGHT - CURVE_MARGIN_SIDES) * ratioX);
+  // Clamp the cursor position to the graph bounds
+  return Math.min(Math.max(0, cursorPosition), maxPosition);
+};
+
+export const getSnappedStop = (cursorX: number, width: number, store: Store) => {
```
Yohh: should be in `utils.ts` and tested

Yohh (commented):
nice! thanks for this

need more time to properly review the reticle

`getStopPosition` doesn't pass in tests check

Yohh:
are the two horizontal lines on the curves pointing to the left axis refined and confirmed by @thibautsailly ?

image

flomonster:
> are the two horizontal lines on the curves pointing to the left axis refined and confirmed by @thibautsailly ?

It's part of the mock-up:

image

Yohh:
> > are the two horizontal lines on the curves pointing to the left axis refined and confirmed by @thibautsailly ?
> 
> It's part of the mock-up:

yes I saw it, but the mockup is versioned, so what we can see at first view could change at anytime.

ui-speedspacechart/src/components/layers/GridX.tsx
flomonster: It doesn't contains the ticks. It feels weird to named it axis since it's more a layer over the x axis.

flomonster:
I've moved part of this PR into another ([here](https://github.com/OpenRailAssociation/osrd-ui/pull/480)) for easier review.

ui-speedspacechart/src/components/utils.ts
```diff
@@ -266,24 +267,133 @@ export const createSvgBlobUrl = (svgString: string): string => {
...
+export const clamp = (value: number, min: number, max: number) =>
+  Math.min(Math.max(value, min), max);
+
+/** Filter stops to avoid overlapping draw */
+export const filterStops = (
+  stops: LayerData<string>[],
+  ratioX: number,
+  width: number,
+  maxPosition: number,
+  offset = 8
```
jacomyal: I think `offset` is a confusing name here, as it does not describe an offset, but the minimal space between two consecutive displayed stops. May be `minSpace` would be better.

ui-speedspacechart/src/components/common/DetailsBox.tsx
jacomyal, line 1: They were already there before this PR, but there are a lot of "magic numbers" in that file, and some of them look like they depend on each other. It might be better to store them in constant variables, with names that explain what these numbers correspond to on screen.

ui-speedspacechart/src/components/helpers/drawElements/gridX.ts
jacomyal, line 1: There are now three different colors hard-coded within the rendering code. It would be nice to have them in constant variables instead, with names describing what they correspond to precisely.

ui-speedspacechart/src/components/helpers/drawElements/reticle.ts
```diff
@@ -79,225 +70,179 @@ export const drawCursor = ({ ctx, width, height, store }: DrawFunctionParams) =>
...
 
-    // Get the power restriction code based on the position of the cursor
-    if (powerRestrictions) {
-      const currentPowerRestriction = powerRestrictions.find(
-        ({ position }) =>
-          cursor.x! >= xPositionReference(position.start) &&
-          cursor.x! <= xPositionReference(position.end!)
-      );
-      powerRestrictionText = currentPowerRestriction?.value.powerRestriction || '';
+  // Check if cursor is snappping to stop
```
jacomyal: Typo (snapp**p**ing)

ui-speedspacechart/src/components/helpers/drawElements/reticle.ts
```diff
@@ -79,225 +70,179 @@ export const drawCursor = ({ ctx, width, height, store }: DrawFunctionParams) =>
...
-        ({ position }) =>
-          xPositionReference(position.start) <= cursor.x! &&
-          xPositionReference(position.end!) >= cursor.x!
-      )?.value.electricalProfile;
-
-      if (electricalProfileValue) {
-        electricalProfileText = electricalProfileValue;
-      }
-    }
+  let cursorPosition = getCursorPosition(cursor.x!, width, store);
```
jacomyal: The `!` is not necessary here, because of the previous early escape on `cursor.x === null`. Normally, TypeScript should "understands" that here `cursor.x` is necessarily a number. This comment also applies for the next occurrences of `cursor.x!` in this function.

thibautsailly:
> are the two horizontal lines on the curves pointing to the left axis refined and confirmed by @thibautsailly ?
> 
> image

Confirmed

ui-speedspacechart/src/components/helpers/drawElements/reticle.ts
flomonster: Done. TypeScript is not able to detect that the value not null when used in the lambda function. Check [here](https://github.com/OpenRailAssociation/osrd-ui/pull/455/files#diff-60f8e6d4b91b5be2ea8787dbc4ae1dcb69c6d99d9f6876a6a1fb8c90833b47dbR99-R100)

ui-speedspacechart/src/components/helpers/drawElements/gridX.ts
flomonster: I reduced the number of magic numbers. 

Is it okay to have magic number as proportion of a constant size?
Example: `SNAPPED_CURSOR_HEIGHT * 0.6`

ui-speedspacechart/src/components/common/DetailsBox.tsx
flomonster: I reduced the number of magic numbers. 

Is it okay to have magic number as proportion of a constant size?
Example: `SNAPPED_CURSOR_HEIGHT * 0.6`

ui-speedspacechart/src/components/helpers/drawElements/gridX.ts
flomonster: Fixed. I didn't know how to handle different level of opacity for a same color properly.

I'm crafting colors using format string to add opacity value. I don't think it's very readable. Any advice?

ui-speedspacechart/src/components/layers/GridX.tsx
Yohh: I still have a problem with that:
- every layers have the word `layer` in their name, except `GridX.tsx`
- `TickLayerY.tsx` contains `drawAxisY` that draws ticks, text, minor and major lines.
- `TickLayerX.tsx` contains `drawTickX` that draws only ticks and texts.
- `GridX.tsx` contains `drawGridX` that draws lines, steps and texts.

we're losing clarity, we should keep consistent names, maybe add folders for axis logic to keep an optimal readability and navigation

ui-speedspacechart/src/components/helpers/drawElements/reticle.ts
```diff
@@ -79,225 +73,185 @@ export const drawCursor = ({ ctx, width, height, store }: DrawFunctionParams) =>
...
+  if (predecessorSpeedIndex === speeds.length - 1) predecessorSpeedIndex--;
+
+  let predecessorEcoSpeedIndex = binarySearch(
+    ecoSpeeds,
+    cursorPosition,
+    (element: LayerData<number>) => element.position.start
+  );
+  if (predecessorEcoSpeedIndex === ecoSpeeds.length - 1) predecessorEcoSpeedIndex--;
+
+  const speed1 = speeds[predecessorSpeedIndex];
```
Caracol3: Maybe rename those two const for more clarity. 
'prevSpeed' and 'nextSpeed' ?

ui-speedspacechart/src/components/helpers/drawElements/reticle.ts
```diff
@@ -79,225 +73,185 @@ export const drawCursor = ({ ctx, width, height, store }: DrawFunctionParams) =>
...
+    speed1.value,
+    speed2.position.start,
+    speed2.value,
+    cursorPosition
+  );
+  speedText = speedValue.toFixed(1);
+  const baseReticleY =
+    cursorBoxHeight - (speedValue / maxSpeed) * (cursorBoxHeight - CURVE_MARGIN_TOP) + MARGIN_TOP;
+
+  const ecoSpeed1 = ecoSpeeds[predecessorEcoSpeedIndex];
```
Caracol3: Same as above.

ui-speedspacechart/src/components/const.ts
```diff
@@ -63,6 +65,13 @@ export const LAYERS_SELECTION: Array<keyof Store['layersDisplay']> = [
   'speedLimitTags',
 ];
 
+// Colors
+
+export const BLACK = '#000000';
+export const GREY_50 = '#797671';
+export const GREY_80 = '#312e2b';
+export const LIGHT_BLUE = '#2170b9';
```
Yohh: we usually use rgba() instead of hexa in OSRD-UI, is there a reason to use hexa in this pr?

Caracol3:
Is it a known issue ?
In the example, when zooming in, we see Vanves-Malakoff BI popping but when zooming in again it disappears and is replaced by Vanves-Malakoff 05. It appears again with another zoom level.

https://github.com/user-attachments/assets/815e85bf-0373-4b2c-a8be-42fbd2ab0e40

ui-speedspacechart/src/components/const.ts
flomonster: I took example from the space time chart. 

Kind of related. Do you have a suggestion for this issue: https://github.com/OpenRailAssociation/osrd-ui/pull/455/files/9fab2503afcb3d252edc2304a4f9a2754cfaf383#r1761472603 ?

flomonster:
> Is it a known issue ? In the example, when zooming in, we see Vanves-Malakoff BI popping but when zooming in again it disappears and is replaced by Vanves-Malakoff 05. It appears again with another zoom level.
> Enregistrement.de.l.ecran.2024-09-17.a.15.36.30.mov

I've noticed this issue. It's due to the naivety of the function that selects which stops overlap or not. 
This  behaviour is not introduced by this PR (since I reused the same strategy as previously). I suggest to fix this in another PR if prioritized.

ui-speedspacechart/src/components/layers/GridX.tsx
flomonster: I suggest: `StepsLayer`.

- It's closer to what this layer represents
- It contains the word `Layer`

It makes a lot of sense when we look at this piece of code:

```ts
      {store.layersDisplay.steps && (
        <StepsLayer width={adjustedWidthRightAxis} height={height} store={store} />
      )}
```

> [!NOTE]
> I'm not a big fan of the name "step" everywhere in osrd this object is named "operationnal point" (or "op"). We can rename this later.

Yohh: I named it steps because even if this is obvious for us, I'm no sure this is really relevant for anyone outside of the project how could use this module

Yohh:
> > Is it a known issue ? In the example, when zooming in, we see Vanves-Malakoff BI popping but when zooming in again it disappears and is replaced by Vanves-Malakoff 05. It appears again with another zoom level.
> > Enregistrement.de.l.ecran.2024-09-17.a.15.36.30.mov
> 
> I've noticed this issue. It's due to the naivety of the function that selects which stops overlap or not. This behaviour is not introduced by this PR (since I reused the same strategy as previously). I suggest to fix this in another PR if prioritized.

basically the steps are displayed if they have enough space to not overlap (16px), this is why it can change at each zoom level, but this is not just about the way the text is display while zooming, now the texts and the prs are displayed in their own way each, which seems to be a new unwanted behavior, but yes this should be fixed in another issue.

ui-speedspacechart/src/components/helpers/drawElements/gridX.ts
Yohh: we don't have so many levels of opacity, I think they can be set in their own color variable.

in the other hand we could have variables for colors and variables for opacity levels, then concatenate them if we need instead of having magic-numbers-like in strings, this will be easier if we keep hexa.

flomonster:
> but this is not just about the way the text is display while zooming, now the texts and the prs are displayed in their own way each, which seems to be a new unwanted behavior, but yes this should be fixed in another issue.

It's not an unwanted behavior at all. They should be displayed in their own way given the mock-up.
I agree that not correcting the zoom stability and adding this behavior. The two combined give a strange effect.

ui-speedspacechart/src/components/common/DetailsBox.tsx
jacomyal: OK for me. At least the dependence and the name are clear.

ui-speedspacechart/src/components/helpers/drawElements/gridX.ts
jacomyal: > Fixed. I didn't know how to handle different level of opacity for a same color properly.
> 
> I'm crafting colors using format string to add opacity value. I don't think it's very readable. Any advice?

Personally, I would use the "small-ish" JS library dedicated to colors management [chroma-js](https://www.npmjs.com/package/chroma-js) (already a dependency of the `spacetimechart` by the way). It would do something like `const transparentColor = chroma(color).alpha(0.5).rgba()` for instance.

jacomyal (approved):

Caracol3 (approved):
lgtm

ui-speedspacechart/src/components/helpers/drawElements/gridX.ts
flomonster: Awesome, exactly what I was looking for :pray:

Yohh:
just one question that doesn't have to block this PR, do we really need to install a whole lib (chroma-js) to use it at last only for 3 cases?

edit: To be more precise: we install an entire node module, which we will have to maintain, while 3 variables seem to be enough

flomonster:
> just one question that doesn't have to block this PR, do we really need to install a whole lib (chroma-js) to use it at last only for 3 cases?
> 
> edit: To be more precise: we install an entire node module, which we will have to maintain, while 3 variables seem to be enough

We should refacto completely how colors are used across osrd-ui. We should have a library of defined colors so each component doesn't have to redefined their own colors. Having a library here is okay for readability. Using hexa codes was really bad. 
I could declare all colors with their opacity level but the naming would have been unclear. How do you name the color `GREY_50` with an opacity level of `0.5`?

ui-speedspacechart/src/components/utils.ts
```diff
@@ -266,24 +267,133 @@ export const createSvgBlobUrl = (svgString: string): string => {
...
+export const filterStops = (
+  stops: LayerData<string>[],
+  ratioX: number,
+  width: number,
+  maxPosition: number,
+  minSpace = 8
+) => {
+  let lastDisplayedPosX: number | null = null;
+
+  const filteredStops = stops.filter(({ position }) => {
```
clarani, line 341: Here you suppose that the stops are sorted by position.start right ?

ui-speedspacechart/src/components/utils.ts
```diff
@@ -266,24 +267,133 @@ export const createSvgBlobUrl = (svgString: string): string => {
...
+      return false;
+    }
+    lastDisplayedPosX = posX;
+    return true;
+  });
+
+  return filteredStops;
+};
+
+/** Compute the cursor position on the x-axis given it's position on the canva */
```
clarani:
```suggestion
/** Compute the cursor position on the x-axis given its position on the canva */
```

ui-speedspacechart/src/components/utils.ts
```diff
@@ -266,24 +267,133 @@ export const createSvgBlobUrl = (svgString: string): string => {
...
+export const getCursorPosition = (cursorX: number, width: number, store: Store) => {
+  const { ratioX, leftOffset } = store;
+  const maxPosition = maxPositionValue(store);
+  return (
+    ((cursorX - leftOffset - MARGINS.CURVE_MARGIN_SIDES / 2) * maxPosition) /
+    ((width - MARGINS.MARGIN_LEFT - MARGINS.MARGIN_RIGHT - MARGINS.CURVE_MARGIN_SIDES) * ratioX)
+  );
```
clarani:
```suggestion
  const x = cursorX - leftOffset - MARGINS.CURVE_MARGIN_SIDES / 2;
  const maxX =
    (width - MARGINS.MARGIN_LEFT - MARGINS.MARGIN_RIGHT - MARGINS.CURVE_MARGIN_SIDES) * ratioX;
  return (x * maxPosition) / maxX;
```

ui-speedspacechart/src/__tests__/utils.spec.ts
```diff
@@ -351,8 +348,49 @@ describe('getStopPosition', () => {
     const width = 1000;
     const ratioX = 1;
     const maxPosition = 600;
-    const stopPosition = getStopPosition(position, width, ratioX, maxPosition);
-    expect(stopPosition).toBe(336);
+    const stopPosition = positionToPosX(position.start, maxPosition, width, ratioX);
```
clarani: You forgot to change the name of the tested function: getStopPosition => positionToPosX
```suggestion
    const x = positionToPosX(position.start, maxPosition, width, ratioX);
```

ui-speedspacechart/src/components/layers/StepsLayer.tsx
```diff
@@ -1,17 +1,17 @@
 import React from 'react';
 
 import type { Store } from '../../types/chartTypes';
-import { drawGridX } from '../helpers/drawElements/gridX';
+import { drawSteps } from '../helpers/drawElements/steps';
 import { useCanvas } from '../hooks';
 
-type AxisLayerXProps = {
+type GridXProps = {
```
clarani:
```suggestion
type StepsLayerProps = {
```

ui-speedspacechart/src/components/layers/StepsLayer.tsx
clarani, line 1: I think the id of the layer should be `"steps-layer"` and not `"axis-layer-x"`

ui-speedspacechart/src/components/helpers/drawElements/steps.ts
```diff
@@ -0,0 +1,104 @@
...
+  ctx.strokeStyle = LIGHT_BLUE.alpha(0.5).hex();
+  ctx.lineWidth = 1;
+  ctx.lineCap = 'round';
+  ctx.setLineDash([1, 6]);
+
+  // Vertical lines
+  ctx.beginPath();
+  filteredStops.forEach(({ position }) => {
+    const posX = positionToPosX(position.start, maxPosition, width, ratioX);
+    ctx.moveTo(posX, height - (height + MARGIN_TOP));
```
clarani:
```suggestion
    ctx.moveTo(posX, MARGIN_TOP);
```

ui-speedspacechart/src/components/helpers/drawElements/steps.ts
```diff
@@ -0,0 +1,104 @@
...
+  });
+  ctx.stroke();
+
+  // Step rectangles
+  ctx.lineWidth = 2;
+  ctx.fillStyle = GREY_50.hex();
+  ctx.strokeStyle = GREY_50.hex();
+  ctx.setLineDash([]);
+
+  const snappedStop = cursor.x ? getSnappedStop(cursor.x!, width, store) : null;
```
clarani:
```suggestion
  const snappedStop = cursor.x ? getSnappedStop(cursor.x, width, store) : null;
```

ui-speedspacechart/src/components/helpers/drawElements/reticle.ts
```diff
@@ -56,18 +58,10 @@ export const drawCursor = ({ ctx, width, height, store }: DrawFunctionParams) =>
...
-  let curveY = 0;
-  let x = {
-    a: 0,
-    b: 0,
-  };
-  let y = {
-    a: 0,
-    b: 0,
-  };
+  let reticleX = cursor.x! + MARGIN_LEFT;
```
clarani: Move the lines 76-79 above this block and remove the !
```suggestion
  let reticleX = cursor.x + MARGIN_LEFT;
```

ui-speedspacechart/src/components/helpers/drawElements/reticle.ts
```diff
@@ -79,225 +73,185 @@ export const drawCursor = ({ ctx, width, height, store }: DrawFunctionParams) =>
...
-    if (powerRestrictions) {
-      const currentPowerRestriction = powerRestrictions.find(
-        ({ position }) =>
-          cursor.x! >= xPositionReference(position.start) &&
-          cursor.x! <= xPositionReference(position.end!)
-      );
-      powerRestrictionText = currentPowerRestriction?.value.powerRestriction || '';
+  // Check if cursor is snapping to stop
+  let snapToStop = false;
+  if (store.layersDisplay.steps) {
```
clarani:
```suggestion
  if (layersDisplay.steps) {
```

ui-speedspacechart/src/components/utils.ts
flomonster: Yes

clarani (approved):
LGTM :white_check_mark:

PR #475, opened 2024-09-13
Dco ci fix for merge groups

EthanPERRUZZA:


ghost (approved):
We will need to test it live since we cannot reproduce it, so let's go

PR #476, opened 2024-09-13
ui-manchette: fix crash when OP list is empty

emersion:
Array.length always returns a value >= 0, thus the check was always true. We don't want to enter that if with a zero length array, because we assume Array.at(n) returns a non-undefined value in there.


woshilapin (approved):

Uriel-Sautron (approved):
LGTM !!

clarani (approved):
LGTM not tested :white_check_mark:

PR #496, opened 2024-09-17
ui-core: fix checkbox label

SharglutDev:
As label was non optional and has a hard-coded width, even by passing an empty string inside it, the label was taking space in the dom


Caracol3 (approved):
lgtm :-)

emersion (approved):

clarani (approved):
LGTM :white_check_mark:  (not tested)

Issue #320, opened 2024-08-07
Add time zone handling to DatePicker component


kmer2016:
The current `DatePicker` component does not handle time zones effectively. This enhancement is necessary to provide flexibility for users who need to manage dates either in UTC or in their local time zone.

### Acceptance Criteria:
- Dates should be correctly processed and displayed based on if the component is in UTC or local time zone.
- All new and existing functionalities should pass unit tests.

PR #502, opened 2024-09-18
fix the date picker to ensure correct behavior when both a value and selectable slot are provided.

### LINKED ISSUE #8877: STDCM date selector does not use the search environment variables
### What happened?

When selecting a departure (or arrival) date. The component proposes invalid dates.
However, selecting an invalid date does nothing.

image

In this example, only the 17th and 18th are valid dates.

Only valid dates should be selectable. To know what dates (and times) are legit, the frontend should use the configuration returned by the search environment endpoint.

1. Go to STDCM (new interface)
2. Try to select a departure date

Recette (SNCF)

Firefox

6ea7e5609cfa77d9854d34043a10707adc504fa5

kmer2016:
close https://github.com/OpenRailAssociation/osrd/issues/8877. 
the first commit fix the regression introduced [here](https://github.com/OpenRailAssociation/osrd-ui/commit/8a551bad073d18850a061fab4d9340f06c072739#diff-8a7e22fd2ef54f4d1029746b6c4475678d73b96cfef052a11f5f91ca0f897755)
The last commit adds a story to test that the regression has been corrected. It will be removed before the merge

### DIFF
--- a/ui-core/src/components/inputs/datePicker/Calendar.tsx
+++ b/ui-core/src/components/inputs/datePicker/Calendar.tsx
@@ -34,11 +34,9 @@ const Day = ({ date, isReferenceDate, isSelectable, dayWrapperClassName, onClick
   </div>
 );
 
-const Calendar = ({ displayedMonthStartDate, onDayClick }: CalendarProps) => {
-  const { days, isReferenceDate, isDateSelectable, buildDayWrapperClassName } = useCalendar({
-    displayedMonthStartDate,
-    onDayClick,
-  });
+const Calendar = (props: CalendarProps) => {
+  const { days, isReferenceDate, isDateSelectable, buildDayWrapperClassName } = useCalendar(props);
+  const { displayedMonthStartDate, onDayClick } = props;
   return (
     <div className="calendar-wrapper">
       <div className="calendar-anatomy">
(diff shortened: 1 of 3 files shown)


SarahBellaha (approved):
LGTM :)

SharglutDev (commented):
There is a strange behavior because of the UTC in storybook : 

image

I was wondering why I saw only the 24th and 25th in the calendar when you passed "23/8/2024 22H00" for start date but you actually didn't :)

ui-core/src/components/inputs/datePicker/Calendar.tsx
```diff
@@ -34,16 +34,13 @@ const Day = ({ date, isReferenceDate, isSelectable, dayWrapperClassName, onClick
   </div>
 );
 
-const Calendar = ({ displayedMonthStartDate, onDayClick }: CalendarProps) => {
-  const { days, isReferenceDate, isDateSelectable, buildDayWrapperClassName } = useCalendar({
-    displayedMonthStartDate,
-    onDayClick,
-  });
+const Calendar = (props: CalendarProps) => {
+  const { days, isReferenceDate, isDateSelectable, buildDayWrapperClassName } = useCalendar(props);
```
clarani, line 38: Can you add a line like this ?
```suggestion
  const { days, isReferenceDate, isDateSelectable, buildDayWrapperClassName } = useCalendar(props);
  const { displayedMonthStartDate, onDayClick } = props;
```

clarani (approved):
LGTM :white_check_mark:

kmer2016:
> There is a strange behavior because of the UTC in storybook :
> 
> image
> 
> I was wondering why I saw only the 24th and 25th in the calendar when you passed "23/8/2024 22H00" for start date but you actually didn't :)

You're right, the component currently doesn't handle UTC dates. However, this [issue](https://github.com/OpenRailAssociation/osrd-ui/issues/320) will address and fix that soon. ;)

PR #504, opened 2024-09-18
manchette with spacetimechart, hook integration

Math-R:


ui-manchette-with-spacetimechart/src/hooks/useIsOverFlow.ts
```diff
@@ -0,0 +1,35 @@
...
+
+type CallbackFunction = {
+  (hasOverflow: boolean): void;
+};
+
+export const useIsOverflow = (
+  ref: RefObject<HTMLElement>,
+  callback: CallbackFunction
+): boolean | undefined => {
+  const [isOverflow, setIsOverflow] = useState<boolean | undefined>(undefined);
```
jacomyal: This could be a boolean all the time, it would make the code slightly more readable I think

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
jacomyal, line 1: I really do like that story, I like how it feels at the same time easy to use and easy to customize :+1:

ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -0,0 +1,168 @@
...
+
+import { BASE_OP_HEIGHT, MAX_TIME_WINDOW, MAX_ZOOM_X, MIN_ZOOM_X } from './consts';
+import type {
+  OperationalPointType,
+  ProjectPathTrainResult,
+  StyledOperationalPointType,
+} from '../types';
+import { getHeightWithoutLastWaypoint, msToS } from './utils';
+
+export const calcOperationalPointsToDisplay = (
```
jacomyal, line 15: I know it's not specific to that PR, but this function probably deserves some unit tests.

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -0,0 +1,194 @@
...
+  const toggleMode = useCallback(() => {
+    setState((prev) => ({ ...prev, isProportional: !prev.isProportional }));
+  }, []);
+
+  const computedScales = useMemo(
+    () => getScales(operationalPointsWithPosition, height, isProportional, yZoom),
+    [operationalPointsWithPosition, height, isProportional, yZoom]
+  );
+
+  return {
```
jacomyal: A bit of memoization could be nice here, because the object for `manchetteProps` and `spaceTimeChartProps` are always both regenerated, even if only one of them has actual new values. It's possible that this does not occur too often, but still, I would memoize those two objects.

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -0,0 +1,194 @@
...
+  useIsOverflow(manchette, checkOverflow);
+
+  // Memoize timeWindow to avoid recalculation on each render
+  const timeWindow = useMemo(
+    () => computeTimeWindow(projectPathTrainResult),
+    [projectPathTrainResult]
+  );
+
+  // Memoize computedOperationalPoints to avoid recalculations unless the dependencies change
+  const computedOperationalPointsToDisplay = useMemo(
```
clarani: I think operationPointsToDisplay would be enough for the name

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -0,0 +1,194 @@
...
+    () => calcOperationalPointsToDisplay(operationalPoints, height, isProportional, yZoom),
+    [operationalPoints, height, isProportional, yZoom]
+  );
+  const computedoperationnalPointsHeight = useMemo(
+    () =>
+      operationalPointsHeight(computedOperationalPointsToDisplay, height, yZoom, isProportional),
```
clarani: Can you rename this function computeOperationalPointsHeight ? The current name is not an action

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -0,0 +1,194 @@
...
+    () => computeTimeWindow(projectPathTrainResult),
+    [projectPathTrainResult]
+  );
+
+  // Memoize computedOperationalPoints to avoid recalculations unless the dependencies change
+  const computedOperationalPointsToDisplay = useMemo(
+    () => calcOperationalPointsToDisplay(operationalPoints, height, isProportional, yZoom),
+    [operationalPoints, height, isProportional, yZoom]
+  );
+  const computedoperationnalPointsHeight = useMemo(
```
clarani: operationalPointsHeight would be enough for the name I think

ui-manchette-with-spacetimechart/src/hooks/usePaths.ts
```diff
@@ -0,0 +1,25 @@
+import { useMemo } from 'react';
+
+import { type ProjectPathTrainResult, type SpaceTimeCurves } from '../types';
```
Yohh:
```suggestion
import { type ProjectPathTrainResult, type SpaceTimeCurves } from '@osrd-project/ui-manchette/dist/types';
```

same for `helpers.ts`

ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -0,0 +1,168 @@
...
+    // We start with the first operational point:
+    const result: StyledOperationalPointType[] = [
+      { ...operationalPoints[0], display: true, styles: {} },
+    ];
+    const from = operationalPoints.at(0)!.position;
+    const to = operationalPoints.at(-1)!.position;
+    const totalDistance = to - from;
+    const heightWithoutFinalOp = getHeightWithoutLastWaypoint(height);
+    let lastDisplayedOP = result[0];
+
```
kmer2016: What do you think about initializing `lastVisibleItem` to keep track of the last visible operational point. This will help avoid the need for an additional `findLast` operation, improving the performance of the code.
```suggestion
let lastVisibleItem = result[0];
```

ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -0,0 +1,168 @@
...
+        lastItem.display = true;
+      }
+    }
+
+    return result;
+  }
+  // For non-proportional display, we always display all the operational points:
+  else {
+    return operationalPoints.map((op) => ({ ...op, styles: {}, display: true }));
+  }
```
Yohh:
```suggestion
    return operationalPoints.map((op) => ({ ...op, styles: {}, display: true }));
```

kmer2016: Consider using early return for the non-proportional display case or when there are no operational points. This simplifies the control flow and makes the code more readable.

ui-manchette-with-spacetimechart/src/helpers.ts
kmer2016: It would be very nice if all helpers could be unit tested.

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+    manchette,
+    selectedProjection
+  );
+
+  return (
+    <div className="manchette-space-time-chart-wrapper">
+      <div
+        className="header bg-ambientB-5 w-full border-b border-grey-30"
+        style={{ height: '40px' }}
+      ></div>
```
Yohh, line 44: is it supposed to be filled?

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
kmer2016:
```suggestion
  const computedOperationnalPointsHeight = useMemo(
```

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
Math-R: Done thanks !

ui-manchette-with-spacetimechart/src/hooks/usePaths.ts
Math-R: thanks

ui-manchette-with-spacetimechart/src/hooks/useIsOverFlow.ts
Math-R: Sure !

ui-manchette-with-spacetimechart/src/helpers.ts
Math-R: done

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
Math-R: Your comment is not clear to me, what do you mean ?

ui-manchette-with-spacetimechart/src/helpers.ts
Math-R: I don't get your comment either, if you look line 31 this is already set. The logic that follow is to ensure that the last op from a path is always displayed in case the penultimate had taken priority, during the "available height check".

Can you give me more explanation?

ui-manchette-with-spacetimechart/src/helpers.ts
Math-R: you're both right i added some :)

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
Yohh: does this div have to contain another element? or can it be auto-closed?

ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -0,0 +1,166 @@
...
+  StyledOperationalPointType,
+} from '@osrd-project/ui-manchette/dist/types';
+import type {
+  OperationalPoint,
+  SpaceTimeChartProps,
+} from '@osrd-project/ui-spacetimechart/dist/lib/types';
+
+import { BASE_OP_HEIGHT, MAX_TIME_WINDOW, MAX_ZOOM_X, MIN_ZOOM_X } from './consts';
+import { getHeightWithoutLastWaypoint, msToS } from './utils';
+
```
kmer2016, line 14: As seen in Front meeting, to make the code where we call `calcOperationalPointsToDisplay`, `getScales` and `operationalPointsHeight`, what do you think about create a type for options as in the following suggestion ? This new type group options like `height`, `isProportional` and `zoom`. 
```ts
// type.ts
export type OperationalPointsOptions = {
  isProportional: boolean
  height: number; // perhaps be renamed `containerHeight` to avoid confusion?  isProportional: boolean;
  zoom: number;
};
```
```ts
// helpers.ts
 export const calcOperationalPointsToDisplay = ( 
 operationalPoints: StyledOperationalPointType[],
 options:OperationalPointsOptions
) => {
...
}

export const operationalPointsHeight = (
  operationalPoints: StyledOperationalPointType[],
  options: OperationalPointsOptions
) => {
...
}

export const getScales = (
  ops: OperationalPoint[],
  options: OperationalPointOptions
) => {
...
}
```

ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -0,0 +1,166 @@
...
+    if (lastVisibleItem) {
+      lastVisibleItem.display = false;
+      lastItem.display = true;
+    }
+  }
+
+  return result;
+};
+
+export const operationalPointsHeight = (
```
kmer2016:
```suggestion
export const calcOperationalPointsHeight = (
```

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -0,0 +1,218 @@
...
+  });
+
+  const { xZoom, yZoom, xOffset, yOffset, panY, panning, scrollPosition, isProportional } = state;
+
+  const paths = usePaths(projectPathTrainResult, selectedProjection);
+
+  const checkOverflow = useCallback((isOverflowFromCallback: boolean) => {
+    setState((prev) => ({ ...prev, panY: isOverflowFromCallback }));
+  }, []);
+  useIsOverflow(manchette, checkOverflow);
```
clarani: I don't really get what this piece of code is used for. If I comment it, nothing changes. Do we need it ?

clarani (commented):
Thanks for the PR, good work !

I found some small bugs:
- :white_check_mark: ~when I zoom in km mode, the scale of the GET is not correct anymore~ My bad, I checked and it's the data which is not correct
image
- I can't zoom in the time anymore (maj+scroll)

ui-manchette-with-spacetimechart/src/__tests__/helpers.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+          },
+        ],
+        name: '',
+        id: 0,
+        signal_updates: [],
+        rolling_stock_length: 0,
+      },
+    ];
+    const timeWindow: number = computeTimeWindow(mockLongTrains);
+    expect(timeWindow).toBe(MAX_TIME_WINDOW);
```
kmer2016, line 171: What do you think about adding a verification for the correct `timeWindow` value in the test case ? This would ensure that the calculated `timeWindow` is accurate, not just less than or equal to `MAX_TIME_WINDOW`.

ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -0,0 +1,166 @@
...
+
+export const calcOperationalPointsToDisplay = (
+  operationalPoints: OperationalPointType[],
+  height: number,
+  isProportional: boolean,
+  zoom: number
+): StyledOperationalPointType[] => {
+  if (!isProportional || operationalPoints.length === 0) {
+    // For non-proportional display, we always display all the operational points:
+    return operationalPoints.map((op) => ({ ...op, styles: {}, display: true }));
```
clarani: styles can be undefined, it would be better to not defined it. Same for the other returns below
```suggestion
    return operationalPoints.map((op) => ({ ...op, display: true }));
```

ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -0,0 +1,166 @@
...
+    const op = operationalPoints[i];
+    const diff = op.position - lastDisplayedOP.position;
+    const display = (diff / totalDistance) * heightWithoutFinalOp * zoom >= BASE_OP_HEIGHT;
+    result.push({
+      ...op,
+      display,
+      styles: {},
+    });
+
+    if (display) lastDisplayedOP = result[i];
```
clarani: I think it would be better to add the op to result only if we should display it. Anyway, we don't use the non-displayed op in the code after this point ?
```suggestion
    if (display) {
      result.push({
        ...op,
        display,
      });
      lastDisplayedOP = result[i];
    }
```

The Manchette should display all the waypoints that it receives, so you can remove the display prop in StyledOperationalPointType (maybe rename it StyledWaypoint ?) and remove the early return `if(display) return null` in the component `OperationalPoint` (line 8)

ui-manchette-with-spacetimechart/src/__tests__/helpers.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+    const result: StyledOperationalPointType[] = operationalPointsHeight(
+      mockStyledOperationalPoints,
+      500,
+      1,
+      true
+    );
+    expect(result).toHaveLength(mockStyledOperationalPoints.length);
+    result.forEach((op) => {
+      expect(op.styles?.height).toBeDefined();
+    });
```
kmer2016: By checking the height values directly, we make sure the calculated heights are exactly what we expect. This makes the test stronger and more reliable than just checking if the values are defined. What do you think?

ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -0,0 +1,166 @@
...
+} from '@osrd-project/ui-manchette/dist/types';
+import type {
+  OperationalPoint,
+  SpaceTimeChartProps,
+} from '@osrd-project/ui-spacetimechart/dist/lib/types';
+
+import { BASE_OP_HEIGHT, MAX_TIME_WINDOW, MAX_ZOOM_X, MIN_ZOOM_X } from './consts';
+import { getHeightWithoutLastWaypoint, msToS } from './utils';
+
+export const calcOperationalPointsToDisplay = (
```
clarani, line 15: Maybe `filterWaypointsToDisplay` would be a better name
```suggestion
export const filterWaypointsToDisplay = (
```

ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -0,0 +1,166 @@
...
+export const operationalPointsHeight = (
+  operationalPoints: StyledOperationalPointType[],
+  height: number,
+  zoom: number,
+  isProportional: boolean
+) => {
+  if (operationalPoints.length < 2) return [];
+  const from = operationalPoints.at(0)!.position;
+  const to = operationalPoints.at(-1)!.position;
+  const totalDistance = to - from;
```
kmer2016: I noticed this pattern in at least 3 functions (`getScales`, `operationalPointsHeight`, and `calcOperationalPointsToDisplay`). Shouldn't we move it to the utils?

ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -0,0 +1,166 @@
...
+  // In the end, to make sure the last point is visible, if it's not, we swap
+  // it with the last visible item:
+  const lastItem = result[result.length - 1];
+  if (!lastItem.display) {
+    const lastVisibleItem = result.findLast((op) => op.display);
+    if (lastVisibleItem) {
+      lastVisibleItem.display = false;
+      lastItem.display = true;
+    }
+  }
```
clarani: If you have filter the non visible points, then you can simplify your code like this:
```suggestion
  if (lastItem.id !== operationalPoints.at(-1)!.id) {
    result.splice(-1, 1, operationalPoints.at(-1)!);
  }
```

ui-manchette-with-spacetimechart/src/__tests__/helpers.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+    const result: OperationalPoint[] = getOperationalPointsWithPosition(mockOperationalPoints);
+    expect(result).toHaveLength(mockOperationalPoints.length);
+    result.forEach((op, index) => {
+      expect(op.id).toBe(mockOperationalPoints[index].id);
+      expect(op.position).toBe(mockOperationalPoints[index].position);
+    });
+  });
+});
+
+// Tests for getScales
```
kmer2016, line 187: I noticed that we are not testing the case when the `ops.length` is less than 2. It might be helpful to add a test for that scenario to ensure the function returns an empty array as expected

ui-manchette-with-spacetimechart/src/__tests__/helpers.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+describe('getScales', () => {
+  const mockOpsWithPosition: OperationalPoint[] = mockOperationalPoints.map((op) => ({
+    id: op.id,
+    label: op.id,
+    position: op.position,
+    importanceLevel: 1,
+  }));
+
+  it('should return correct scale coefficients for proportional display', () => {
```
kmer2016, line 206: The current tests check if coefficient and size are defined, which is great. However, we could strengthen the tests by also verifying the actual calculated values to ensure they are correct.

ui-manchette-with-spacetimechart/src/hooks/usePaths.ts
```diff
@@ -0,0 +1,28 @@
...
+  type SpaceTimeCurves,
+} from '@osrd-project/ui-manchette/dist/types'; 
+
+const transformCurve = (curve: SpaceTimeCurves, departureTime: string) =>
+  curve.positions.map((position, i) => ({
+    time: curve.times[i] + new Date(departureTime).getTime(),
+    position,
+  }));
+
+const usePaths = (projectPathTrainResult: ProjectPathTrainResult[], selectedProjection?: number) =>
```
clarani: It's tech debt, but can you rename selectedProjection by selectedTrain ? We don't want to highlight the train used for the projection, but the one selected
```suggestion
const usePaths = (projectPathTrainResult: ProjectPathTrainResult[], selectedTrain?: number) =>
```

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { SAMPLE_PATH_PROPERTIES_DATA, SAMPLE_PATHS_DATA } from '../assets/sampleData';
+import useManchettesWithSpaceTimeChart from '../hooks/useManchetteWithSpaceTimeChart';
+
+const OperationalPointListData = SAMPLE_PATH_PROPERTIES_DATA.operational_points ?? [];
+const PathData = SAMPLE_PATHS_DATA ?? [];
+
+type ManchetteWithSpaceTimeWrapperProps = {
+  operationalPoints: OperationalPointType[];
+  projectPathTrainResult: ProjectPathTrainResult[];
+  selectedProjection: number;
```
clarani: Rename the props selectedProjection to selectedTrain (see my previous comment, it's tech debt)

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+    <div className="manchette-space-time-chart-wrapper">
+      <div
+        className="header bg-ambientB-5 w-full border-b border-grey-30"
+        style={{ height: '40px' }}
+      ></div>
+      <div
+        ref={manchette}
+        className="manchette flex"
+        style={{ height: `${DEFAULT_HEIGHT}px` }}
+        onScroll={manchetteProps.handleScroll}
```
clarani: The props `handleScroll` should not be in the manchetteProps
```suggestion
        onScroll={handleScroll}
```

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+};
+const DEFAULT_HEIGHT = 561;
+
+const ManchetteWithSpaceTimeWrapper = ({
+  operationalPoints,
+  projectPathTrainResult,
+  selectedProjection,
+}: ManchetteWithSpaceTimeWrapperProps) => {
+  const manchette = useRef<HTMLDivElement>(null);
+  const { manchetteProps, spaceTimeChartProps } = useManchettesWithSpaceTimeChart(
```
clarani:
```suggestion
  const { manchetteProps, spaceTimeChartProps, handleScroll } = useManchettesWithSpaceTimeChart(
```

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+  );
+
+  return (
+    <div className="manchette-space-time-chart-wrapper">
+      <div
+        className="header bg-ambientB-5 w-full border-b border-grey-30"
+        style={{ height: '40px' }}
+      ></div>
+      <div
+        ref={manchette}
```
clarani: This is not the manchette ref, but the ref of the container of the spaceTimeChart and the manchette. Another possible name would be containerRef ?
```suggestion
        ref={spaceTimeChartWithManchetteRef}
```

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+        className="manchette flex"
+        style={{ height: `${DEFAULT_HEIGHT}px` }}
+        onScroll={manchetteProps.handleScroll}
+      >
+        <Manchette {...manchetteProps} />
+        <div
+          className="space-time-chart-container w-full sticky"
+          style={{ bottom: 0, left: 0, top: 0, height: `${DEFAULT_HEIGHT - 6}px` }}
```
clarani, line 54: Why 6 px here ?

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+        >
+          <SpaceTimeChart
+            className="inset-0 absolute h-full"
+            spaceOrigin={0}
+            timeOrigin={Math.min(...projectPathTrainResult.map((p) => +new Date(p.departure_time)))}
+            timeScale={spaceTimeChartProps.timeWindow / spaceTimeChartProps.xZoom}
```
clarani: Those 2 props can be computed in the spaceTimeChartProps directly

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+        </div>
+      </div>
+    </div>
+  );
+};
+
+export default {
+  title: 'Manchette/WithSpaceTimeChart',
+  component: ManchetteWithSpaceTimeWrapper,
+} as Meta<typeof ManchetteWithSpaceTimeWrapper>;
```
clarani: I have a ts error:
```suggestion
const ManchetteWithSpaceTimeStory: Meta<typeof ManchetteWithSpaceTimeWrapper> = {
  title: 'Manchette/WithSpaceTimeChart',
  component: ManchetteWithSpaceTimeWrapper,
};

export default ManchetteWithSpaceTimeStory;
```

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+export const Default = {
+  args: {
+    operationalPoints: OperationalPointListData,
+    projectPathTrainResult: PathData,
```
clarani:
```suggestion
    operationalPoints: SAMPLE_PATH_PROPERTIES_DATA.operational_points,
    projectPathTrainResult: SAMPLE_PATHS_DATA,
```

ui-manchette-with-spacetimechart/src/__tests__/helpers.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+      500,
+      true,
+      1
+    );
+    expect(result[result.length - 1].display).toBe(true);
+  });
+});
+
+// Tests for operationalPointsHeight
+describe('operationalPointsHeight', () => {
```
kmer2016: The function handles cases with fewer than 2 operational points. We could add a unit test for this scenario to ensure it's fully covered

ui-manchette-with-spacetimechart/src/index.ts
```diff
@@ -0,0 +1,4 @@
+import '@osrd-project/ui-core/dist/theme.css';
+import './styles/main.css';
```
clarani: Since we export only a hook, should we import the style here ?

ui-manchette-with-spacetimechart/src/assets/sampleData.ts
```diff
@@ -0,0 +1,1505 @@
...
+      [-0.11892761130523458, 49.50049510188119],
+      [-0.1149, 49.50296],
+      [-0.1149, 49.50997],
+      [-0.1099, 49.512899999999995],
+      [-0.10924003772863546, 49.512899999999995],
+      [-0.108, 49.512899999999995],
+      [-0.10638, 49.512899999999995],
+      [-0.0945, 49.512899999999995],
+    ],
+  },
```
clarani: You only need the operational_points
```suggestion
```

ui-manchette-with-spacetimechart/src/assets/sampleData.ts
```diff
@@ -0,0 +1,1505 @@
...
+          0, 42000, 44000, 46000, 48000, 54000, 60000, 68000, 86000, 108000, 134000, 164000, 204000,
+          286000, 406000, 442000, 456000, 558000, 594000, 610000, 724159, 733630, 737630, 743630,
+          747630, 751630, 755630, 757630, 759630, 761630, 821630, 823630, 825630, 827630, 831630,
+          837630, 843630, 857630, 875630, 897630, 923630, 953630, 999630, 1053630, 1215630, 1269630,
+          1327630, 1389630, 1437630, 1493630, 1661630, 1713630, 1761630, 1807630, 1903630, 2031630,
+          2561539, 2569617, 2573617, 2579617, 2583617, 2589617, 2593617, 2595617, 2597617, 2599617,
+        ],
+      },
+    ],
+    signal_updates: [
```
clarani: You don't need the signal_updates, so you can remove it too for each train

ui-manchette-with-spacetimechart/src/assets/sampleData.ts
```diff
@@ -0,0 +1,1505 @@
+import { type PathProperties, type ProjectPathTrainResult } from '../../types';
```
clarani: The import here is not correct

clarani (commented):
I added some comments.
- You have some warnings in the console
- Don't forget to clean the ui-manchette folder

kmer2016 (commented):
Thank you for this PR! I have a few suggestions, but the component works well, and I didn't find any bugs in the new component. However, the stories for `OperationalPoint`, `OperationalPointList`, and `Manchette` are empty. Should we consider deleting them?

Also, there are some warnings from the linter.

ui-manchette-with-spacetimechart/src/utils/vector.ts
```diff
@@ -0,0 +1,11 @@
+export type Point = {
```
anisometropie: use the same name for the file and the type.

anisometropie (changes requested):

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+            ))}
+          </SpaceTimeChart>
+        </div>
+      </div>
+    </div>
+  );
+};
+
+export default {
+  title: 'Manchette/WithSpaceTimeChart',
```
Yohh: we could display it in her own storybook folder:
```suggestion
  title: 'Manchette with SpaceTimeChart/rendering',
```

then remove `OperationalPoint.stories.ts` and `OperationalPointList.stories.ts`

finally, set data in `Manchette.stories.ts` and change its storybook title with `title: 'Manchette/rendering'`

edit: is `Manchette.stories.ts` really usefull? (maybe a larger refacto in another PR)

ui-manchette-with-spacetimechart/src/helpers.ts
clarani: We can delay the resolution of this comment

ui-manchette-with-spacetimechart/src/helpers.ts
clarani: We can delay the resolution of this comment

ui-manchette-with-spacetimechart/src/utils/vector.ts
clarani: I think that the type is really used to describe a point and not a Vector, so I would keep this name for the type.
This type and file names do not bother me particularly

ui-manchette-with-spacetimechart/src/helpers.ts
Math-R: yes done it

ui-manchette-with-spacetimechart/src/helpers.ts
Math-R: Since the refacto with solely the point display push in the result, i think it would be better to do it at this time, because for now they are all returned

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
Math-R: It checked if the manchette got overflow, and if not it prevents the spaceTimeChart to being able to pan vertically

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
Math-R: @kmer2016 made a commentary for the naming of this function, and i choose his one

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
Math-R: yes also did the same for computedOperationalPointsHeight => operationalPointsWithHeight

ui-manchette-with-spacetimechart/src/assets/sampleData.ts
Math-R: done

ui-manchette-with-spacetimechart/src/assets/sampleData.ts
Math-R: done

ui-manchette-with-spacetimechart/src/assets/sampleData.ts
Math-R: done

ui-manchette-with-spacetimechart/src/helpers.ts
Math-R: After a discusiion with @clarani, for the rest of the comment we decided that this was a bit overenginnered but the boolean will be clearer passing a props options

Math-R: done

ui-manchette-with-spacetimechart/src/helpers.ts
Math-R: thanks

ui-manchette-with-spacetimechart/src/helpers.ts
Math-R: Nice thanks !

ui-manchette-with-spacetimechart/src/helpers.ts
Math-R: again thanks !

ui-manchette-with-spacetimechart/src/helpers.ts
Math-R: done

ui-manchette-with-spacetimechart/src/__tests__/helpers.spec.ts
Math-R: done

ui-manchette-with-spacetimechart/src/__tests__/helpers.spec.ts
Math-R: done

ui-manchette-with-spacetimechart/src/__tests__/helpers.spec.ts
Math-R: done

ui-manchette-with-spacetimechart/src/__tests__/helpers.spec.ts
Math-R: done

ui-manchette-with-spacetimechart/src/__tests__/helpers.spec.ts
Math-R: it will be done by @clarani in a futur PR

ui-manchette-with-spacetimechart/src/hooks/usePaths.ts
Math-R: done

ui-manchette-with-spacetimechart/src/index.ts
Math-R: it's used by the story, if it not expotr it can be compile to the dist and the style in the story fail

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
Math-R: done

ui-manchette-with-spacetimechart/src/utils/vector.ts
Math-R: yes it is !

ui-manchette-with-spacetimechart/src/helpers.ts
Math-R: i don't see the difference but maybe it's a doublon from the @kmer2016 [commentary](https://github.com/OpenRailAssociation/osrd-ui/pull/504#discussion_r1771464943)

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
Math-R: it can be autoclosed in the story but may contain some stuff if reproduced in a project integration

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
Math-R: done

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
Math-R: done but with the name of the project

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
Math-R: done

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
Math-R: done

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
Math-R: i have not and @Yohh neither can you show the error please

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
Math-R: Because without this is not aligned in the integration

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
Math-R: Good point !

Math-R:
> Thanks for the PR, good work !
> 
> I found some small bugs:
> 
> * ✅ ~when I zoom in km mode, the scale of the GET is not correct anymore~ My bad, I checked and it's the data which is not correct
>   image
> * I can't zoom in the time anymore (maj+scroll)

The zoom x is now fixed

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
clarani: I think you forgot this one

ui-manchette-with-spacetimechart/src/index.ts
clarani: Since it is used only in the story, can we import it in the story ?
The index file is what is exposed outside the library, I think it's not a good practice to expose things that the users won't use

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
clarani: image

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -134,32 +136,41 @@ const useManchettesWithSpaceTimeChart = (
...
-        handleScroll,
         manchetteRef: manchette,
         yZoom,
         isProportional,
       },
       spaceTimeChartProps: {
         operationalPoints: operationalPointsWithPosition,
         spaceScales: computedScales,
+        timeScale: timeWindow / xZoom,
         timeWindow,
```
clarani: You don't need to return the timeWindow anymore
```suggestion
```

ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -62,32 +56,28 @@ export const calcOperationalPointsToDisplay = (
   return result;
 };
 
-export const operationalPointsHeight = (
+export const calcOperationalPointsHeight = (
   operationalPoints: StyledOperationalPointType[],
-  height: number,
-  zoom: number,
-  isProportional: boolean
+  options: OperationalPointsOptions
```
clarani: It's a suggestion to make the code more readable
```suggestion
  { height, isProportional, yZoom }: OperationalPointsOptions
```

ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -124,22 +114,17 @@ export const getOperationalPointsWithPosition = (
     importanceLevel: 1,
   }));
 
-export const getScales = (
-  ops: OperationalPoint[],
-  height: number,
-  isProportional: boolean,
-  zoomY: number
-) => {
+export const getScales = (ops: OperationalPoint[], options: OperationalPointsOptions) => {
```
clarani: Same here
```suggestion
export const getScales = (ops: OperationalPoint[], { height, isProportional, yZoom }: OperationalPointsOptions) => {
```

ui-manchette/src/components/helpers.ts
clarani, line 1: I think you can remove this file

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -0,0 +1,231 @@
...
+    setState((prev) => ({ ...prev, isProportional: !prev.isProportional }));
+  }, []);
+
+  const computedScales = useMemo(
+    () => getScales(operationalPointsWithPosition, { height, isProportional, yZoom }),
+    [operationalPointsWithPosition, height, isProportional, yZoom]
+  );
+
+  // Memoize the returned object to avoid unnecessary recalculations
+  return useMemo(
```
jacomyal: My concern remains here: Since it's all memoized together, a component depending on `manchetteProps` (typically, the manchette) will be rerendered when `spaceTimeChartProps` is updated, and vice versa.

The two big branches of what you return from that function (`manchetteProps` and `spaceTimeChartProps`) should be memoized separately.

Something like:

```typescript
const manchetteProps = useMemo(/* ... */);
const spaceTimeChartProps = useMemo(/* ... */);

return {
  manchetteProps,
  spaceTimeChartProps,
  /* ... */
};
```

clarani (commented):
In ui-manchette: 
- you can also remove the files ./consts.ts
- you need to update the sampleData too (clean `SAMPLE_PATH_PROPERTIES_DATA` and remove `SAMPLE_PATHS_DATA`)
- in the story, remove the const `PathData` and directly give the `SAMPLE_PATH_PROPERTIES_DATA.operational_points` line 34 in the args
- maybe remove the css of the manchette
- can you check that the utils are still necessary ?

jacomyal (changes requested):

ui-manchette-with-spacetimechart/src/utils/vector.ts
anisometropie: use **point** for both then, if vector doesn’t exist

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
Math-R: done

ui-manchette-with-spacetimechart/src/index.ts
Math-R: Edit: it works =)

ui-manchette-with-spacetimechart/src/utils/vector.ts
Math-R: done

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
Math-R: thanks =)

Math-R:
$

> In ui-manchette:
> 
> * you can also remove the files ./consts.ts

not entirely

-------------------------------------------------------------------------------------
> * you need to update the sampleData too (clean `SAMPLE_PATH_PROPERTIES_DATA` and remove `SAMPLE_PATHS_DATA`)
> * in the story, remove the const `PathData` and directly give the `SAMPLE_PATH_PROPERTIES_DATA.operational_points` line 34 in the args

I think the stories need a rework if we would like to keep it

-------------------------------------------------------------------------------------
> * maybe remove the css of the manchette

Can't remove the css because it is useful here, manchette-with-spacetimechart just provide a hook and an example of integration but the Manchette the OperationalPointList and operationalPoint still need their css

-------------------------------------------------------------------------------------
> * can you check that the utils are still necessary ?

I clean the file but there is still one util used

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
Math-R: done

ui-manchette-with-spacetimechart/src/helpers.ts
Math-R: done

ui-manchette-with-spacetimechart/src/helpers.ts
Math-R: done

ui-manchette/src/components/helpers.ts
Math-R: done

ui-manchette-with-spacetimechart/src/index.ts
clarani: Nice !

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -0,0 +1,235 @@
...
+  operationalPointsChart: OperationalPoint[];
+  panY: boolean;
+  scales: SpaceScale[];
+};
+
+const useManchettesWithSpaceTimeChart = (
+  operationalPoints: OperationalPointType[],
+  projectPathTrainResult: ProjectPathTrainResult[],
+  manchette: React.RefObject<HTMLDivElement>,
+  selectedProjection?: number,
```
clarani: Can you rename selectedProjection in selectedTrain here too ?

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -0,0 +1,235 @@
...
+  );
+
+  // Memoize manchetteProps separately
+  const manchetteProps = useMemo(
+    () => ({
+      operationalPoints: operationalPointsWithHeight,
+      zoomYIn,
+      zoomYOut,
+      toggleMode,
+      manchetteRef: manchette,
```
clarani: The Manchette does not need the manchetteRef anymore
```suggestion
```

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -0,0 +1,235 @@
...
+  isProportional: boolean;
+  operationalPointsChart: OperationalPoint[];
+  panY: boolean;
+  scales: SpaceScale[];
+};
+
+const useManchettesWithSpaceTimeChart = (
+  operationalPoints: OperationalPointType[],
+  projectPathTrainResult: ProjectPathTrainResult[],
+  manchette: React.RefObject<HTMLDivElement>,
```
clarani: Can you also rename this props ?
```suggestion
  manchetteWithSpaceTimeChartContainer: React.RefObject<HTMLDivElement>,
```

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+      <div
+        ref={manchetteWithSpaceTimeChartRef}
+        className="manchette flex"
+        style={{ height: `${DEFAULT_HEIGHT}px` }}
+        onScroll={handleScroll}
+      >
+        <Manchette {...manchetteProps} />
+        <div
+          className="space-time-chart-container w-full sticky"
+          style={{ bottom: 0, left: 0, top: 0, height: `${DEFAULT_HEIGHT - 6}px` }}
```
clarani, line 54: I think you can remove this line
```suggestion
```

kmer2016 (approved):
Good job ! LGTM

anisometropie (approved):

Yohh (approved):

jacomyal (approved):

jacomyal:
LGTM :+1:

ui-manchette-with-spacetimechart/src/index.ts
Math-R: finally, it doesn't, cause certain tailwind class are not built without this export

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
clarani: Ok seen, we need it.

clarani (commented):
Almost good, there is still some warnings in the console:
image

Math-R:
> Almost good, there is still some warnings in the console: image

it's fixed !

PR #508, opened 2024-09-19
Change value type to string in ComboBox

achrafmohye:


kmer2016 (approved):
LGTM !

kmer2016 (commented):
ComboBox must be display options list only if the user focus on it.

achrafmohye:
Like discussed with @axrolld added the functionality: when you return to a field with valid input, make modifications (or not), and then click outside to cancel your changes, the previous selection will remain intact instead of clearing everything.

kmer2016 (approved):

PR #509, opened 2024-09-19
Speedspacechart fix curves

### LINKED ISSUE #8333: Front: Speed space chart is broken when there is no stop at the end
### What happened?

When we put no stop at the end, the GEV is broken : 

Image

We should have a correct GEV

1. Create a TrainSchedule
2. Remove the stop at the end

Recette (SNCF)

Google Chrome

staging 0a34296

flomonster:
Curve drawing doesn't work when the train starts at non-zero speed/ends at non-zero speed.

fixes https://github.com/OpenRailAssociation/osrd/issues/8333

## Before

image

## After

image

### DIFF
--- a/ui-speedspacechart/src/components/helpers/drawElements/curve.ts
+++ b/ui-speedspacechart/src/components/helpers/drawElements/curve.ts
@@ -1,34 +1,35 @@
-import type {
-  CanvasConfig,
-  CurveConfig,
-  DrawFunctionParams,
-  LayerData,
-} from '../../../types/chartTypes';
+import type { DrawFunctionParams, LayerData } from '../../../types/chartTypes';
 import { MARGINS } from '../../const';
 import { clearCanvas, maxPositionValue, maxSpeedValue } from '../../utils';
 
 const { CURVE_MARGIN_TOP, CURVE_MARGIN_SIDES } = MARGINS;
 
-const drawSpecificCurve = (
-  curveConfig: CurveConfig,
-  canvasConfig: CanvasConfig,
+const computeCurvePoints = (
+  canvasConfig: { width: number; height: number },
+  curveConfig: { maxSpeed: number; maxPosition: number; ratioX: number },
   specificSpeeds: LayerData<number>[]
 ) => {
   const { maxSpeed, maxPosition, ratioX } = curveConfig;
-  const { width, height, ctx } = canvasConfig;
+  const { width, height } = canvasConfig;
 
   const adjustedWidth = width - CURVE_MARGIN_SIDES;
   const halfCurveMarginSides = CURVE_MARGIN_SIDES / 2;
   const adjustedHeight = height - CURVE_MARGIN_TOP;
   const xcoef = (adjustedWidth / maxPosition) * ratioX;
+  const points: { x: number; y: number }[] = [];
 
-  return specificSpeeds.forEach(({ position, value }) => {
+  specificSpeeds.forEach(({ position, value }) => {
     // normalize speed based on range of values
     const normalizedSpeed = value / maxSpeed;
     const x = position.start * xcoef + halfCurveMarginSides;
     const y = height - normalizedSpeed * adjustedHeight;
-    ctx.lineTo(x, y);
+    points.push({ x, y });
   });
+
+  // Close the path
+  points.push({ x: maxPosition * xcoef + halfCurveMarginSides, y: height });
+  points.push({ x: halfCurveMarginSides, y: height });
+  return points;
 };
 
 export const drawCurve = ({ ctx, width, height, store }: DrawFunctionParams) => {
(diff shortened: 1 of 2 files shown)


Yohh:
be careful, `ctx.globalCompositeOperation` purpose was to be able to display the declivities under the curves, removing it totally hides them:
[globalCompositeOperation](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation)


image

ui-speedspacechart/src/components/helpers/drawElements/curve.ts
```diff
@@ -42,27 +47,29 @@ export const drawCurve = ({ ctx, width, height, store }: DrawFunctionParams) =>
...
   const maxPosition = maxPositionValue(store);
 
-  ctx.lineWidth = 0.5;
-  ctx.fillStyle = 'rgba(17, 101, 180, 0.02)';
-  ctx.strokeStyle = 'rgb(17, 101, 180, 0.5)';
-
   ctx.beginPath();
-  drawSpecificCurve({ maxSpeed, maxPosition, ratioX }, { width, height, ctx }, speeds);
-  ctx.closePath();
+  ctx.fillStyle = 'rgba(17, 101, 180, 0.02)';
```
SharglutDev, line 63: No chroma for the colors in this file ?

ui-speedspacechart/src/components/helpers/drawElements/curve.ts
```diff
@@ -9,10 +9,11 @@ import { clearCanvas, maxPositionValue, maxSpeedValue } from '../../utils';
 
 const { CURVE_MARGIN_TOP, CURVE_MARGIN_SIDES } = MARGINS;
 
-const drawSpecificCurve = (
+const traceCurve = (
   curveConfig: CurveConfig,
   canvasConfig: CanvasConfig,
-  specificSpeeds: LayerData<number>[]
+  specificSpeeds: LayerData<number>[],
+  closePath: boolean
```
SharglutDev: Even though it's in the same file, we recently decided to use an object when passing a boolean in a function param so it can be more explicit why we call i with `true` or `false`

```suggestion
  { closePath: boolean }
```

ui-speedspacechart/src/components/helpers/drawElements/curve.ts
SharglutDev, line 1: Not sure to understand why do we pass from 2 traceCurve to 4 ?

flomonster: Ideally, I would call the function twice. The issue is to draw a shape with borders using different styles. Please let me know if you know how to do so.

ui-speedspacechart/src/components/helpers/drawElements/curve.ts
flomonster: I did not change anything yet. I will adapt curve colors in a future PR.

ui-speedspacechart/src/components/helpers/drawElements/curve.ts
flomonster: Okay... Is it optimized during the JS transformation?
It's something that the code editor should handle.

flomonster: Fixed here: https://github.com/OpenRailAssociation/osrd-ui/pull/509/commits/0e65f5a6a56c0fa79afce0a8271343676bbb2945.

I'm really not a big fan of this convention. What about putting the boolean into a viriable instead of buildign a record?

SharglutDev: Ah sorry for the confusion, I wasn't very clear maybe. The goal was to use the function like the following in osrd for example : 
`export function secToHoursString(sec: number | null, { withSeconds = false } = {}): TimeString {...}`

and then use it like that when calling the function : 
`secToHoursString(durationInSecond, { withSeconds: true })`

Is it ok for you ?

ui-speedspacechart/src/components/helpers/drawElements/curve.ts
SharglutDev: Don't think of something different but maybe we can add a comment for each step or to explain the purpose of using them twice each ?

ui-speedspacechart/src/components/helpers/drawElements/curve.ts
flomonster: In my case I have one variable and I don't want a default value to it. I don't see how do you want me handle this.

> [!NOTE]
> Making the function arguments more explicit seems like a good idea, but it makes the function declaration less readable.

ui-speedspacechart/src/components/helpers/drawElements/curve.ts
```diff
@@ -42,27 +41,39 @@ export const drawCurve = ({ ctx, width, height, store }: DrawFunctionParams) =>
...
+  ctx.beginPath();
+  ctx.lineWidth = 0.5;
+  ctx.strokeStyle = 'rgb(17, 101, 180, 0.5)';
+  traceCurve({ width, height, ctx }, { maxSpeed, maxPosition, ratioX, isClosePath: false }, speeds);
   ctx.stroke();
 
+  ctx.beginPath();
   ctx.fillStyle = 'rgba(255, 255, 255)';
-  ctx.strokeStyle = 'rgb(17, 101, 180)';
+  ctx.strokeStyle = 'rgba(0, 0, 0, 0)';
```
jacomyal: This line should be unnecessary I think, since in this part you only call `ctx.fill`, and `strokeStyle` will be reset later, before calling `ctx.stroke.`

ui-speedspacechart/src/components/helpers/drawElements/curve.ts
```diff
@@ -1,34 +1,33 @@
...
 import { MARGINS } from '../../const';
 import { clearCanvas, maxPositionValue, maxSpeedValue } from '../../utils';
 
 const { CURVE_MARGIN_TOP, CURVE_MARGIN_SIDES } = MARGINS;
 
-const drawSpecificCurve = (
-  curveConfig: CurveConfig,
+const traceCurve = (
   canvasConfig: CanvasConfig,
+  curveConfig: { maxSpeed: number; maxPosition: number; ratioX: number; isClosePath: boolean },
```
jacomyal: `isClosePath` sounds weird to me. I'd rather name that `closePath`, `doClosePath` or `isPathClosed`.

ui-speedspacechart/src/components/helpers/drawElements/curve.ts
```diff
@@ -42,27 +41,39 @@ export const drawCurve = ({ ctx, width, height, store }: DrawFunctionParams) =>
   const maxSpeed = maxSpeedValue(store);
   const maxPosition = maxPositionValue(store);
 
-  ctx.lineWidth = 0.5;
-  ctx.fillStyle = 'rgba(17, 101, 180, 0.02)';
-  ctx.strokeStyle = 'rgb(17, 101, 180, 0.5)';
-
+  // Curves must be drawn twice one for the fill and one for the stroke
```
jacomyal, line 57: You can draw the curve only once, if you do the things in this order:
- Describe curve path (with `moveTo` and `lineTo`)
- Draw the curve (with `stroke`)
- Close the path (with the last `lineTo` calls and `closePath`)
- Fill the curve (with `fill`)

You can see an example here:
https://codesandbox.io/p/sandbox/d367fh?file=%2Fsrc%2Findex.mjs

Yohh: if you fill after stroke, the area filled will partially overlap the stroke, right?

ui-speedspacechart/src/components/helpers/drawElements/curve.ts
flomonster: Done

ui-speedspacechart/src/components/helpers/drawElements/curve.ts
flomonster: I removed this parameter.

ui-speedspacechart/src/components/helpers/drawElements/curve.ts
flomonster: Indeed. It doesn't seem possible to do this in one pass.

> [!NOTE]
> I did a little optimization. Now, x/y points positions are computed only once and reuse for the stroke.

ui-speedspacechart/src/components/helpers/drawElements/curve.ts
```diff
@@ -1,34 +1,35 @@
...
+import type { DrawFunctionParams, LayerData } from '../../../types/chartTypes';
 import { MARGINS } from '../../const';
 import { clearCanvas, maxPositionValue, maxSpeedValue } from '../../utils';
 
 const { CURVE_MARGIN_TOP, CURVE_MARGIN_SIDES } = MARGINS;
 
-const drawSpecificCurve = (
-  curveConfig: CurveConfig,
-  canvasConfig: CanvasConfig,
+const curvePoints = (
```
SharglutDev: Feels weird to have a function named like a const. It could be `computeCurvePoints` or something like that ?

SharglutDev (approved):
Lgtm

PR #523, opened 2024-09-20
ui-spacetimechart: work schedules on the space time chart

### LINKED ISSUE #401: osrd-ui: display work schedule blocks in GET
https://github.com/osrd-project/osrd-confidential/issues/531

reuse design proposed in `ui-spacetimechart/src/stories/additional-data.stories.tsx` to display dashed regions on the Space time

anisometropie:
closes https://github.com/OpenRailAssociation/osrd-ui/issues/401

### DIFF
--- a/ui-spacetimechart/src/index.ts
+++ b/ui-spacetimechart/src/index.ts
@@ -4,3 +4,5 @@ export * from './components/SpaceTimeChart';
 export * from './components/PathLayer';
 export * from './components/ConflictLayer';
 export * from './components/OccupancyBlockLayer';
+export * from './components/WorkScheduleLayer';
+export * from './components/PatternRect';
--- a/ui-spacetimechart/src/types.ts
+++ b/ui-spacetimechart/src/types.ts
@@ -1,0 +1,6 @@
+export type WorkSchedule = {
+  type: 'TRACK' | 'CATENARY';
+  timeStart: Date;
+  timeEnd: Date;
+  spaceRanges: [number, number][];
+};
--- a/ui-spacetimechart/src/assets/images/ScheduledMaintenanceUp.svg
+++ b/ui-spacetimechart/src/assets/images/ScheduledMaintenanceUp.svg
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="56px" height="56px" viewBox="0 0 56 56" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <title>Assets/Maintenance/ScheduledMaintenanceUp</title>
+    <g id="Assets/Maintenance/ScheduledMaintenanceUp" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <path d="M56,0 L56,28 L28,56 L0,56 L56,0 Z M28,0 L0,28 L0,0 L28,0 Z" id="Shape" fill="#F5F5F5"></path>
+    </g>
+</svg>
--- a/ui-spacetimechart/src/components/PatternRect.tsx
+++ b/ui-spacetimechart/src/components/PatternRect.tsx
@@ -1,0 +1,56 @@
+import { useCallback } from 'react';
+
+import { useDraw } from '../hooks/useCanvas';
+import { type DrawingFunction } from '../lib/types';
+
+export type PatternRectProps = {
+  timeStart: Date;
+  timeEnd: Date;
+  spaceStart: number; // mm
+  spaceEnd: number; // mm
+  imageElement: HTMLImageElement;
+};
+
+/**
+ * draws a repeating pattern in the space time chart
+ */
+export const PatternRect = ({
+  timeStart,
+  timeEnd,
+  spaceStart,
+  spaceEnd,
+  imageElement,
+}: PatternRectProps) => {
+  const drawRegion = useCallback<DrawingFunction>(
+    (ctx, { getSpacePixel, getTimePixel, spaceAxis }) => {
+      const timeStartPixel = getTimePixel(Number(timeStart));
+      const endTimePixel = getTimePixel(Number(timeEnd));
+      const spaceStartPixel = getSpacePixel(spaceStart);
+      const spaceEndPixel = getSpacePixel(spaceEnd);
+
+      const areaSpaceSize = spaceEndPixel - spaceStartPixel;
+      const areaTimeSize = endTimePixel - timeStartPixel;
+      if (!areaSpaceSize || !areaTimeSize) return;
+
+      const pattern = ctx.createPattern(imageElement, 'repeat');
+      if (!pattern) {
+        return;
+      }
+
+      ctx.save();
+      ctx.fillStyle = pattern;
+      if (spaceAxis === 'x') {
+        ctx.translate(spaceStartPixel, timeStartPixel);
+        ctx.fillRect(0, 0, areaSpaceSize, areaTimeSize);
+      } else {
+        ctx.translate(timeStartPixel, spaceStartPixel);
+        ctx.fillRect(0, 0, areaTimeSize, areaSpaceSize);
+      }
+      ctx.restore();
+    },
+    [timeStart, timeEnd, spaceStart, spaceEnd, imageElement]
+  );
+  useDraw('background', drawRegion);
+
+  return null;
+};
--- a/ui-spacetimechart/src/components/WorkScheduleLayer.tsx
+++ b/ui-spacetimechart/src/components/WorkScheduleLayer.tsx
@@ -1,0 +1,45 @@
+import React, { useEffect, useState } from 'react';
+
+import { PatternRect } from './PatternRect';
+import { type WorkSchedule } from '../types';
+
+type WorkScheduleLayerProps = {
+  workSchedules: WorkSchedule[];
+  imageUrl: string;
+};
+
+/**
+ * Displays the workschedule projection on the Space time chart.
+ * A workschedule has a start time and an end time.
+ * It can contain several portions of a track occupied during this period.
+ * Each portion is represented by a rectangle region
+ */
+export const WorkScheduleLayer = ({ workSchedules, imageUrl }: WorkScheduleLayerProps) => {
+  const [imageElement, setImageElement] = useState<HTMLImageElement>();
+  useEffect(() => {
+    if (imageUrl) {
+      const newImage = new Image();
+      newImage.src = imageUrl;
+      newImage.onload = () => {
+        setImageElement(newImage);
+      };
+    }
+  }, [imageUrl]);
+
+  if (!imageElement) {
+    return null;
+  }
+
+  return workSchedules.flatMap((ws) =>
+    ws.spaceRanges.map(([spaceStart, spaceEnd]) => (
+      <PatternRect
+        key={`${ws.type}-${ws.timeStart}-${ws.timeEnd}-${spaceStart}-${spaceEnd}`}
+        timeStart={ws.timeStart}
+        timeEnd={ws.timeEnd}
+        spaceStart={spaceStart}
+        spaceEnd={spaceEnd}
+        imageElement={imageElement}
+      />
+    ))
+  );
+};
--- a/ui-spacetimechart/src/config.ts
+++ b/ui-spacetimechart/src/config.ts
@@ -1,0 +1,4 @@
+declare module '*.svg' {
+  const content: string;
+  export default content;
+}
--- a/ui-spacetimechart/src/stories/work-schedules.stories.tsx
+++ b/ui-spacetimechart/src/stories/work-schedules.stories.tsx
@@ -1,0 +1,139 @@
+import React, { useState } from 'react';
+
+import '@osrd-project/ui-core/dist/theme.css';
+
+import type { Meta } from '@storybook/react';
+
+import { OPERATIONAL_POINTS, PATHS } from './lib/paths';
+import upward from '../assets/images/ScheduledMaintenanceUp.svg';
+import { PathLayer } from '../components/PathLayer';
+import { SpaceTimeChart } from '../components/SpaceTimeChart';
+import { WorkScheduleLayer } from '../components/WorkScheduleLayer';
+import { type Point, type PathData, type OperationalPoint } from '../lib/types';
+import { type WorkSchedule } from '../types';
+import { getDiff } from '../utils/vectors';
+
+const SAMPLE_WORK_SCHEDULES: WorkSchedule[] = [
+  {
+    type: 'TRACK',
+    timeStart: new Date('2024-04-02T00:00:00Z'),
+    timeEnd: new Date('2024-04-02T00:15:00Z'),
+    spaceRanges: [
+      [20000, 35000],
+      [45000, 60000],
+    ],
+  },
+  {
+    type: 'TRACK',
+    timeStart: new Date('2024-04-02T00:15:00Z'),
+    timeEnd: new Date('2024-04-02T01:00:00Z'),
+    spaceRanges: [
+      [80000, 100000],
+      [110000, 140000],
+    ],
+  },
+  {
+    type: 'TRACK',
+    timeStart: new Date('2024-04-02T01:30:00Z'),
+    timeEnd: new Date('2024-04-02T02:30:00Z'),
+    spaceRanges: [[50000, 100000]],
+  },
+];
+
+const DEFAULT_HEIGHT = 550;
+
+type WorkSchedulesWrapperProps = {
+  operationalPoints: OperationalPoint[];
+  paths: (PathData & { color: string })[];
+  workSchedules: WorkSchedule[];
+};
+
+const WorkSchedulesWrapper = ({
+  operationalPoints = [],
+  paths = [],
+  workSchedules,
+}: WorkSchedulesWrapperProps) => {
+  const [state, setState] = useState<{
+    xOffset: number;
+    yOffset: number;
+    panning: null | { initialOffset: Point };
+  }>({
+    xOffset: -300,
+    yOffset: 0,
+    panning: null,
+  });
+  const simpleOperationalPoints = operationalPoints.map(({ id, position }) => ({
+    id,
+    label: id,
+    position,
+  }));
+  const spaceScale = [
+    {
+      from: 0,
+      to: 75000,
+      coefficient: 300,
+    },
+  ];
+  return (
+    <div className="manchette-space-time-chart-wrapper" style={{ height: `${DEFAULT_HEIGHT}px` }}>
+      <SpaceTimeChart
+        className="inset-0 absolute h-full"
+        spaceOrigin={0}
+        xOffset={state.xOffset}
+        yOffset={state.yOffset}
+        timeOrigin={+new Date('2024/04/02')}
+        operationalPoints={simpleOperationalPoints}
+        timeScale={10000}
+        spaceScales={spaceScale}
+        onPan={({ initialPosition, position, isPanning }) => {
+          const diff = getDiff(initialPosition, position);
+          setState((s) => {
+            // Stop panning:
+            if (!isPanning) {
+              return { ...s, panning: null };
+            }
+            // Start panning:
+            else if (!s.panning) {
+              return {
+                ...s,
+                panning: {
+                  initialOffset: {
+                    x: s.xOffset,
+                    y: s.yOffset,
+                  },
+                },
+              };
+            }
+            // Keep panning:
+            else {
+              const { initialOffset } = s.panning;
+              return {
+                ...s,
+                xOffset: initialOffset.x + diff.x,
+                yOffset: initialOffset.y + diff.y,
+              };
+            }
+          });
+        }}
+      >
+        {paths.map((path) => (
+          <PathLayer key={path.id} path={path} color={path.color} />
+        ))}
+        <WorkScheduleLayer workSchedules={workSchedules} imageUrl={upward} />
+      </SpaceTimeChart>
+    </div>
+  );
+};
+
+export default {
+  title: 'SpaceTimeChart/Workschedules',
+  component: WorkSchedulesWrapper,
+} as Meta<typeof WorkSchedulesWrapper>;
+
+export const Default = {
+  args: {
+    operationalPoints: OPERATIONAL_POINTS,
+    paths: PATHS.slice(2, 4),
+    workSchedules: SAMPLE_WORK_SCHEDULES,
+  },
+};


Caracol3 (approved):
lgtm (tested). Even if it's only the beginning, it's cool to finally see the work schedules in the GET 😄.

ui-manchette-with-spacetimechart/src/stories/work-schedules.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+            spaceOrigin={0}
+            timeOrigin={Math.min(...projectPathTrainResult.map((p) => +new Date(p.departure_time)))}
+            {...spaceTimeChartProps}
+          >
+            {/* SpaceTimeChart is using mm (with spaceScales prop),
+             * so all the children need to provide mm too */}
+            {spaceTimeChartProps.paths.map((path) => (
+              <PathLayer key={path.id} path={path} color={path.color} />
+            ))}
+            {<WorkSchedules workSchedules={workSchedules} pattern={upward} />}
```
clarani:
```suggestion
            <WorkSchedules workSchedules={workSchedules} pattern={upward} />
```

ui-manchette-with-spacetimechart/src/stories/work-schedules.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+          style={{ bottom: 0, left: 0, top: 0, height: `${DEFAULT_HEIGHT - 6}px` }}
+        >
+          <SpaceTimeChart
+            className="inset-0 absolute h-full"
+            spaceOrigin={0}
+            timeOrigin={Math.min(...projectPathTrainResult.map((p) => +new Date(p.departure_time)))}
+            {...spaceTimeChartProps}
+          >
+            {/* SpaceTimeChart is using mm (with spaceScales prop),
+             * so all the children need to provide mm too */}
```
clarani: You can remove this comment but specify the unit of intersection in the type WorkSchedule directly
```suggestion
```

ui-spacetimechart/src/components/workSchedules/Region.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+import { useDraw } from '../../hooks/useCanvas';
+import { type DrawingFunction } from '../../lib/types';
+import { type SpaceTimeRegion } from '../../types';
+
+const Region = ({
+  timeStart, // x date timestamp
+  timeEnd, // x date timestamp
+  spaceStart, // y metre
+  spaceEnd, // y metre
```
clarani: Remove this comment since the chart can be rotated
```suggestion
  timeStart,
  timeEnd,
  spaceStart,
  spaceEnd,
```

ui-spacetimechart/src/components/workSchedules/Region.tsx
```diff
@@ -0,0 +1,62 @@
...
+  image,
+}: SpaceTimeRegion) => {
+  const [imageElement, setImageElement] = useState<HTMLImageElement | null>(null);
+  useEffect(() => {
```
clarani: Add blank lines to make the code easier to read
```suggestion
  const [imageElement, setImageElement] = useState<HTMLImageElement | null>(null);
  
  useEffect(() => {
```

ui-manchette-with-spacetimechart/src/assets/sampleData.ts
```diff
@@ -188,3 +189,24 @@ export const SAMPLE_PATHS_DATA: ProjectPathTrainResult[] = [
     ],
   },
 ];
+
+export const SAMPLE_WORK_SCHEDULES: WorkSchedule[] = [
+  {
+    type: 'TRACK',
+    timeStart: '2024-06-26T14:00:00Z',
+    timeEnd: '2024-06-26T14:15:00Z',
+    intersections: [
```
emersion: Nit: maybe `spaceRanges` would be a better name?

I'm a bit puzzled by "intersections", not sure what intersects what and I don't think the component should care which mathematical operation was performed to obtain this data.

ui-spacetimechart/src/components/workSchedules/Region.tsx
emersion: This stuff seems already documented in the type definition.

ui-spacetimechart/src/components/workSchedules/Region.tsx
```diff
@@ -0,0 +1,62 @@
+import { useCallback, useEffect, useState } from 'react';
+
+import { useDraw } from '../../hooks/useCanvas';
+import { type DrawingFunction } from '../../lib/types';
+import { type SpaceTimeRegion } from '../../types';
+
+const Region = ({
```
emersion: I wonder if we can find a better name, "Region" seems a bit generic to me. What this component truly does is paint a rectangle with an image pattern. What about "PatternRect"?

ui-spacetimechart/src/components/workSchedules/Region.tsx
```diff
@@ -0,0 +1,62 @@
...
+  timeStart, // x date timestamp
+  timeEnd, // x date timestamp
+  spaceStart, // y metre
+  spaceEnd, // y metre
+  image,
+}: SpaceTimeRegion) => {
+  const [imageElement, setImageElement] = useState<HTMLImageElement | null>(null);
+  useEffect(() => {
+    if (image) {
+      const newImage = new Image();
```
emersion: Nit: `newImage` doesn't really convey the difference between this variable and `image`. `image` is the URL and `newImage` is the `Image` object. Maybe rename `image` to `imageUrl` instead?

ui-spacetimechart/src/types.ts
```diff
@@ -0,0 +1,15 @@
+export type SpaceTimeRegion = {
+  type?: string;
```
emersion: This seems unused.

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Display the workschedule projection on the Space time chart.
+ * a workschedule can contains several track occupied during a period of this
+ * each is represented by a rectangle region
+ */
+function WorkSchedules({ workSchedules, pattern }: WorkSchedulesProps) {
+  return workSchedules.flatMap((ws) =>
+    ws.intersections.map(([spaceStart, spaceEnd]) => (
+      <Region
+        key={`${ws.timeStart}-${spaceStart}`}
```
emersion: I'm worried we might end up with duplicate keys here.

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
```diff
@@ -0,0 +1,30 @@
...
+};
+/**
+ * Display the workschedule projection on the Space time chart.
+ * a workschedule can contains several track occupied during a period of this
+ * each is represented by a rectangle region
+ */
+function WorkSchedules({ workSchedules, pattern }: WorkSchedulesProps) {
+  return workSchedules.flatMap((ws) =>
+    ws.intersections.map(([spaceStart, spaceEnd]) => (
+      <Region
```
emersion: Is it fine to create one separate component per rect? Wouldn't it be better to pass a list of rects to a component responsible for drawing the pattern?

That way, we wouldn't end with a big list of components in the DOM and we wouldn't load the pattern image each time we draw an individual rect.

ui-spacetimechart/src/types.ts
```diff
@@ -0,0 +1,15 @@
...
+  timeEnd: number; // timestamp
+  spaceStart: number; // mm
+  spaceEnd: number; // mm
+  image?: string;
+};
+
+export type WorkSchedule = {
+  type: 'TRACK' | 'CATENARY';
+  timeStart: string;
+  timeEnd: string;
```
emersion: Could we define these as `Date` instead of `string`?

ui-manchette-with-spacetimechart/src/config.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.svg' {
+  const content: string;
+  export default content;
+}
```
emersion: I'm not familiar with this kind of declaration, what does this achieve? Are these instructions for the build system?

ui-spacetimechart/src/components/workSchedules/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as WorkSchedules } from './WorkSchedules';
+export { default as Region } from './Region';
```
emersion: Is `Region` something we really want to expose to library users and maintain as public API?

ui-spacetimechart/src/components/workSchedules/Region.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+      const pattern = ctx.createPattern(imageElement, 'repeat');
+      if (!pattern) {
+        return;
+      }
+
+      ctx.save();
+      ctx.fillStyle = pattern;
+      if (spaceAxis === 'x') {
+        ctx.translate(spaceStartPixel, timeStartPixel);
```
emersion: Is the `translate()` call necessary? Could we just pass the x/y as the first argument to `fillRect()`?

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
```diff
@@ -0,0 +1,30 @@
...
+type WorkSchedulesProps = {
+  workSchedules: WorkSchedule[];
+  pattern: string;
+};
+/**
+ * Display the workschedule projection on the Space time chart.
+ * a workschedule can contains several track occupied during a period of this
+ * each is represented by a rectangle region
+ */
+function WorkSchedules({ workSchedules, pattern }: WorkSchedulesProps) {
```
clarani: Use const since it is a component

ui-spacetimechart/src/components/workSchedules/Region.tsx
anisometropie: It’s necessary, we want the upper left corner of the svg to be 0,0. otherwise the diagonal stripes move with the chart when we scroll the chart horizontally. I think it has to do with the way `ctx.createPattern` works.

ui-spacetimechart/src/components/workSchedules/index.ts
anisometropie: yeah, I guess we should not. removed.

ui-manchette-with-spacetimechart/src/config.ts
anisometropie: TypeScript doesn’t recognise the import otherwise, even though it’s correctly imported at runtime.

ui-spacetimechart/src/types.ts
anisometropie: done

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
anisometropie: I think this is already pretty optimised. Region doesn’t create anything in the DOM (return `null`). It just calls the canvas context apis.

emersion: Indeed. Still, many `Image` objects are created, each resulting in an HTTP request (potentially cached) and SVG parsing etc.

ui-manchette-with-spacetimechart/src/assets/sampleData.ts
anisometropie: You’re right, changed that.

ui-spacetimechart/src/components/workSchedules/Region.tsx
anisometropie: removed

anisometropie: done

ui-spacetimechart/src/components/workSchedules/Region.tsx
anisometropie: done

ui-spacetimechart/src/types.ts
```diff
@@ -0,0 +1,15 @@
...
+  spaceStart: number; // mm
+  spaceEnd: number; // mm
+  image?: string;
+};
+
+export type WorkSchedule = {
+  type: 'TRACK' | 'CATENARY';
+  timeStart: Date;
+  timeEnd: Date;
+  spaceRanges: [number, number][];
```
clarani, line 5: Why spaceRange is `[number, number][]` and not `[number, number]` ?

ui-spacetimechart/src/types.ts
```diff
@@ -0,0 +1,15 @@
+export type SpaceTimeRegion = {
+  type?: string;
+  timeStart: Date; // timestamp
+  timeEnd: Date; // timestamp
```
clarani:
```suggestion
  timeStart: Date;
  timeEnd: Date;
```

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
```diff
@@ -0,0 +1,30 @@
...
+import Region from './Region';
+import { type WorkSchedule } from '../../types';
+
+type WorkSchedulesProps = {
+  workSchedules: WorkSchedule[];
+  pattern: string;
+};
+/**
+ * Display the workschedule projection on the Space time chart.
+ * a workschedule can contains several track occupied during a period of this
```
clarani: - several typos
- I think your sentence is not finished ? (this -> time ?)
```suggestion
 * A workschedule can contain several tracks occupied during a period of this
```

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
```diff
@@ -0,0 +1,30 @@
...
+ * each is represented by a rectangle region
```
clarani:
```suggestion
 * Each work schedule is represented by a rectangle region
```

ui-spacetimechart/src/components/workSchedules/Region.tsx
```diff
@@ -0,0 +1,57 @@
+import { useCallback, useEffect, useState } from 'react';
+
+import { useDraw } from '../../hooks/useCanvas';
+import { type DrawingFunction } from '../../lib/types';
+import { type SpaceTimeRegion } from '../../types';
+
+const Region = ({ timeStart, timeEnd, spaceStart, spaceEnd, image }: SpaceTimeRegion) => {
```
clarani: I don't really like the term Region, because it's not only a space object, but a space object during a certain period of time. 
SpaceTimeRectangle might be a better name

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
clarani, line 1: Do we really need this layer ? I think that exposing the Region layer is enough and more generic

ui-spacetimechart/src/types.ts
```diff
@@ -0,0 +1,15 @@
...
+  spaceStart: number; // mm
+  spaceEnd: number; // mm
+  image?: string;
```
clarani: - should not be possibly undefined
- this image is really the image itself ? Or the url of the image ?
```suggestion
  image: string;
```

ui-manchette-with-spacetimechart/src/config.ts
clarani: So you transform the svg in string for typescript ?

clarani (commented):
- I think it would be nice to add a story in ui-spacetimechart, and to add some workSchedules in the rotated story

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
clarani: Ah yes, good point. I agree with @emersion

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
anisometropie: Not sure if we should expose `Region` directly.
https://github.com/OpenRailAssociation/osrd-ui/pull/523#discussion_r1786265573

emersion: I think we have two choices:

- Either we expose a `WorkSchedule` layer. This would make sense if we're planning in the future to add more work schedule related stuff, e.g. a tooltip or fancier decorations or more detailed info?
- Or we expose a generic layer to paint a pattern on a region. This would make sense if osrd never needs to do anything more than painting a pattern.

anisometropie:
it is actually already a string when you do `import upward from '../assets/images/ScheduledMaintenanceUp.svg';` 
looks like this
`/@fs/home/ut/soft/Programmation/javascript/code/osrd-ui/ui-manchette-with-spacetimechart/src/assets/images/ScheduledMaintenanceUp.svg`

Here it’s just avoid getting typescrit errors, there seem to be so config missing

ui-spacetimechart/src/types.ts
anisometropie: it’s the URL. renamed to imageUrl, and made it not optional

ui-spacetimechart/src/components/workSchedules/Region.tsx
anisometropie: done

ui-spacetimechart/src/types.ts
anisometropie: removed

ui-spacetimechart/src/components/workSchedules/Region.tsx
anisometropie: SpaceTimeRectangle it is :)

emersion: Hm, this doesn't reflect the fact that it takes an image and repeats it in a pattern. Also a rect in a space time chart has necessarily one axis for space and one axis for time, not sure it's useful to repeat it.

Ref https://github.com/OpenRailAssociation/osrd-ui/pull/523#discussion_r1786253686

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
anisometropie: I reworked all this discription

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
anisometropie: done

ui-spacetimechart/src/types.ts
anisometropie: done (moved to SpaceTimeRectangle.tsx)

ui-spacetimechart/src/types.ts
anisometropie: you have several portions blocked for the same period of time. if the track goes from 0 to 100 for example, you can have [[10,20], [50, 80]].

it’s also defined like this in editoast

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
anisometropie: done

ui-manchette-with-spacetimechart/src/stories/work-schedules.stories.tsx
anisometropie: The intent of the stories are to provide some documentation, I don’t think it hurts to have this here. it’s not trivial to see.

ui-manchette-with-spacetimechart/src/stories/work-schedules.stories.tsx
anisometropie: done

ui-spacetimechart/src/components/workSchedules/Region.tsx
anisometropie: done https://github.com/OpenRailAssociation/osrd-ui/pull/523#discussion_r1787379091

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
anisometropie: added some more stuff.

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
anisometropie: I moved the image loading to the parent

anisometropie: I moved the image loading to the parent @emersion @clarani 
what do you think?

ui-manchette-with-spacetimechart/src/config.ts
anisometropie: it is actually already a string when you do import upward from '../assets/images/ScheduledMaintenanceUp.svg';
looks like this
/@fs/home/ut/soft/Programmation/javascript/code/osrd-ui/ui-manchette-with-spacetimechart/src/assets/images/ScheduledMaintenanceUp.svg

ui-spacetimechart/src/components/workSchedules/Region.tsx
anisometropie: Seeing this now, as it’s in the spacetime chart folder,  I tend more towards `PatternRect`

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
anisometropie: There is a high probability `WorkSchedule` will have more stuff in the future and that we, or someone using osrd-ui in the future, might want to draw something on the graph that is not a `WorkSChedule`
In any case, I think this is way beyond the scope of this ticket, we need to dedicate a work task to this component design

ui-spacetimechart/src/types.ts
clarani: Ok clear, thanks :pray:

ui-spacetimechart/src/components/workSchedules/Region.tsx
clarani: Fine for me :+1:

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
clarani: I think we can have a 3rd option: expose both the `WorkSchedule` layer and the generic layer.
We will have to add more work schedule related stuff (the tickets are already ready in the PM board), so we will definitely have to keep the `WorkSchedule` layer.

I think we can keep the 2 layers

ui-manchette-with-spacetimechart/src/assets/sampleData.ts
clarani, line 1: I think that this story should be in ui-manchette and not in ui-manchette-with-spacetimechart

ui-spacetimechart/src/components/workSchedules/SpaceTimeRectangle.tsx
```diff
@@ -0,0 +1,55 @@
+import { useCallback, useEffect, useState } from 'react';
```
clarani:
```suggestion
import { useCallback, useState } from 'react';
```

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
```diff
@@ -0,0 +1,47 @@
...
+};
+
+/**
+ * Displays the workschedule projection on the Space time chart.
+ * A workschedule has a start time and an end time.
+ * It can contain several portions of a track occupied during this period.
+ * Each portion is represented by a rectangle region
+ */
+const WorkSchedules = ({ workSchedules, imageUrl }: WorkSchedulesProps) => {
+  const [imageElement, setImageElement] = useState<HTMLImageElement | null>(null);
```
clarani: More concise like this
```suggestion
  const [imageElement, setImageElement] = useState<HTMLImageElement>();
```

ui-spacetimechart/src/index.ts
```diff
@@ -3,3 +3,6 @@ import './styles/main.css';
 export * from './components/SpaceTimeChart';
 export * from './components/PathLayer';
 export * from './components/ConflictLayer';
+export * from './hooks/useCanvas';
+export * from './components/workSchedules';
```
clarani: I'm not sure that we need a subFolder WorkSchedule.
Here you can expose WorkSchedule and PatternRect.
```suggestion
export * from './components/WorkSchedules';
export * from './components/PatternRect';
```

ui-spacetimechart/src/index.ts
```diff
@@ -3,3 +3,6 @@ import './styles/main.css';
 export * from './components/SpaceTimeChart';
 export * from './components/PathLayer';
 export * from './components/ConflictLayer';
+export * from './hooks/useCanvas';
```
clarani: Why do you expose useCanvas ? I don't think that it is necessary

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
```diff
@@ -0,0 +1,47 @@
...
+  imageUrl: string;
+};
+
+/**
+ * Displays the workschedule projection on the Space time chart.
+ * A workschedule has a start time and an end time.
+ * It can contain several portions of a track occupied during this period.
+ * Each portion is represented by a rectangle region
+ */
+const WorkSchedules = ({ workSchedules, imageUrl }: WorkSchedulesProps) => {
```
clarani: It's a suggestion, to harmonize the layers name
```suggestion
const WorkScheduleLayer = ({ workSchedules, imageUrl }: WorkSchedulesProps) => {
```

ui-spacetimechart/src/components/workSchedules/Region.tsx
anisometropie: PatternRect it is

ui-spacetimechart/src/components/workSchedules/SpaceTimeRectangle.tsx
anisometropie: done

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
anisometropie: done

ui-spacetimechart/src/index.ts
anisometropie: alright done

ui-spacetimechart/src/index.ts
anisometropie: yes, I removed

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
anisometropie: done

ui-manchette-with-spacetimechart/src/assets/sampleData.ts
anisometropie: why  ui-manchette ?

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
emersion: I _think_ this is fine for the purposes of work schedules, but if we're going to expose `PatternRect` as a generic layer in the future I don't think this is something we can assume. We wouldn't need keys if the layer was responsible for painting multiple rects instead of a single one.

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
emersion: I think the performance concerns are addressed. I still believe a layer painting a region is a better design than a layer painting a single rect (see https://github.com/OpenRailAssociation/osrd-ui/pull/523#discussion_r1801149689 for another reason why) but since it's not public API I'm fine with keeping it as-is.

ui-manchette-with-spacetimechart/src/assets/sampleData.ts
emersion: Maybe it should be in ui-spacetimechart? But yeah, probably not in ui-manchette :P

ui-spacetimechart/src/types.ts
```diff
@@ -0,0 +1,6 @@
+export type WorkSchedule = {
+  type: 'TRACK' | 'CATENARY';
```
emersion, line 2: This seems unused still, but I reckon we'll want to use this for the tooltip in the future? Since it's well-defined I don't mind introducing it in this PR.

ui-spacetimechart/src/assets/workScheduleSampleData.ts
```diff
@@ -0,0 +1,212 @@
+/* eslint-disable import/no-unresolved */
+import {
+  type PathProperties,
+  type ProjectPathTrainResult,
+} from '@osrd-project/ui-manchette/dist/types';
+import { type WorkSchedule } from '@osrd-project/ui-spacetimechart/dist/types';
+
+export const SAMPLE_PATH_PROPERTIES_DATA: PathProperties = {
```
emersion: I'd slightly prefer _not_ having this in `assets/` since it's not images/CSS/etc that need to be included in the final NPM package and it's only used in the story. But it seems like we already do the same thing for ui-manchette-with-spacetimechart. Feel free to resolve this comment.

ui-manchette-with-spacetimechart/src/assets/sampleData.ts
anisometropie: Moving to ui-spacetimechart :)

ui-spacetimechart/src/types.ts
```diff
@@ -0,0 +1,63 @@
...
+  timeStart: Date;
+  timeEnd: Date;
+  spaceRanges: [number, number][];
+};
+
+export type OperationalPointType = ArrayElement<PathProperties['operational_points'] | null>;
```
emersion: Hm, seems like all of the types below are re-defined? They're already in ui-manchette, but they're not used in ui-spacetimechart?

emersion:
I'm trying to open the storybook in this branch, but for the life of me I can't manage to get past this error:

> The requested module 'http://localhost:6006/node_modules/.cache/storybook/ee6bf224cd156d2a19aeb1406fef9897dbfe874b3eae8c467c24b2f62e4ecd1a/sb-vite/deps/@osrd-project_ui-manchette-with-spacetimechart.js?v=c3b63b3c' doesn't provide an export named: 'computeTimeWindow'

Tried re-building everything, deleting all of the `dist/` directories, and nuking `node_modules/.cache/`.

ui-spacetimechart/src/types.ts
emersion: BTW, `ui-spacetimechart/src/stories/lib/paths.ts` already has sample trains and OPs. Maybe would make sense to use that. Also, maybe it would make sense to integrate this new story into `ui-spacetimechart/src/stories/layers.stories.tsx`.

ui-spacetimechart/src/components/workSchedules/Region.tsx
emersion: WFM!

ui-spacetimechart/src/assets/workScheduleSampleData.ts
anisometropie: agreed, keeping that in the stories is better

ui-spacetimechart/src/types.ts
anisometropie: alright, since it is also defined in editoast I think we can leave it

ui-spacetimechart/src/types.ts
anisometropie: I tidied up all this, only using the existing spacetime types for this story

clarani (approved):
LGTM :white_check_mark:

ui-spacetimechart/src/index.ts
```diff
@@ -4,3 +4,6 @@ export * from './components/SpaceTimeChart';
 export * from './components/PathLayer';
 export * from './components/ConflictLayer';
 export * from './components/OccupancyBlockLayer';
+export * from './components/WorkScheduleLayer';
+export * from './components/PatternRect';
+export * from './types';
```
clarani: Not sure it's needed
```suggestion
```

anisometropie: it was not, it’s removed

ui-spacetimechart/src/components/workSchedules/WorkSchedules.tsx
anisometropie: Merging for now, we can rework this when need arises

PR #538, opened 2024-09-24
ui-core: nuke value in Select

emersion:
We don't really need a value here, we can just use indexes instead. This makes it easier to use Select for small lists (e.g. 3 static choices) and simplifies the code.

Discussion: https://github.com/OpenRailAssociation/osrd/pull/8838#discussion_r1766745894

### DIFF
--- a/ui-core/src/components/Select.tsx
+++ b/ui-core/src/components/Select.tsx
@@ -12,11 +12,8 @@ export type SelectProps<T> = Omit<
     options: Array<T>;
     value: T;
     getOptionLabel: (option: T) => string;
-    getOptionValue: (option: T) => string;
     onChange: (option?: T) => void;
   };
-
-const PLACEHOLDER_VALUE = '__PLACEHOLDER__';
 
 const Select = <T,>({
   id,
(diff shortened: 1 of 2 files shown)


ui-core/src/components/Select.tsx
```diff
@@ -64,16 +58,14 @@ const Select = <T,>({
...
         disabled={disabled || readOnly}
         onChange={handleOnChange}
       >
-        {placeholder && <option value={PLACEHOLDER_VALUE}>{`– ${placeholder} –`}</option>}
+        {placeholder && <option>{`– ${placeholder} –`}</option>}
         {options.map((option) => (
-          <option key={getOptionValue(option)} value={getOptionValue(option)}>
-            {getOptionLabel(option)}
-          </option>
+          /* eslint-disable-next-line react/jsx-key */
```
emersion, line 67: We loose the `key` here, but since these lists are quite short I don't believe it matters.

SharglutDev: So we will have a warning in console each time we use the `Select` component ?

ui-core/src/components/Select.tsx
```diff
@@ -64,16 +58,14 @@ const Select = <T,>({
...
+          <option selected={option === selectedOption}>{getOptionLabel(option)}</option>
```
kmer2016, line 68: I noticed that the comparison `option === selectedOption` in the Select component may not work as expected when T is an object. because objects in JavaScript are compared by reference, not by their content.

For example, if `selectedOption` is stored in a Redux store, it could result in a different reference than the one in the options array, even if the objects contain the same data. In this case, the select would not highlight the correct option, since the comparison `option === selectedOption` would return `false`

emersion: Yeah, that's a deliberate design choice in this PR. The `value` prop needs to be one of the elements from the possible options.

ui-core/src/components/Select.tsx
anisometropie: That’s a generic component, We don’t know who is going to use it, and it could be a select of any size.
Besides, I think even with smaller lists react can get confused and mix up elements between renders if you don’t use keys.

emersion:
I don't have time to push this forward, let's drop this for now.

PR #547, opened 2024-09-24
ui-core: enable the user to type in the date picker

### LINKED ISSUE #526: ui-core: enable to type in the datePicker
### Description and goal

Enable to type in the date Picker

### Acceptance criteria

A list of conditions which define when this task shall be considered complete

clarani:
closes #526 

Some remarks:
- when the user clicks on the input, the calendar is not opened automatically anymore
- if the user types a string which is not a date, the date value is not updated
- if the user types a valid string but the date is not in the selectable slot, then an error is displayed (the invalid date is not updated in the parent component)
- if the calendar is opened and the user enters a valid date, then the calendar view is correctly updated
- if an invalid date is entered and the user picks a date from the calendar, then the error is not displayed anymore

Technical remarks:
- I moved the selectableSlot props so it is more easier to access it

When the input is invalid:
image

### DIFF
--- a/ui-core/src/components/inputs/datePicker/CalendarPicker.tsx
+++ b/ui-core/src/components/inputs/datePicker/CalendarPicker.tsx
@@ -15,12 +15,12 @@ export type CalendarPickerPrivateProps = {
     left: number;
   };
   calendarPickerRef: React.RefObject<HTMLDivElement>;
+  selectableSlot?: CalendarSlot;
 };
 
 export type CalendarPickerPublicProps = {
   initialDate?: Date;
   numberOfMonths?: 1 | 2 | 3;
-  selectableSlot?: CalendarSlot;
 };
 
 export type CalendarPickerProps = CalendarPickerPrivateProps & CalendarPickerPublicProps;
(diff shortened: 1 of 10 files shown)


achrafmohye (approved):
LGTM and tested 👍

kmer2016 (commented):
Nice work !

Please, could you update tests for `handleInputClick` and add tests for `handleInputOnChange` ?

I noticed a strange behavior with the date validation:

1. When I enter a date that is out of the allowed range, the date picker shows the error message correctly.
2. If I then type something that is not a date (e.g., "toto"), the date picker stays in error mode, which is expected.
3. However, when I enter a valid date afterward, the date picker still shows as invalid, even though it should reset.

Also, when I type a non-date string (e.g., "toto") from the beginning, the input does not show an invalid message.

clarani:
Thanks for your review, I modified a bit the behaviors:
- if the user enters a string which does not contain only slashes and numbers, then an error is displayed (example `toto`)
- if the user enters a string which contains only slashes and numbers but which is longer than 8 characters, then an error is displayed (for instance `01/01/2024`)

kmer2016 (approved):
Good job !

PR #557, opened 2024-09-26
ui-core: accept undefined as Select value

emersion:
When a placeholder is used, the value of a Select can be undefined, as reflected in the type of the onChange callback. However the component user had no way to reset the value of the Select from the outside via the value property. Make it optional to fix that.

Discussion: https://github.com/OpenRailAssociation/osrd/pull/8838#discussion_r1774839347


kmer2016 (approved):
Thank you for this PR

PR #581, opened 2024-09-30
speedspacechart: fix boundaries iteration

flomonster:
We can not modify elements from the state


clarani (approved):
LGTM :white_check_mark:

PR #597, opened 2024-10-02
Fix speed space chart curves color

flomonster:
## Before

image

## Now

image


theocrsb (commented):
I've selected the `base speed curve` on the mockup and I see an alpha of 0.5 ?

flomonster:
> I've selected the `base speed curve` on the mockup and I see an alpha of 0.5 ?

We have for now only two curves to show. Eco and base, the `0.5` is used for a third curve which is not implemented yet.

theocrsb (approved):
LGTM and tested

SharglutDev (approved):
Lgtm and tested, thank you

PR #619, opened 2024-10-06
ui-spacetimechart: add occupancy blocks layer

### LINKED ISSUE #9158: UI: Display occupancy blocks in GET
Image

Akctarus:
closes https://github.com/OpenRailAssociation/osrd/issues/9158

### DIFF
--- a/ui-spacetimechart/src/components/OccupancyBlockLayer.tsx
+++ b/ui-spacetimechart/src/components/OccupancyBlockLayer.tsx
@@ -1,0 +1,37 @@
+import { useCallback } from 'react';
+
+import { useDraw } from '../hooks/useCanvas';
+import { type DrawingFunction } from '../lib/types';
+
+export type OccupancyBlock = {
+  timeStart: number;
+  timeEnd: number;
+  spaceStart: number;
+  spaceEnd: number;
+  color: string;
+};
+
+export type OccupancyBlockLayerProps = {
+  occupancyBlocks: OccupancyBlock[];
+};
+
+export const OccupancyBlockLayer = ({ occupancyBlocks }: OccupancyBlockLayerProps) => {
+  const drawOccupancyBlockLayer = useCallback<DrawingFunction>(
+    (ctx, { getTimePixel, getSpacePixel }) => {
+      for (const occupancyBlock of occupancyBlocks) {
+        const x = getTimePixel(occupancyBlock.timeStart);
+        const y = getSpacePixel(occupancyBlock.spaceStart);
+        const width = getTimePixel(occupancyBlock.timeEnd) - x;
+        const height = getSpacePixel(occupancyBlock.spaceEnd) - y;
+
+        ctx.fillStyle = occupancyBlock.color;
+        ctx.fillRect(x, y, width, height);
+      }
+    },
+    [occupancyBlocks]
+  );
+
+  useDraw('background', drawOccupancyBlockLayer);
+
+  return null;
+};
--- a/ui-spacetimechart/src/stories/layers.stories.tsx
+++ b/ui-spacetimechart/src/stories/layers.stories.tsx
@@ -4,8 +4,14 @@ import type { Meta, StoryObj } from '@storybook/react';
 
 import { OPERATIONAL_POINTS, PATHS, START_DATE } from './lib/paths';
 import { X_ZOOM_LEVEL, Y_ZOOM_LEVEL } from './lib/utils';
-import { ConflictLayer, SpaceTimeChart, PathLayer } from '../';
-import { KILOMETER, MINUTE } from '../lib/consts';
+import { ConflictLayer, OccupancyBlockLayer, SpaceTimeChart, PathLayer } from '../';
+import {
+  KILOMETER,
+  MINUTE,
+  OCCUPANCY_FREE,
+  OCCUPANCY_SEMAPHORE,
+  OCCUPANCY_WARNING,
+} from '../lib/consts';
 
 import '@osrd-project/ui-spacetimechart/dist/theme.css';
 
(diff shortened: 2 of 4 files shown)


ui-spacetimechart/src/components/OccupancyBlocLayer.tsx
```diff
@@ -0,0 +1,39 @@
...
+  timeEnd: number;
+  spaceStart: number;
+  spaceEnd: number;
+};
+
+export type OccupancyBlocLayerProps = {
+  occupancyBlocs: OccupancyBloc[];
+};
+
+const SIGNALS = ['#CAEDDB', '#FFD4D8', '#FFEABF'];
```
Yohh: could you export well named variables from `/src/lib/consts.ts` ?

emersion: I'd personally prefer to keep these in this file, since it's really only related to the occupancy layer and not used by anything else. But maybe I'm just not used to the existing practices in osrd-ui.

emersion:
"Bloc" seems… very French :P 

Should it be "block"? Or is there a better English translation for "bloc d'occupation"?

Edit: reading [the docs](https://osrd.fr/en/docs/reference/design-docs/signaling/), seems like "block" is the right word to use here.

ui-spacetimechart/src/components/OccupancyBlocLayer.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const drawOccupancyBlocLayer = useCallback<DrawingFunction>(
+    (ctx, { getTimePixel, getSpacePixel }) => {
+      for (let i = 0; i < occupancyBlocs.length; i++) {
+        const occupancyBloc = occupancyBlocs[i];
+        const x = getTimePixel(occupancyBloc.timeStart);
+        const y = getSpacePixel(occupancyBloc.spaceStart);
+        const width = getTimePixel(occupancyBloc.timeEnd) - x;
+        const height = getSpacePixel(occupancyBloc.spaceEnd) - y;
+
+        ctx.fillStyle = SIGNALS[i % SIGNALS.length];
```
emersion: Is there some kind of a guarantee that the blocks are ordered by signal state? If so, it doesn't seem straightforward so we should document it. But would it make sense to add a `state` property to `OccupancyBloc` instead?

ui-spacetimechart/src/components/OccupancyBlocLayer.tsx
Yohh: I think that `occupancySemaphore` is far more readable than `#FFD4D8`, from now we have three colors for three signals, in the future this will grow

but if you think it's better that way we can keep it in mind and see later

emersion:
I'd also recommend not doing the same mistake I did :P https://github.com/OpenRailAssociation/osrd-ui/pull/624

ui-spacetimechart/src/components/OccupancyBlocLayer.tsx
emersion: My reaction was about the bit to move this to a separate file only.

I agree that the index is not very descriptive, maybe an enum would be more explicit? Also see related discussion in https://github.com/OpenRailAssociation/osrd-ui/pull/619#discussion_r1791930112.

ui-spacetimechart/src/components/OccupancyBlocLayer.tsx
Yohh: it makes sense for those three signals specifically, red -> yellow -> green -> red ... in a normal situation, ~~but we can also add some other signals like blinking-yellow that doesn't follow this logic~~ 

edit, even with the blinking-yellow it will follow those steps, it seems ok to me

ui-spacetimechart/src/components/OccupancyBlocLayer.tsx
Yohh: ah yes, np, keeping it in this file seems ok, we should just rename it to be more explicit

ui-spacetimechart/src/components/OccupancyBlocLayer.tsx
emersion: If another train crosses the projected path and we want to show that train's occupancy blocks, we might only have one yellow/red block, I think?

Yohh: sure, but the order will be the same

emersion:
Do we have a mockup for this?

Akctarus:
> Do we have a mockup for this?

Yes : 
336398851-a2fe7587-e43f-4591-8add-88454b3bcaf4

and : https://www.sketch.com/s/bcbac4ff-7f74-4f1a-9e10-08d115c47ed8/p/C0FA571F-D55C-40A1-B0E0-B61D135726CD/canvas

ui-spacetimechart/src/components/OccupancyBlocLayer.tsx
Akctarus: fixed !

Akctarus:
> "Bloc" seems… very French :P
> 
> Should it be "block"? Or is there a better English translation for "bloc d'occupation"?
> 
> Edit: reading [the docs](https://osrd.fr/en/docs/reference/design-docs/signaling/), seems like "block" is the right word to use here.

You're right, done fixing !

ui-spacetimechart/src/components/OccupancyBlocLayer.tsx
Akctarus: fixed !

Akctarus:
> I'd also recommend not doing the same mistake I did :P #624

Done fixing

ui-spacetimechart/src/components/OccupancyBlockLayer.tsx
```diff
@@ -0,0 +1,45 @@
+import { useCallback } from 'react';
+
+import { useDraw } from '../hooks/useCanvas';
+import { type DrawingFunction } from '../lib/types';
+
+// Should we keep it ?
+export enum SignalType {
```
emersion: If it's unused, I'd say nuke it - we can reintroduce it easily if/when we need it.

ui-spacetimechart/src/components/OccupancyBlockLayer.tsx
```diff
@@ -0,0 +1,45 @@
...
+};
+
+export type OccupancyBlockLayerProps = {
+  occupancyBlocks: OccupancyBlock[];
+};
+
+export const OccupancyBlockLayer = ({ occupancyBlocks }: OccupancyBlockLayerProps) => {
+  const drawOccupancyBlockLayer = useCallback<DrawingFunction>(
+    (ctx, { getTimePixel, getSpacePixel }) => {
+      for (let i = 0; i < occupancyBlocks.length; i++) {
```
emersion: Could use a `for (const occupancyBlock of occupancyBlocks)` loop here instead.

emersion (commented):
Apart from these two minor comments, LGTM!

ui-spacetimechart/src/components/OccupancyBlockLayer.tsx
Akctarus: good point, done

ui-spacetimechart/src/components/OccupancyBlockLayer.tsx
Akctarus: done !

emersion (approved):

ui-spacetimechart/src/stories/layers.stories.tsx
```diff
@@ -30,6 +30,51 @@ const CONFLICTS = [
   },
 ];
 
+const OCCUPANCY_BLOCKS = [
+  {
+    timeStart: +START_DATE + 42 * MINUTE,
+    timeEnd: +START_DATE + 43 * MINUTE,
+    spaceStart: 10 * KILOMETER,
+    spaceEnd: 14 * KILOMETER,
+    color: '#CAEDDB',
```
Yohh: so no explicit color name?

Math-R: Yes as it is repeated several time, it should be in variable for maintainability, even if it's a story, it's an example of implemntation that we or anybody can use

Akctarus:
```
signal_updates: {
   ...
   The color of the aspect (Bits 24-31 are alpha, 16-23 are red, 8-15 are green, 0-7 are blue) */
    color: number;
   ...
  }[];
```
It will be only useful for this repo since colors don't have explicit names returned by the back either. But if it improves the clarity of the code I can add it.

Yohh: we should really consider the stories like documentation

Akctarus: How should I name them ? 
- occupied : SEMAPHORE
- free : ?
- warning : ?

Yohh: image

Akctarus: done

Yohh (approved):

Math-R (approved):
lgtm

PR #620, opened 2024-10-07
github: switch dependabot to weekly, add storybook group

emersion:
Daily upgrades are a bit too spammy and we don't need to be _that_ up-to-date, so switch to weekly updates.

Group all storybook upgrades altogether in a single PR.


SharglutDev (approved):
Feature of the year

woshilapin (approved):

Khoyo (approved):

PR #621, opened 2024-10-07
ui-manchette-with-spacetimechart: fix spacetimechart scaling if hidde…

SharglutDev:
…n waypoints

Fix an issue where the space time chart wouldn't correspond to the manchette in linear mode when a bunch of waypoints were hidden in the middle of the path
> add a new condition in `getScales` function

Fix an issue where the space time chart wouldn't correspond to the manchette when the first x waypoints were hidden
> The code used to consider that the space origin only starts at 0 but it might not always be the case when hiding waypoints.

> [!IMPORTANT]  
> To test these changes, you need to be able to launch osrd-ui inside osrd and be on this [PR](https://github.com/OpenRailAssociation/osrd/pull/8796) (to hide waypoints on the manchette)

### DIFF
--- a/ui-manchette-with-spacetimechart/src/__tests__/helpers.spec.ts
+++ b/ui-manchette-with-spacetimechart/src/__tests__/helpers.spec.ts
@@ -221,7 +221,7 @@ describe('getScales', () => {
       yZoom: 1,
     });
 
-    expect(result).toHaveLength(1);
+    expect(result).toHaveLength(2);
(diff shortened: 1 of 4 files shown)


clarani (commented):
LGTM, can you separate your commit in 2 ? 1 for the SpaceTimeChart, 1 for the Manchette

clarani (approved):
LGTM :white_check_mark:

PR #624, opened 2024-10-08
ui-spacetimechart: add missing ConflictLayer export

emersion:
We're not able to use the component from outside if it's not exported in index.ts.

Oversight on my end, sorry!


clarani (approved):
LGTM :white_check_mark:

theocrsb (approved):

PR #627, opened 2024-10-09
front: select text overflow

### LINKED ISSUE #626: select: placeholder and value text overflow
### What happened?

Image

The best approach would be to truncate the text with ‘…’ and add some padding.

Reduce the size of the select or use a very long text

_No response_

Firefox, Google Chrome

Wadjetz:
fix #626 
Capture d’écran 2024-10-09 à 16 29 57

### DIFF
--- a/ui-core/src/styles/select.css
+++ b/ui-core/src/styles/select.css
@@ -5,7 +5,10 @@
     0px 1px 1px rgba(0, 0, 0, 0.2),
     0px 2px 4px rgba(0, 0, 0, 0.1);
   height: 2.5rem;
-  padding-left: 0.813rem;
+  padding-inline: 0.813rem 1.875rem;
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
   letter-spacing: 0;
   text-align: left;
   line-height: 1.5rem;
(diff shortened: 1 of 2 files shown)


achrafmohye (approved):
LGTM and tested :)

thibautsailly:
Could you remove 8px to the label's width? It will give more room to the vertical arrows on their left side, and will make it easier for users to spot the select component at a glance when eye scanning a form. Thanks!

ui-core/src/styles/select.css
```diff
@@ -6,6 +6,10 @@
     0px 2px 4px rgba(0, 0, 0, 0.1);
   height: 2.5rem;
   padding-left: 0.813rem;
+  padding-right: 1.875rem;
```
SharglutDev: We could win one line of code

```suggestion
  padding-inline: 0.813rem 1.875rem;
```

SharglutDev (approved):
Lgtm and tested. Should we also add a `title` attribute @thibautsailly ?

thibautsailly:
> Lgtm and tested. Should we also add a `title` attribute @thibautsailly ?

The option label should be displayed entirely once you click the selector, which doesn't interfere with data input in the form. So I don't think it's necessary.

PR #628, opened 2024-10-09
ui-core: make combobox filter optional

### LINKED ISSUE #578: ComboBox: Clear button does not appear when a default value is pre-selected
### What happened?

The current issue is that the clear button does not appear when a default value is pre-selected that matches a suggestion. This behavior has been confirmed with @thibautsailly 

The clear button to be available when the value matches a suggestion, regardless of whether the value was user-selected or pre-populated.

Choose a story where the combo-box is pre-populated with a valid value, notice the clear button

Local Storybook, Deployed Storybook

Google Chrome

SarahBellaha:
This PR adds a new prop to the ComboBox component so that the default suggestion filter can be deactivated.
This modification is necessary in cases where suggestions need to be displayed in a specific non-alphabetical order.

Also closes #578

### DIFF
--- a/ui-core/src/components/inputs/ComboBox/ComboBox.tsx
+++ b/ui-core/src/components/inputs/ComboBox/ComboBox.tsx
@@ -33,7 +33,7 @@ const ComboBox = <T,>({
   customLabel,
   numberOfSuggestionsToShow = 5,
   exactSearch = false,
-  value,
+  value = '',
   small,
   onSelectSuggestion,
   disableDefaultFilter = false,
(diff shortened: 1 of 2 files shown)


ui-core/src/components/inputs/ComboBox/ComboBox.tsx
```diff
@@ -32,22 +33,28 @@ const ComboBox = <T,>({
   customLabel,
   numberOfSuggestionsToShow = 5,
   exactSearch = false,
-  value,
+  value = '',
```
clarani, line 36: Nice ! :+1:

clarani (approved):
LGTM :white_check_mark:

PR #560, opened 2024-09-26
ui-speedspacechart: speed limit tags enhancements

Caracol3:
- adding logic to make the tooltip follow the cursor move horizontally and vertically
- sticking the tooltip to the chart when the cursor goes outside
- adding logic to prevent text going outside its tag
Signed-off-by: Mathieu <mathieu.coulibaly@sncf.fr>

PR #633, opened 2024-10-10
ui-speedspacechart: fix speedlimittags layer blinking

### LINKED ISSUE #427: Speed space chart is blinking
### What happened?

Open all context information.
When we are moving the mouse over the speed space chart, it blinks. Same when we zoom in and out.

`Speed Limit Tag` is the most visible blinking component. 

We should not see the chart blinking.

1. Go [here](https://openrailassociation.github.io/osrd-ui/?path=/story/speedspacechart-rendering--speed-space-chart-default)
2. Open all context information.
3. Play around

_No response_

_No response_

Synar:
Close #427

There were 2 issues I could find leading to the blinking : 
- the store object was used as a dependency of useEffect, which means it was triggered every time
- the tooltip computation and the canvas drawing were tied together in a single function, despite the tooltip computation depending on the cursor position (and we don't want the canvas to be redrawn whenever the cursor moves)

There seems to be an overreliance on the store object in the code, but it may take a major refactor to fix it.

### DIFF
--- a/ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
+++ b/ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
@@ -1,10 +1,15 @@
-import type { DrawFunctionParams, tooltipInfos } from '../../../types/chartTypes';
+import type {
+  DrawFunctionParams,
+  SpeedLimitTagsLayerDrawFunctionParams,
+  tooltipInfos,
+} from '../../../types/chartTypes';
 import {
   COLOR_DICTIONARY,
   LINEAR_LAYER_SEPARATOR_HEIGHT,
   LINEAR_LAYERS_BACKGROUND_COLOR,
   LINEAR_LAYERS_HEIGHTS,
   MARGINS,
+  WHITE,
 } from '../../const';
 import {
   clearCanvas,
@@ -43,23 +48,22 @@ export const drawSpeedLimitTags = async ({
   width,
   height: marginTop,
   store,
-}: DrawFunctionParams): Promise<tooltipInfos | null> => {
+}: SpeedLimitTagsLayerDrawFunctionParams) => {
   const {
     speedLimitTags,
     ratioX,
     leftOffset,
     layersDisplay: { electricalProfiles, powerRestrictions },
-    cursor,
   } = store;
 
-  const { MARGIN_TOP, MARGIN_BOTTOM, MARGIN_LEFT, MARGIN_RIGHT } = MARGINS;
+  const { MARGIN_BOTTOM, MARGIN_LEFT, MARGIN_RIGHT } = MARGINS;
 
   clearCanvas(ctx, width, LINEAR_LAYERS_HEIGHTS.SPEED_LIMIT_TAGS_HEIGHT);
 
   ctx.save();
   ctx.translate(leftOffset, 0);
 
-  const maxPosition = maxPositionValue(store);
+  const maxPosition = maxPositionValue(store.speeds);
 
   const questionImage = await loadSvgImage(questionBlobUrl);
   const alertFillImage = await loadSvgImage(alertFillBlobUrl);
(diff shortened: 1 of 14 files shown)


Synar:
Close #427 on osrd-ui

Therefore fix half of https://github.com/OpenRailAssociation/osrd/issues/8853 on osrd

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
```diff
@@ -82,108 +85,135 @@ export const drawSpeedLimitTags = async ({
     LINEAR_LAYERS_HEIGHTS.SPEED_LIMIT_TAGS_HEIGHT - LINEAR_LAYER_SEPARATOR_HEIGHT
   );
 
-  let tooltip: tooltipInfos | null = null;
+  if (speedLimitTags == undefined) return;
```
Yohh:
```suggestion
  if (!speedLimitTags) return;
```

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
```diff
@@ -82,108 +85,135 @@ export const drawSpeedLimitTags = async ({
...
+    };
+
+    const { tag } = value;
+    if (!(tag in tooltipTextMap)) continue;
+    const potential_tooltip_text = tooltipTextMap[tag];
+
+    const x = positionOnGraphScale(position.start, maxPosition, width, ratioX, MARGINS);
+    const nextBoundary = positionOnGraphScale(position.end!, maxPosition, width, ratioX, MARGINS);
+
+    if (nextBoundary == null || cursor.x == null || cursor.y == null) break;
```
Yohh:
```suggestion
    if (nextBoundary === null || cursor.x === null || cursor.y === null) break;
```

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
```diff
@@ -82,108 +85,135 @@ export const drawSpeedLimitTags = async ({
...
+  const { speedLimitTags, ratioX, leftOffset, cursor } = store;
+
+  const { MARGIN_TOP, MARGIN_LEFT } = MARGINS;
+
+  const maxPosition = maxPositionValue(store.speeds);
+
+  let tooltip: tooltipInfos | null = null;
 
-    return tooltip;
+  if (speedLimitTags == undefined) return null;
```
Yohh:
```suggestion
  if (!speedLimitTags) return null;
```

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
```diff
@@ -82,108 +85,135 @@ export const drawSpeedLimitTags = async ({
...
+        );
+
+        drawSvgImageWithColor(
+          ctx,
+          image,
+          iconXPosition + ICON_OFFSET,
+          iconYPosition + ICON_OFFSET,
+          ICON_WIDTH,
+          ICON_HEIGHT,
+          '#FFFFFF'
```
Yohh: nit: you could import { WHITE } from `const` then use it with `hex()` as in `reticle.ts`

Yohh (commented):
very nice!

a few comments before reviewing deeper

Yohh:
I can see a little unwanted space at the end of some rects, depending on the zoom level:
image

Yohh:
I take the liberty of assigning @Caracol3 , he has a pending [PR](https://github.com/OpenRailAssociation/osrd-ui/pull/560) that will have many conflicts after this one

Synar:
> I can see a little unwanted space at the end of some rects, depending on the zoom level:
> image
> 

Is this behavior that wasn't present before the pr? I tried not to touch the drawing code and dimension computations at all, so if something changed there I messed up

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
Synar: Sure!

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
Synar: Since == null matches exactly null and undefined and nothing else, this is actually the only case where I prefer == to ===. So I would rather keep it like this unless you insist.

Yohh:
> > I can see a little unwanted space at the end of some rects, depending on the zoom level:
> 
> Is this behavior that wasn't present before the pr? I tried not to touch the drawing code and dimension computations at all, so if something changed there I messed up

you're right, it's not your work, that was already present

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
Synar: Will do!

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
```diff
@@ -18,12 +18,44 @@ const SpeedLimitTagsLayer = ({ width, marginTop, store }: SpeedLimitTagsLayerPro
   const tooltip = useRef<tooltipInfos | null>();
 
   useEffect(() => {
-    const updateTooltip = async () => {
+    const updateCanvas = async () => {
       const currentCanvas = canvas.current as HTMLCanvasElement;
       const ctx = currentCanvas.getContext('2d') as CanvasRenderingContext2D;
-      tooltip.current = await drawSpeedLimitTags({ ctx, width, height: marginTop, store });
+      const restrictedStore = {
```
emersion, line 24: Hm, what is the purpose of creating an object containing a subset of `store` here? It seems like `drawSpeedLimitTags()` just pulls in part of these props anyways.

Note, TypeScript is pretty liberal when it comes to type compatibility for objects. This compiles just fine:

```ts
const obj = { foo: 42, bar: 43 };
type Foo = { foo: number };
const asdf: Foo = obj;
```

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
```diff
@@ -18,12 +18,44 @@ const SpeedLimitTagsLayer = ({ width, marginTop, store }: SpeedLimitTagsLayerPro
...
+          electricalProfiles: store.layersDisplay.electricalProfiles,
+          powerRestrictions: store.layersDisplay.powerRestrictions,
+        },
+        speeds: store.speeds,
+      };
+      await drawSpeedLimitTags({
+        ctx,
+        width,
+        height: marginTop,
+        store: restrictedStore as RestrictedStore,
```
emersion: I prefer to avoid `as` when possible, because it basically turns off the type checks. I don't think it's necessary here, `const restrictedStore: RestrictedStore = …` should work fine I believe (or just leave it up to type inference, should work too).

ui-speedspacechart/src/types/chartTypes.ts
```diff
@@ -93,6 +93,20 @@ export type DrawFunctionParams = {
...
+> & {
+  layersDisplay: Pick<Store['layersDisplay'], 'electricalProfiles' | 'powerRestrictions'>;
+};
+
+export type RestrictedDrawFunctionParams = {
+  ctx: CanvasRenderingContext2D;
+  width: number;
+  height: number;
+  store: RestrictedStore;
+};
```
emersion: Hm, I wonder if "restricted" is a good way to describe this. Is this going to be generally enough for any drawing function to re-use, or is this really specific to drawing speed limit tags, or a specific kind of layer?

Also not a fan of repeating the types… Do we really gain that much by narrowing the type here?

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
Yohh: you're right, I just prefer to opt for the most restrictive when it's possible, it seems safer to me, but I don't want to look pyschorigid so as you wish

ui-speedspacechart/src/components/utils.ts
```diff
@@ -23,23 +23,20 @@ export const getGraphOffsets = (width: number, height: number, declivities?: boo
...
- * @param store
+ * Given a list of speed data, return the maxSpeed
+ * @param speeds
  */
-export const maxSpeedValue = (store: Store) => {
-  const speeds = store.speeds;
-  return Math.max(...speeds.map(({ value }) => value));
-};
-
+export const maxSpeedValue = (speeds: LayerData<number>[]) =>
```
emersion, line 29: So, what is the motivation for changing the type here?

Is it to make it more obvious that only the `speeds` field from the store is used, so that callers can adjust their dependencies accordingly? 

(Would be nice to record this kind of info in the commit message :P )

emersion:
> There were 2 issues I could find leading to the blinking

Next time, if you're feeling up for it, would be easier for reviewers if these two changes were split into separate commits!

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
emersion: We try to avoid `==` because it has quite weird rules. In general, I go for `if (!thing)` instead of `if (thing == null)` (but `if (thing === null)` is fine by me as well).

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
```diff
@@ -82,108 +85,135 @@ export const drawSpeedLimitTags = async ({
...
+    if (nextBoundary == null || cursor.x == null || cursor.y == null) break;
+
+    const tagWidth = nextBoundary - x - RECTANGLE_SPACING;
+    if (
+      cursor.x >= x - MARGIN_LEFT + leftOffset &&
+      cursor.x <= x - MARGIN_LEFT + leftOffset + tagWidth &&
+      cursor.y >= marginTop - MARGIN_TOP + Y_POSITION - 2 &&
+      cursor.y <= marginTop - MARGIN_TOP + Y_POSITION - 1 + RECT_HEIGHT
+    ) {
+      tooltip = {
```
emersion: I think we can straight up `return` here without the intermediate variable?

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
Yohh: another way
```suggestion
    if (!nextBoundary || !cursor.x || !cursor.y) break;
```

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
Synar: Good catch. I think the code behaved the exact same before I did this, but I got an eslint warning because useEffect only contained a subset of the store and to the linter it looked like I was using all of the store. I changed a few things around since I got this linting error, so I will check again.

ui-speedspacechart/src/types/chartTypes.ts
Synar: I agree with you, the name is a bit imprecise, if I keep it I'll rename it.

The reason I did this was to be able to restrict what I fed to the canvas drawing function. I don't particularly enjoy this workaround either, so maybe I should fully remove the dependence of the drawing function on the store and its associated types and give it each of its arguments independently.

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
Synar: No, !x would be different here, because these are numbers, so x == null test if the value is nullish (only null and undefined) but !x tests if it's falsy, which includes values like 0

The equivalent would be x === null || x === undefined

Yohh: ah yes of course, my bad

Synar: Tbf I did this to avoid caring about whether they could be null or undefined, especially since in the previous version I think the type of one of them appeared like it could be neither null nor undefined, despite being tested. So perhaps I should actually investigate it instead of opting for the lazy way '^^

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
Synar: Indeed we can, thanks

ui-speedspacechart/src/components/utils.ts
Synar: Yes! The drawing function depends on this function, which depended on a fully populated store. So to decouple the first from the store, the second also needed to be decoupled.

Also I don't see why a function that computes things on an array and only depends on it should have any argument beyond the array in the first place

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
Synar: Ok, will do!

Synar: Ok so I hadn't constrained my type correctly with Pick, that's fixed

ui-speedspacechart/src/types/chartTypes.ts
Synar: To keep some consistency I have renamed and kept the type for now, but I think another pass to reduce the use of the store in all drawing functions could be appreciated. But this should probably be another pr.

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
Synar: Yeah still needed if we want the lint to pass. We could also deactivate this warning, but I do not dare do something like that.

Synar:
> I take the liberty of assigning @Caracol3 , he has a pending [PR](https://github.com/OpenRailAssociation/osrd-ui/pull/560) that will have many conflicts after this one

Thanks! I'll confess I did not notice there was another PR on this component. But I don't mind being the one to do the conflict resolution

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
emersion: What is the linting error exactly?

Synar: calling await drawSpeedLimitTags({ctx, width, height: marginTop, store: store}) in the useEffect without an explicit store dependence (which we don't want, and neither do we want all of its attributes as dependences because that includes cursor) causes :

/home/alice/dev/osrd-ui/ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
  32:6  warning  React Hook useEffect has a missing dependency: 'store'. Either include it or remove the dependency array  react-hooks/exhaustive-deps

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
emersion: Yet another way, if you prefer:

```ts
typeof cursor.x === 'number'
```

ui-speedspacechart/src/components/utils.ts
emersion: Makes sense!

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
```diff
@@ -82,108 +86,132 @@ export const drawSpeedLimitTags = async ({
...
+  if (!speedLimitTags) return null;
+
+  for (const { position, value } of speedLimitTags) {
+    const tooltipTextMap: Record<string, string> = {
+      incompatible: 'Incompatible with the infrastructure',
+    };
+
+    const { tag } = value;
+    if (!(tag in tooltipTextMap)) continue;
+    const potential_tooltip_text = tooltipTextMap[tag];
```
emersion: Style nit: we're using camelCase. Probably could be simply named `text` since that's the name of the field below.

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
```diff
@@ -82,108 +86,132 @@ export const drawSpeedLimitTags = async ({
...
+  const { speedLimitTags, ratioX, leftOffset, cursor } = store;
+
+  const { MARGIN_TOP, MARGIN_LEFT } = MARGINS;
+
+  const maxPosition = maxPositionValue(store.speeds);
+
+  if (!speedLimitTags) return null;
+
+  for (const { position, value } of speedLimitTags) {
+    const tooltipTextMap: Record<string, string> = {
```
emersion, line 189: Sounds a bit weird to have an object with just a single entry, and I don't think an explicit type is required here, but I don't mind too much.

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
emersion: ACK. Could be simplified a tad with `pick()` from lodash, but I'm fine with the current logic.

ui-speedspacechart/src/types/chartTypes.ts
```diff
@@ -93,6 +93,20 @@ export type DrawFunctionParams = {
   setStore?: React.Dispatch<React.SetStateAction<Store>>;
 };
 
+export type SpeedLimitTagsLayerDrawingStore = Pick<
+  Store,
+  'speedLimitTags' | 'ratioX' | 'leftOffset' | 'speeds'
+> & {
+  layersDisplay: Pick<Store['layersDisplay'], 'electricalProfiles' | 'powerRestrictions'>;
```
emersion, line 100: Question: do we really need to optimize the case where the user updates the visibility of a layer other than these two?

Just wondering if it's really worth the more involved type and picking.

ui-speedspacechart/src/types/chartTypes.ts
emersion: The new name looks good to me!

emersion (commented):
I'm mostly nitpicking at this point. Feel free to discard/ignore comments you disagree with.

ui-speedspacechart/src/types/chartTypes.ts
Synar: Honestly not really, but I still did this in order to avoid having non primary types in the useEffect dependency array

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
Synar: I'll confess I didn't look at the tooltip code at all, but I'll do

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
Synar: Sure, same as below

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
Synar: I wish.
Doing something like
```js
const restrictedStore = {
        ...pick(store, ['speedLimitTags', 'ratioX', 'leftOffset', 'speeds']),
        layersDisplay: pick(store.layersDisplay, ['electricalProfiles', 'powerRestrictions']),
        };
```
still leads to the same linting issue, as the linter does not particularly care to know what pick does, it only sees that store is used somewhere in the function. I'm sure there is some way to sidestep this issue, perhaps with useMemo, but it's not immediatly obvious to me

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
Synar: Ah no I thought you were talking about something else but I remember now

Due to how the code was structured I assumed there might be tooltips added for other tags in the future, and I wanted for their addition to be as smooth as possible, without nested if conditions like were present in the previous version of the code.

And some type indication is needed for potentialTooltipText to later know it is a string (even though I agree the ts engine should have been able to deduce that by itself)

ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
Synar: Fixed ^^

emersion (approved):

PR #634, opened 2024-10-11
front: fix combo box input props change

### LINKED ISSUE #9302: front: the stdcm form is not completly reset when clicking on "Effectuer une nouvelle requête"
### What happened?

When I reset the stdcm form, the form is reset in the store, but the inputs are not emptied. 

The fields should be emptied, otherwise, I think that they are still filled.

1. Go to stdcm, launch a request
2. Select the simulation
3. Click on reset the form
4. The form is still filled but you can see that the rollingStock image has disappeared

Recette (SNCF)

Firefox

.

Wadjetz:
fix [#9302](https://github.com/OpenRailAssociation/osrd/issues/9302)

### DIFF
--- a/ui-core/src/components/inputs/ComboBox/ComboBox.tsx
+++ b/ui-core/src/components/inputs/ComboBox/ComboBox.tsx
@@ -106,9 +106,14 @@ const ComboBox = <T,>({
       : []), // Conditionally include the chevron icon only when suggestions are not empty
   ];
 
+  /* eslint-disable react-hooks/exhaustive-deps */
   useEffect(() => {
     if (value) {
       setInputValue(value);
+    } else {
+      setInputValue('');
+      setSelectedOption(null);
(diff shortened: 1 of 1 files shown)


ui-core/src/components/inputs/ComboBox/ComboBox.tsx
```diff
@@ -80,15 +80,26 @@ const ComboBox = <T,>({
...
+    clearInputValue();
     focusInput();
   };
 
+  /* eslint-disable react-hooks/exhaustive-deps */
+  useEffect(() => {
+    if (!value) {
+      clearInputValue();
+    }
+  }, [value]);
```
clarani: Can you modify the existing hook line 120 instead of adding a new one ?
Also, why do you call clearInputValue ? setValue(value) is not enough ?:thinking:

clarani (approved):
LGTM, tested in osrd :white_check_mark:

PR #635, opened 2024-10-14
ui manchette/add waypoint click interactivity

### LINKED ISSUE #399: Manchette : expose a click event on each operational point in the manchette
### Description and goal

To be able to manipulate a specific operational point in the manchette, we should be able to click on it and pass any information

### Acceptance Criteria

- [ ] Expose a click event when clicking on an operational point 
- [ ] We should be able to pass any kind of action with the `onClick`

SharglutDev:
close #399 

See individual commits

Mockup: https://www.sketch.com/s/a26d84be-5fb2-41e4-9e84-96d90f46a997/a/OeJV9yG#Inspect


tailwind-preset.js
```diff
@@ -6,6 +6,7 @@ export default {
         black: {
           1: '#00000003',
           2: '#00000005',
+          5: 'rgb(0, 0, 0, 0.05)',
```
jacomyal: I have two issues with this line:

- `rgb` should be 3 channels, so here with the alpha, it should be `rgba`
- Mixing `rgba` notation with hex with alpha notations might be misleading, because the `05` at the end of `#00000005` means `5 / 256`, while the `0.05` at the end of `rgba(0, 0, 0, 0.05)` means `5 / 100` - I think mixing them makes them hard to compare


I'd suggest using `#0000000C`, as it seems to have to smallest impact.

ui-manchette/src/components/OperationalPoint.tsx
```diff
@@ -1,14 +1,35 @@
...
+import cx from 'classnames';
 
 import { type StyledOperationalPointType } from '../types';
 import '@osrd-project/ui-core/dist/theme.css';
 import { positionMmToKm } from '../utils';
 
-const OperationalPoint = ({ extensions, id, position, display }: StyledOperationalPointType) => {
+type OperationalPointProps = {
+  operationalPoint: StyledOperationalPointType;
+  activeOperationalPointId?: string;
```
jacomyal: I personnally prefer to use `isActive` kind of props. I think this component has no reason to know the active OP id, except if its its own. Instead, using `isActive` moves the "decision" of which OP is active to the parent, which makes more sense to me.

But it's a very minor suggestion, the code remains very understandable and maintainable like this.

ui-manchette/src/stories/Manchette.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+    setMenuPosition({ top: position.bottom - 2, left: position.left });
+    setActiveOperationalPointId(id);
+  };
+
+  return (
+    <Manchette
+      operationalPoints={OperationalPointListData.map((op) => ({
+        ...op,
+        onClick: (id, ref) => {
+          handleWaypointClick(id, ref);
```
jacomyal, line 72: As a user, it felt weird not to be able to close the menu, unless by clicking an actual action. This could be implemented directly in the application side, but it would be nice to have it directly in this example.

One simple way to handle it, is to give the `Menu` component a `onClickElsewhere` prop, that would allow you to catch when the users clicks anywhere else:

```tsx
type MenuProps = {
  menuRef: React.RefObject<HTMLDivElement>;
  items: MenuItem[];
  style?: React.CSSProperties;
  onClickElsewhere?: () => void;
};

const Menu = ({ menuRef, items, style, onClickElsewhere }: MenuProps) => {
  useEffect(() => {
    const handleClickBody = (e: MouseEvent) => {
      if (!menuRef.current || !onClickElsewhere) return;

      const node = e.target as Node;
      if (!menuRef.current.contains(node)) {
        onClickElsewhere();
      }
    };

    setTimeout(() => {
      document.body.addEventListener('click', handleClickBody);
    });
    return () => {
      document.body.removeEventListener('click', handleClickBody);
    };
  }, [menuRef, onClickElsewhere]);

  return (
    <div ref={menuRef} className="menu" style={style}>
      {items.map(({ title, icon, onClick }) => (
        <button key={title} type="button" className="menu-item" onClick={onClick}>
          <span className="icon">{icon}</span>
          <span>{title}</span>
        </button>
      ))}
    </div>
  );
};
```

Then you can close the menu when the user clicks somewhere else like this:

```tsx
<Menu
  menuRef={menuRef}
  items={menuItems}
  style={{ width: '305px', top: menuPosition.top, left: menuPosition.left }}
  onClickElsewhere={() => {
    setMenuPosition(null);
  }}
/>
```

ui-manchette/src/styles/operational-point.css
Synar, line 1: Is the background of each waypoint supposed to darken when hovered? I don't see any hover behavior when testing the story

Synar (changes requested):
Very nice! Tested and approved.

ui-manchette/src/stories/Manchette.stories.tsx
SharglutDev: You're totally right, we already have this feature on osrd so to keep it simple in the story I chose to have a very raw menu. My intention with the story was just to demonstrate that a user could pass this kind of menu to handle some actions by clicking on a waypoint.
If you think it's very frustrating, I can indeed add what you suggested.

ui-manchette/src/components/OperationalPoint.tsx
SharglutDev: Thank you for this suggestion, I agree with you, made the change

tailwind-preset.js
SharglutDev: Good point, we can keep these values harmonized. Copilot tells me `#0000000D` is closer to 5% opacity (even though our eyes can't probably see the difference).

Synar (approved):

ui-manchette/src/stories/Manchette.stories.tsx
jacomyal: You're right, it's probably not necessary, the code is indeed simpler that way. As long as it's done "on the other side", it's all good to me :+1:

Math-R (approved):
lgtm, nice work

PR #646, opened 2024-10-15
ui-spacetimechart: use local timezone to format time

emersion:
First, the logic was inconsistent: UTC was used for hours but the local timezone was used for minutes. This works by chance because most timezones are 1h apart, however that's not true in general (e.g. Sri Lanka has +05:30 and Australia has +09:30).

Second, we never want to display UTC to the user. Users are always interested in times given in their timezone.


jacomyal (approved):

clarani (approved):
LGTM :white_check_mark:

PR #647, opened 2024-10-15
ui-spacetimechart: use paths layer for conflicts

emersion:
The overlay layer is rendered on top of the captions, which is undesirable here.

Example bug:

out

### DIFF
--- a/ui-spacetimechart/src/components/ConflictLayer.tsx
+++ b/ui-spacetimechart/src/components/ConflictLayer.tsx
@@ -45,7 +45,7 @@ export const ConflictLayer = ({ conflicts }: ConflictLayerProps) => {
     [conflicts]
   );
 
-  useDraw('overlay', drawConflictLayer);
+  useDraw('paths', drawConflictLayer);
 
   return null;
 };


Akctarus (approved):
LGTM

SharglutDev (commented):
Doesn't seem like it fixes the issue ? Or did I do something wrong ?

Capture d’écran 2024-10-17 à 09 57 56

clarani:
I still have the issue too :/

emersion:
Apologies, I messed up when testing. I've updated the PR. Here's a hack patch to test this in the storybook:

```diff
diff --git a/ui-spacetimechart/src/stories/layers.stories.tsx b/ui-spacetimechart/src/stories/layers.stories.tsx
index 028ee5b5f118..2c6479c0f5d3 100644
--- a/ui-spacetimechart/src/stories/layers.stories.tsx
+++ b/ui-spacetimechart/src/stories/layers.stories.tsx
@@ -32,7 +32,7 @@ const CONFLICTS = [
     timeStart: +START_DATE + 35 * MINUTE,
     timeEnd: +START_DATE + 37 * MINUTE,
     spaceStart: 39 * KILOMETER,
-    spaceEnd: 41 * KILOMETER,
+    spaceEnd: 99999 * KILOMETER,
   },
 ];
 
```

SharglutDev (approved):
Lgtm and tested. Should we add another commit to do the same for occupation blocks ?

emersion:
That one uses the `background` layer, because it's drawn behind the paths (so it doesn't have the same bug).

PR #649, opened 2024-10-17
ui-manchette: add fit button

### LINKED ISSUE #410: manchette: add fit button
closes https://github.com/osrd-project/osrd-confidential/issues/636

## Description

The header represents the spatial axis of the Space-Time Diagram (GET). From the header, there should be a FIT button that allows switching to a view where the entire path is displayed.
On the first render, the selected train must be optimally displayed horizontally.

## Acceptance Criteria

- [ ] The user has a button to reset the zoom level in order to see a train in its entirety.
- [ ] On the first render, the user can see the selected train in full.
- [ ] On the first render, the horizontal scale must be adjusted to the trains in the timetable.

### osrd-ui (size 0.75)
- In the header, add a button to dispatch an event that resets the scale.
- In the Header+GET component, add logic to:
  - Adjust the maximum allowed zoom level on the GET.
  - Modify the vertical scales (space) when clicking the button.
  - Modify the horizontal scale (time) for the first render.

### osrd (size 0.25)

- Integrate the new version of the header, providing the new props if necessary.

clarani:
closes #410 

image

### DIFF
--- a/ui-manchette/src/components/Manchette.tsx
+++ b/ui-manchette/src/components/Manchette.tsx
@@ -12,6 +12,7 @@ type ManchetteProps = {
   activeOperationalPointId?: string;
   zoomYIn: () => void;
   zoomYOut: () => void;
+  resetZoom: () => void;
   height?: number;
   yZoom?: number;
   children?: React.ReactNode;
@@ -22,6 +23,7 @@ type ManchetteProps = {
 const Manchette = ({
   zoomYIn,
   zoomYOut,
+  resetZoom,
   yZoom = 1,
   operationalPoints,
   activeOperationalPointId,
@@ -31,7 +33,7 @@ const Manchette = ({
 }: ManchetteProps) => (
   <div className="manchette-container">
     <div
-      className=" bg-ambientB-10 border-r border-grey-30"
+      className="bg-ambientB-10 border-r border-grey-30"
       style={{ minHeight: `${INITIAL_OP_LIST_HEIGHT}px` }}
     >
       <OperationalPointList
@@ -40,23 +42,16 @@ const Manchette = ({
       />
       {children}
     </div>
-    <div className="manchette-actions flex items-center">
-      <div className=" flex items-center ">
-        <button
-          className="h-full px-3 w-full zoom-out"
-          onClick={zoomYOut}
-          disabled={yZoom <= MIN_ZOOM_Y}
-        >
+    <div className="manchette-actions">
+      <div className="zoom-buttons">
+        <button className="zoom-out" onClick={zoomYOut} disabled={yZoom <= MIN_ZOOM_Y}>
           <ZoomOut />
         </button>
-      </div>
-      <div className=" flex items-center  border-x border-black-25  h-full">
-        <button
-          className="h-full px-3 w-full zoom-in"
-          disabled={yZoom >= MAX_ZOOM_Y}
-          onClick={zoomYIn}
-        >
+        <button className="zoom-in" onClick={zoomYIn} disabled={yZoom >= MAX_ZOOM_Y}>
           <ZoomIn />
+        </button>
+        <button className="zoom-reset" onClick={resetZoom}>
+          Fit
         </button>
       </div>
       <div className="flex items-center ml-auto text-sans font-semibold">
--- a/ui-manchette/src/stories/Manchette.stories.tsx
+++ b/ui-manchette/src/stories/Manchette.stories.tsx
@@ -74,6 +74,7 @@ const ManchetteWithWaypointMenu = () => {
       }))}
       zoomYIn={() => {}}
       zoomYOut={() => {}}
+      resetZoom={() => {}}
       toggleMode={() => {}}
       activeOperationalPointId={activeOperationalPointId}
     >
--- a/ui-manchette/src/styles/manchette.css
+++ b/ui-manchette/src/styles/manchette.css
@@ -1,6 +1,12 @@
 .manchette-container {
   .manchette-actions {
+    align-items: center;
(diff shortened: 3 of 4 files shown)


ui-manchette/src/components/Manchette.tsx
```diff
@@ -12,6 +12,7 @@ type ManchetteProps = {
   activeOperationalPointId?: string;
   zoomYIn: () => void;
   zoomYOut: () => void;
+  resetZoom: () => void;
```
emersion, line 15: Because this is not optional, and this prop is not supplied until the next commit, the build breaks when checking out this commit. Suggestion: (1) merge the two commits or (2) add the button without the prop in this commit, add the prop in the next commit instead.

ui-manchette/src/styles/manchette.css
```diff
@@ -9,20 +15,34 @@
...
+
+      .zoom-reset {
+        color: #797671;
+        font-weight: 600;
+        font-size: 14px;
+      }
     }
 
-    .toggle-mode {
+    .toggle-mode{
```
emersion: Style: missing space before `{`

ui-manchette/src/styles/manchette.css
```diff
@@ -9,20 +15,34 @@
...
+
+      button {
+        border-right: solid 1px #d3d1cf;
+        border-top: solid 1px white;
+        height: 100%;
+        padding-inline: 12px;
+      }
+
+      button:disabled {
+        span {
```
emersion: I'd prefer not to use `span` here, because this adds assumptions about the internal HTML structure of icon components. I believe the CSS rule should work fine with that selector dropped?

ui-manchette/src/components/Manchette.tsx
```diff
@@ -40,24 +42,17 @@ const Manchette = ({
...
-        <button
-          className="h-full px-3 w-full zoom-in"
-          disabled={yZoom >= MAX_ZOOM_Y}
-          onClick={zoomYIn}
-        >
+        <button className="zoom-in" onClick={zoomYIn} disabled={yZoom >= MAX_ZOOM_Y}>
           <ZoomIn />
         </button>
+        <button className="zoom-reset" onClick={resetZoom}>
+          Fit
```
emersion, line 54: Seems kind of weird to have this untranslated string next to icon buttons. But it seems like it's what the mockup wants so shrug.

ui-manchette/src/styles/manchette.css
```diff
@@ -9,20 +15,34 @@
...
+          color: #B6B2AF
+        }
+      }
+
+      .zoom-reset {
+        color: #797671;
```
emersion: Could we use the design system's colors instead of hardcoded hex codes? Something like `@apply text-grey-50`? (Ditto for other colors)

ui-manchette/src/styles/manchette.css
```diff
@@ -1,6 +1,12 @@
 .manchette-container {
   .manchette-actions {
+    align-items: center;
     background-color: rgba(250, 249, 245, 0.6);
+    border-top: solid 1px;
+    border-right: solid 1px;
+    border-radius: 0px 0px 0px 10px;
```
emersion: Nit: no need to specify the unit for zero lengths.

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
```diff
@@ -69,10 +69,12 @@ const ManchetteWithSpaceTimeWrapper = ({
   );
 };
 
-export default {
+const meta = {
```
emersion: Seems like this unrelated change slipped into this commit?

ui-manchette/src/styles/manchette.css
```diff
@@ -9,17 +15,29 @@
...
 
-    .zoom-in:disabled {
-      opacity: 0.4;
+    .zoom-buttons {
+      display: flex;
+      align-items: center;
+      height: 100%;
+
+      button {
+        border-right: solid 1px #d3d1cf;
```
emersion: Could we use `theme()` to avoid hardcoding a color here?

https://tailwindcss.com/docs/functions-and-directives#theme

ui-manchette/src/stories/Manchette.stories.tsx
```diff
@@ -23,78 +19,14 @@ const meta: Meta<typeof Manchette> = {
...
-    zoomYIn: {
-      action: 'zoomYIn',
-    },
-    zoomYOut: {
-      action: 'zoomYOut',
-    },
   },
 };
 
-const ManchetteWithWaypointMenu = () => {
```
SharglutDev: You don't want to keep this story ? :face_holding_back_tears:

clarani: It is not used for the moment, so currently it's dead code :)
I tried to see what how it works, but it was a bit broken, so I decided to remove it. If you prefer, we can try to fix it, or we can remove it in this commit and put it again later, once we have a more precise use case ?

ui-manchette/src/styles/manchette.css
clarani: Sure

ui-manchette/src/stories/Manchette.stories.tsx
clarani: Sorry, I misunderstood ! The story is back :white_check_mark:

emersion:
Seems like the first commit "ui-manchette: clean manchette's story" is just shuffling stuff around now? Maybe it should be dropped?

clarani:
You're right, it's done :)

emersion (approved):
LGTM, thanks!

PR #650, opened 2024-10-18
core: fix combo box clear icon

SarahBellaha:
PR to address the issue of the clear icon being displayed even if the input is empty when the ComboBox is initialized with a default value.

### DIFF
--- a/ui-core/src/components/inputs/ComboBox/ComboBox.tsx
+++ b/ui-core/src/components/inputs/ComboBox/ComboBox.tsx
@@ -80,15 +80,20 @@ const ComboBox = <T,>({
 
     // Call the parent's onChange handler with the synthetic event
     onChange?.(syntheticEvent);
+  };
+
+  const clearAndFocusInput = () => {
+    clearInput();
     focusInput();
   };
 
   const icons = [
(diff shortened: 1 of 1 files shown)


ui-core/src/components/inputs/ComboBox/ComboBox.tsx
```diff
@@ -111,9 +116,7 @@ const ComboBox = <T,>({
     if (value) {
       setInputValue(value);
     } else {
-      setInputValue('');
-      setSelectedOption(null);
-      onSelectSuggestion?.(undefined);
+      clearInput();
```
Wadjetz, line 119: The clearInput function calls onSelectSuggestion, but I think we just want to reset the internal state of the component. This also causes a bug I detected when using it in SDCM: it’s called initially and, in my case, resets the value (after reloading the page) that was set previously.

Maybe we can create a new function that just resets the internal state without calling onSelectSuggestion?

Wadjetz (changes requested):

emersion:
Superseded by https://github.com/OpenRailAssociation/osrd-ui/pull/779

PR #652, opened 2024-10-18
ui-manchette-with-spacetimechart: fix op display

### LINKED ISSUE #654: Wrong op height in manchette
### What happened?

Currently, the op height used in the space time chart is 32px accordingly to the mockup

Image

But in the manchette, operational points don't have any height resulting the waypoint menu to be to close to the op name comparing to the mockup.

Fix the op height, its line-height and padding in manchette

Compare op height when hovering it in manchette with the mockup

Deployed Storybook

Firefox

### LINKED ISSUE #8930: GET and manchette horizontal lines for operational points are not in sync
### What happened?

In many cases, Km or Linear, the lines from the machette aren't aligned with the horizontal blue lines of the GET:
image

Same here with a lot of zoom:
image

And in a case of a very unzoomed view, there are too many horizontal lines in GET:
image

I'm expecting the lines to be in sync.

1. Manipulate the GET

Dev (SNCF), Recette (SNCF)

Google Chrome

68e0513

Yohh:
- set only displayable operational points in calcOperationalPointsToDisplay in helpers.ts
- add padding-top to operational-point-list

Capture d’écran 2024-10-21 à 14 20 07

close https://github.com/OpenRailAssociation/osrd/issues/8930
close #654

### DIFF
--- a/ui-manchette/src/styles/operational-point-list.css
+++ b/ui-manchette/src/styles/operational-point-list.css
@@ -4,4 +4,12 @@
   padding-inline: 0.5rem 1.625rem;
 
   @apply bg-ambientB-10;
+
+  .operational-point-wrapper:last-child {
+    .op {
+      &::after {
+        bottom: 16px;
+      }
+    }
+  }
 }
--- a/ui-manchette/src/styles/operational-point.css
+++ b/ui-manchette/src/styles/operational-point.css
@@ -1,5 +1,8 @@
 .op {
   @apply text-grey-80;
+  height: 2rem;
+  line-height: 1.5rem;
+  padding-block: 3px 5px;
   width: 315px;
   position: relative;
   &-separator {
@@ -13,7 +16,7 @@
     &::after {
       content: '';
       position: absolute;
-      bottom: 10.75px;
+      bottom: 11px;
       left: 0;
       width: 100%;
       height: 0.5px;
@@ -82,7 +85,7 @@
     content: '';
     position: absolute;
     display: inline-block;
-    bottom: 10.75px;
+    bottom: 16px;
     right: -1.625rem;
     width: 1.625rem;
     height: 0.5px;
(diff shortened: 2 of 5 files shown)


ui-manchette/src/styles/operational-point.css
SharglutDev, line 1: Looks like that the horizontal line doesn't go until the edge of the manchette 

image

But on the mockup most of them are (seems to depends on another feature that we don't have yet) : 

image

ui-manchette/src/styles/operational-point.css
```diff
@@ -1,5 +1,8 @@
 .op {
   @apply text-grey-80;
+  height: 2rem;
+  line-height: 1.5rem;
+  padding-block: 0 0.3125rem;
```
SharglutDev: To match the mockup vertical alignment

```suggestion
  padding-block: 3px 0.3125rem;
``

File: ui-manchette/src/styles/operational-point-list.css [hunk 2]
```diff
@@ -1,7 +1,7 @@
 .operational-point-list {
   width: 349px;
   border-bottom-left-radius: 0.25rem;
-  padding-inline: 0.5rem 1.625rem;
+  padding-inline: 0.125rem 1.625rem;
```
SharglutDev: Is there any reason why you changed this value ?

The mockup says 8px 
image

ui-manchette/src/styles/operational-point.css
```diff
@@ -13,7 +16,7 @@
     &::after {
       content: '';
       position: absolute;
-      bottom: 10.75px;
+      bottom: 13px;
```
SharglutDev: In the mockup, the horizontal line is aligned with the middle of the letters 
image

Right now, it seems we miss a few pixels
image

SharglutDev (commented):
Thank you for the PR. Left some comments.

Otherwise, the fix for the alignment seems to be working.

ui-manchette/src/styles/operational-point.css
Yohh: my bad

ui-manchette/src/styles/operational-point.css
Yohh: seems too high, but fine with 1.5px

ui-manchette/src/styles/operational-point-list.css
Yohh: no that's a mistake

ui-manchette/src/styles/operational-point.css
Yohh: fixed with the padding-block

SharglutDev (approved):
Lgtm and tested, thank you for the fix !

PR #653, opened 2024-10-18
ui-spacetimechart: handles better empty data

### LINKED ISSUE #648: Space time chart crashing when incomplete datas
### What happened?

In some situations, the space time chart crashes when there isn't enough computed scales or when there are trains with no curves.

For now, we did some small fixes on osrd side (see TODOs in `SimulationResult.tsx` and `ManchetteWithSpaceTimeChartWrapper.tsx` which will be added in this [PR](https://github.com/OpenRailAssociation/osrd/pull/8796 ) on this [commit](https://github.com/OpenRailAssociation/osrd/pull/8796/commits/31eca1810fac71bc10844ad74e3ebf8901a66c61)) but the should rather be on osrd-ui.

Handle more edge case in ui-spacetimechart to be able to remove those checks and formating in osrd

Check the specified files

Deployed Storybook

Firefox

jacomyal:
This commit fixes #648.

Details:
- Adds a new "Use empty data" in the SpaceTimeChart/Rendering story, to use no OP, no scales and only a path with no point, as data
- Removes the "Invalid scales" error, and replaces it by generating a dummy placeholder NormalizedScaleTree
- In PathLayer.tsx, handles the only case where it's actually mandatory to have some point, with an early exit

### DIFF
--- a/ui-spacetimechart/src/components/PathLayer.tsx
+++ b/ui-spacetimechart/src/components/PathLayer.tsx
@@ -271,6 +271,8 @@ export const PathLayer = ({
    */
   const drawExtremities = useCallback<DrawingFunction>(
     (ctx, { getTimePixel, getSpacePixel, swapAxis }) => {
+      if (!path.points.length) return;
+
       const pathDirection = getDirection(path);
       const from = path.points[0];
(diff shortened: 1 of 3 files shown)


SharglutDev (approved):
Lgtm and tested, thank you !

PR #664, opened 2024-10-22
dependency: bump eslint-plugin-import to 2.31.0

emersion:
This requires manual changes because '..' is now sorted before other imports.


SharglutDev (approved):
Thank you for taking care of that :)

PR #666, opened 2024-10-24
ui-speedspacechart: update tooltip correctly

### LINKED ISSUE #9466: GEV is only updating tooltip every other time
### What happened?

tooltip is not updated and displays 25000V on a dead zone:
https://github.com/user-attachments/assets/11e577c4-31a9-428e-a9cf-e381e0025972

-

-

Local

Google Chrome

-

anisometropie:
closes [#9466](https://github.com/OpenRailAssociation/osrd/issues/9466)


Yohh (approved):
thanks!

LGTM :muscle:

emersion (approved):
Good catch! Refs must never be used as dependencies.

PR #651, opened 2024-10-18
ui-manchette: refactor manchette props

clarani:
closes #645

Issue #645, opened 2024-10-15
ui-manchette: redesign props


emersion:
### Description and goal

Current props are very OSRD-specific: https://github.com/OpenRailAssociation/osrd-ui/blob/dev/ui-manchette/src/types.ts

Proposal:

```ts
type Waypoint = {
  id: string
  name: string
  secondaryCode: string
  position: number // en mm
}

type TrainPath = {
  id: number
  name: string
  points: { time: number; position: number }[][] // à re-discuter avec Alexis et Mathieu
}

type SpaceTimeChartWithManchetteProps = {
  paths: TrainPath[]
  selectedTrainId: number
  waypoints: Waypoint[]
}

type ManchetteProps = {
  waypoints: Waypoint[];
}
```

### Acceptance criteria

.

PR #668, opened 2024-10-24
ui-manchette: fix waypoint menu positioning

### LINKED ISSUE #667: Waypoint menu position not dynamic
### What happened?

When the waypoint menu is open, if the manchette can be scrolled (long list of waypoints), the menu stays at the same origin position, it's not dynamic.

- [ ] The manchette becomes not scrollable when the menu is displayed
- [ ] We scroll in the page if we try to scroll when the menu is displayed, not in the manchette
- [ ] Zoom and mode buttons are disabled when menu is displayed

1. In sampleData.ts, add `pathProperties` data from a long path so the manchette can be scrolled
2. Click on a waypoint to open its menu
3. Scroll and see the menu not moving

Deployed Storybook

Firefox

SharglutDev:
- Move the menu next to its related waypoint so it can always be positioned relatively to it
- To keep osrd-ui as neutral as possible, the menu has no styles, they need to be passed by props

fix #667 

After discussion with @thibautsailly, it has been decided to totally block the scroll in the manchette (and only in the manchette) as long as the menu is displayed. If we still try to scroll, the whole page will scroll.


ui-manchette/src/components/OperationalPoint.tsx
```diff
@@ -27,7 +32,7 @@ const OperationalPoint = ({
       id={id}
       ref={opRef}
       onClick={() => {
-        if (onClick) onClick(id, opRef.current);
+        if (onClick) onClick(`${id}-${position}`); // to handle waypoints with same id
```
emersion: Hm, passing in here a custom string which isn't an ID seems pretty surprising… Especially since the argument is named `opId` in the type definition.

But why do we need to pass any argument here? The callers already provide a callback per-OP, no?

ui-manchette/src/types.ts
```diff
@@ -51,7 +51,19 @@ export type OperationalPointType = ArrayElement<PathProperties['operational_poin
...
   styles?: CSSProperties;
   display?: boolean;
-  onClick?: (opId: string, opRef: HTMLDivElement | null) => void;
+  onClick?: (opId: string) => void;
+};
+
+export type WaypointMenuData = {
+  activeOperationalPointId?: string;
+  waypointMenuItems?: WaypointMenuItem[];
+  waypointMenuClassName?: string;
```
emersion: Do we really need to allow callers to customize the class name? Wouldn't the menu already have the correct size, since it's now styled in osrd-ui? If callers really want to customize something, can't they use a regular CSS selector?

ui-manchette/src/types.ts
```diff
@@ -51,7 +51,19 @@ export type OperationalPointType = ArrayElement<PathProperties['operational_poin
 export type StyledOperationalPointType = OperationalPointType & {
   styles?: CSSProperties;
   display?: boolean;
-  onClick?: (opId: string, opRef: HTMLDivElement | null) => void;
+  onClick?: (opId: string) => void;
+};
+
+export type WaypointMenuData = {
+  activeOperationalPointId?: string;
```
emersion: Same comment as https://github.com/OpenRailAssociation/osrd-ui/pull/668/files#r1816147595

Could we maybe use an index instead of a custom string to refer to a particular OP in the list?

ui-manchette/src/styles/manchette.css
```diff
@@ -1,5 +1,44 @@
...
+      &:not(:first-child) {
+        border-top: 0.0625rem solid #ffffff;
+      }
+
+      &:not(:only-of-type, :last-child) {
+        border-bottom: 0.0625rem solid #d3d1cf;
+      }
+
+      /* .icon { */
+      span:first-child {
```
emersion: Should we use the class name here instead? Any reason why `:first-child` is used?

ui-manchette/src/components/WaypointMenu.tsx
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * The menu doesn't have any styles by default so the user can customize it as needed
+ *
+ * Styles need to be send from the application with props
+ */
+const WaypointMenu = ({ items, className }: WaypointMenuProps) => (
+  <div className={className}>
+    {items.map(({ title, icon, onClick }) => (
+      <button
+        key={title}
```
emersion: Human-readable strings aren't supposed to be used as keys in general, because they're not guaranteed to be unique.

ui-manchette/src/types.ts
SharglutDev: My understanding was that we want to keep the manchette/spacetime as neutral as possible. That's why I went on this path. If the team using this package wants to use the waypoint menu it can use its custom style for it to match the rest of their app design.
The `.menu` class has been added here only for the related story.
Furthermore, if the team using the menu doesn't want to keep any of the current style, they will need to invalidate all of them even if it's a css property that they may not have used. It seems pretty painful to me.

Should we have another discussion for that ?

ui-manchette/src/styles/manchette.css
SharglutDev: Because I went to the full customize option for the waypoint menu (see other comment). This would mean the user needs to pass more props for each `className` of the menu, it seemed pretty complicated to me.

ui-manchette/src/components/WaypointMenu.tsx
SharglutDev: Good point. Used `index` as the menu items won't change.

ui-manchette/src/components/OperationalPoint.tsx
SharglutDev: Hmm but if we don't pass any arguments, how can we have the information of which waypoint has been clicked ?

ui-manchette/src/types.ts
SharglutDev: Was afraid that when a waypoint is hidden when zooming out or when we hide one when clicking on it, it will break something but you're right it should be ok.

ui-manchette/src/components/OperationalPoint.tsx
emersion: Ah, apologies, I mixed things up with the menu item callback!

ui-manchette/src/components/WaypointMenu.tsx
```diff
@@ -0,0 +1,33 @@
...
+ */
+const WaypointMenu = ({ items, className }: WaypointMenuProps) => (
+  <div className={className}>
+    {items.map(({ title, icon, onClick }, index) => (
+      <button
+        key={index}
+        type="button"
+        onClick={(e) => {
+          onClick(e);
+        }}
```
emersion: Nit: can be simplified to `onClick={onClick}`

ui-manchette/src/styles/manchette.css
emersion: Hm, that sounds like a somewhat awkward middle ground to me: on one hand the structure of the menu is entirely managed by osrd-ui, on the other hand all of the styling is performed outside.

What's the use-case for keeping the styling outside? The menu appearance needs to be consistent with the manchette's styling, so I'm not sure I understand the motivation here?

If we really want to keep the styling out of osrd-ui, why do we need to have the menu and menu item components inside osrd-ui? We could pass the menu component from the outside as a prop.

ui-manchette/src/types.ts
emersion: Right, I'm not really getting why the manchette's styling would be in osrd-ui but the manchette menu's styling could not be. I must admit, I don't see why someone would want to change the menu's appearance, because it would then become inconsistent with the manchette.

ui-manchette/src/types.ts
emersion: Ah right, that's a good thing to keep in mind indeed! Maybe at some point we should introduce a unique waypoint ID (which would be different if the same OP appears twice), similar to OSRD's `PathItemLocation.id`?

But TBH, why does osrd-ui even need to access the OP ID? Couldn't we always pass a guaranteed unique waypoint ID in there? I suppose all of these questions are related to https://github.com/OpenRailAssociation/osrd-ui/issues/645.

ui-manchette/src/components/WaypointMenu.tsx
SharglutDev: Oh I did it again 🫣

ui-manchette/src/styles/manchette.css
SharglutDev: I honestly don't have any passion to keep this at it is now. I would even agree that it would be nice for a team to have a fully functional waypoint menu if they wanted it. I just went that way because for my understanding it had been decided that the spacetime/manchette had to remain as neutral as possible.

But I agree with you that right now we are in a middle ground :)

ui-manchette/src/types.ts
emersion: After discussion, it seems like the consensus is to pass the whole menu component from OSRD.

ui-manchette/src/types.ts
```diff
@@ -51,7 +51,12 @@ export type OperationalPointType = ArrayElement<PathProperties['operational_poin
 export type StyledOperationalPointType = OperationalPointType & {
   styles?: CSSProperties;
   display?: boolean;
-  onClick?: (opId: string, opRef: HTMLDivElement | null) => void;
+  onClick?: (waypointIndex: number) => void;
+};
+
+export type WaypointMenuData = {
+  menu: React.ReactNode;
+  activeWaypointIndex?: number;
```
emersion: Once #651 is merged, we'll be able to switch back to a waypoint ID here if we want to (and in the `onClick` callback)

SharglutDev: Ok so do I leave it like that and it will be dealt with in #651 or do we want to wait for it to be done ?

emersion: I suppose it depends in which order the PRs get merged…

ui-manchette/src/styles/manchette.css
```diff
@@ -1,5 +1,6 @@
 .manchette-container {
   .manchette-actions {
+    z-index: 11;
```
emersion: Not 100% sure we want to place the menu behind the toolbar… Wouldn't this always hide the menu for the very last waypoint?

out

emersion (approved):
LGTM apart from this last question!

ui-manchette/src/styles/manchette.css
SharglutDev: It was a request from @Tguisnet but you're right we will have an issue for the last waypoints !

ui-manchette-with-spacetimechart/src/stories/useElementInView.ts
```diff
@@ -0,0 +1,48 @@
...
+      ([entry]) => {
+        setIsInView(entry.isIntersecting);
+      },
+      {
+        ...options,
+        root,
+      }
+    );
+
+    observer.observe(currentRef.parentElement!);
```
emersion: In the `Waypoint` component the ref is moved to a child, and here we grab the parent: could we just leave the ref on the parent in `Waypoint` instead?

But IIRC watching the parent causes issues in proportional mode when scrolling up?

ui-manchette-with-spacetimechart/src/stories/useElementInView.ts
```diff
@@ -0,0 +1,48 @@
...
+ * - threshold is used to configure how much of the element needs to be visible before the callback is invoked
+ * - rootMargin is used to configure the margin around the root (container) where the intersection observer will trigger
+ * @returns a boolean indicating if the element is in the viewport
+ */
+const useElementInView = <T extends HTMLElement>(
+  ref?: RefObject<T>,
+  containerRef?: RefObject<HTMLElement>,
+  options: IntersectionObserverInit = {
+    threshold: 1,
+    rootMargin: `0px 0px ${MANCHETTE_ACTIONS_HEIGHT} 0px`,
```
emersion: This hooks sounds like it's generic, but then we're hardcoding details specific to a particular use-cases in the default props. Should we move these defaults to the caller site instead?

ui-manchette/src/types.ts
```diff
@@ -10,7 +10,13 @@ export type Waypoint = {
...
   styles?: CSSProperties;
   display?: boolean;
-  onClick?: (opId: string, opRef: HTMLDivElement | null) => void;
+  onClick?: (waypointId: string, waypointRef: React.RefObject<HTMLDivElement>) => void;
+};
+
+export type WaypointMenuData = {
+  menu: React.ReactNode;
+  activeWaypointId?: string;
+  isWaypointInView?: boolean;
```
emersion: Instead of adding a new `isWaypointInView` field, can't the caller pass a null `menu` field instead? (I believe `React.ReactNode` allows for `null`.)

ui-manchette/src/types.ts
```diff
@@ -10,7 +10,13 @@ export type Waypoint = {
 export type InteractiveWaypoint = Waypoint & {
   styles?: CSSProperties;
   display?: boolean;
-  onClick?: (opId: string, opRef: HTMLDivElement | null) => void;
+  onClick?: (waypointId: string, waypointRef: React.RefObject<HTMLDivElement>) => void;
```
emersion: Note to self: I'm a bit sad to have the caller take care of figuring out whether the waypoint is visible or not, it would be nicer API. Plus the underlying `HTMLDivElement` passed here to the callback can become stale while the menu is displayed (React can choose to re-render it in another DOM element any time).

 But IIRC this causes other Issues™.

ui-manchette/src/components/Waypoint.tsx
```diff
@@ -9,13 +9,18 @@ import { positionMmToKm } from '../utils';
 type WaypointProps = {
   waypoint: InteractiveWaypoint;
   isActive: boolean;
+  waypointMenu?: React.ReactNode;
+  waypointsContainerRef: React.RefObject<HTMLDivElement>;
```
emersion: Sounds like this prop is unused?

ui-manchette-with-spacetimechart/src/stories/useElementInView.ts
SharglutDev: Didn't we check that if the ref was on the parent, the menu was not displayed as expected because the parent height is large when zoomed in ?

ui-manchette-with-spacetimechart/src/stories/useElementInView.ts
SharglutDev: Totally agreed, done.

ui-manchette/src/components/Waypoint.tsx
SharglutDev: Right, done

ui-manchette/src/types.ts
SharglutDev: As discussed, gonna keep the menu visibility in osrd-ui and the caller will only be responsible to pass the container ref to the manchette if he want's to have the hiding behavior.

storybook/.storybook/main.ts
```diff
@@ -24,13 +24,13 @@ const config: StorybookConfig = {
...
-      },
-    });
-  },
+  // async viteFinal(config) {
+  //   return mergeConfig(config, {
+  //     resolve: {
+  //       preserveSymlinks: true,
+  //     },
+  //   });
+  // },
```
emersion: Note: this should be left out of this PR, since these changes are also here: https://github.com/OpenRailAssociation/osrd-ui/pull/721

ui-manchette-with-spacetimechart/src/stories/useElementInView.ts
emersion: Hm, not sure I understand…

I think we should keep passing the same element to `observer.observe()`, but I don't think we should be passing `parentElement`. Instead, we should make it the caller's responsibility to pass a ref targeting the parent element. Does that make sense?

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
```diff
@@ -1,7 +1,6 @@
-/* eslint-disable import/no-unresolved */
 import React, { useRef } from 'react';
 
-import { Manchette } from '@osrd-project/ui-manchette';
+import Manchette from '@osrd-project/ui-manchette';
 import type { ProjectPathTrainResult, Waypoint } from '@osrd-project/ui-manchette/dist/types';
```
emersion: Now that these are exported from `@osrd-project/ui-manchette` (without `dist/`), maybe we could use that instead?

emersion (commented):
Looks good apart from these remaining comments :)

storybook/.storybook/main.ts
SharglutDev: Oups was to hurry to eat the couscous :x

ui-manchette-with-spacetimechart/src/stories/base.stories.tsx
SharglutDev: Good point, done

ui-manchette/src/hooks/useElementInView.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+): boolean => {
+  const [isInView, setIsInView] = useState(false);
+
+  useEffect(() => {
+    const currentRef = ref?.current;
+    const root = containerRef?.current || null;
+
+    if (!currentRef) {
+      // If the user implementing ui-manchette doesn't provide a ref to handle the menu visibility,
```
anisometropie: Why would one use this hook and not provide a ref ?

I guess it would also be the case if the component is not yet loaded, why set isInView as true in that case?

ui-manchette/src/hooks/useElementInView.ts
```diff
@@ -0,0 +1,50 @@
+import { useState, useEffect, type RefObject } from 'react';
```
anisometropie: This seems we could reuse this hook for general purpose, why restrain it to ui-manchette only ?

anisometropie: like move it to ui-core?

ui-manchette/src/hooks/useElementInView.ts
anisometropie: if it’s a generic component it shouldn’t mention manchette specific behavior

anisometropie (changes requested):

anisometropie (changes requested):
the behavior is not consistent.
open the menu, then scroll down, and click and the menu shows on top of the waypoint the next time
https://github.com/user-attachments/assets/d5762a6b-62e8-4454-8fe9-455157bdc641

ui-manchette/src/hooks/useElementInView.ts
SharglutDev: Like we discussed, it was a leftover from the old way I used this hook. I updated the comment to match the new logic.

ui-manchette/src/hooks/useElementInView.ts
SharglutDev: As discussed, I think we can leave it here for now and if it's needed in another subrepo, we can move it at this moment.

ui-manchette/src/hooks/useElementInView.ts
SharglutDev: Same as previous comment. It's only related to the manchette for now

SharglutDev:
After discussion with @thibautsailly, it has been decided to totally block the scroll in the manchette (and only in the manchette) as long as the menu is displayed. If we still try to scroll, the whole page will scroll. (reported in the PR and issue description)

ui-manchette/src/hooks/useElementInView.ts
SharglutDev: Because of design changes, the hook has been removed.

ui-manchette/src/hooks/useElementInView.ts
SharglutDev: See previous comment.

ui-manchette/src/hooks/useElementInView.ts
SharglutDev: See previous comment.

ui-manchette/src/components/Waypoint.tsx
```diff
@@ -6,42 +6,57 @@ import { type InteractiveWaypoint } from '../types';
...
+        <div className="waypoint-separator"></div>
+        <div className="waypoint-ch font-mono justify-self-end">{secondaryCode}</div>
+        <div className="waypoint-separator"></div>
+
+        <div className="waypoint-type"></div>
+        <div className="waypoint-separator"></div>
+        {waypointMenu && isActive && waypointRef.current && (
+          <div
+            className="menu-wrapper"
+            style={{ top: waypointRef.current.getBoundingClientRect().bottom - MENU_OFFSET }}
```
emersion: This performs some layout-related computations in the rendering function. This is unreliable, since at rendering time the DOM hasn't been updated yet.

ui-manchette/src/styles/waypoint.css
```diff
@@ -81,12 +80,22 @@
...
 
+  .menu-wrapper {
+    position: absolute;
+    z-index: 12;
+  }
+}
+
+.last-separator {
+  width: 26px;
+  height: 2rem;
```
emersion: What is the purpose of this?

ui-manchette/src/components/Waypoint.tsx
emersion: This sets a position relative to the viewport, but this breaks as soon as a parent is set as `position: relative`. Since the manchette is a re-usable component, it shouldn't assume anything about the positioning mode of its parent.

I believe this won't work in OSRD since `.chart-container` has `position: relative`.

emersion:
This approach still sounds pretty fragile to me.

- If we really want to block interactions while the menu is opened, I'd suggest [using a portal](https://react.dev/reference/react-dom/createPortal#rendering-to-a-different-part-of-the-dom) to insert a fullscreen transparent overlay preventing all interactions except for the menu, and positioning the menu inside that overlay.
- If we want to allow interactions while the menu is opened, I'd suggest we insert the menu outside of the manchette via a portal, then use `requestAnimationFrame` to update the position of the menu dynamically if the waypoint is visible. Alternatively, give a library such as [floating-ui](https://github.com/floating-ui/floating-ui) a shot.

ui-manchette/src/styles/waypoint.css
SharglutDev: Was needed when removing the position relative for the waypoint. It is not needed anymore with the new logic and has been reverted.

ui-manchette/src/components/Waypoint.tsx
SharglutDev: You're totally right. Logic has been changed and the menu is now positioned relatively to the manchette/spacetime wrapper.

ui-manchette/src/components/Manchette.tsx
```diff
@@ -26,40 +29,82 @@ const Manchette = ({
...
+  useEffect(() => {
+    const manchetteWrapperPosition =
+      waypointMenuData?.manchetteWrapperRef?.current?.getBoundingClientRect().top;
+    const waypointPosition =
+      waypointMenuData?.activeWaypointRef?.current?.getBoundingClientRect().top;
+
+    if (!manchetteWrapperPosition || !waypointPosition) {
+      setMenuPosition(undefined);
+    } else {
+      setMenuPosition(window.scrollY * 2 + manchetteWrapperPosition + waypointPosition);
```
emersion: Hm, not sure how this works, and seems like this doesn't work on my machine… Why multiply the window scroll offset by 2?

The menu has `position: absolute`, and it's inside a wrapper which has `position: relative`. This means we need to compute the distance between these two elements. `waypointPosition - manchetteWrapperPosition` worked well for me.

ui-manchette/src/components/Manchette.tsx
```diff
@@ -26,40 +29,82 @@ const Manchette = ({
...
-          <ZoomIn />
-        </button>
-        <button className="zoom-reset" onClick={resetZoom}>
-          Fit
-        </button>
+}: ManchetteProps) => {
+  const [menuPosition, setMenuPosition] = useState<number>();
+
+  // Allow to track the menu position after we might have scrolled in the page
+  useEffect(() => {
```
emersion: We measure DOM elements from inside this function, so we should use `useLayoutEffect()` instead, so that our callback isn't called before the DOM is updated.

ui-manchette/src/components/Manchette.tsx
```diff
@@ -26,40 +29,82 @@ const Manchette = ({
...
+    if (!manchetteWrapperPosition || !waypointPosition) {
+      setMenuPosition(undefined);
+    } else {
+      setMenuPosition(window.scrollY * 2 + manchetteWrapperPosition + waypointPosition);
+    }
+  }, [waypointMenuData]);
+
+  return (
+    <div className="manchette-container">
+      {waypointMenuData?.menu && menuPosition && (
```
emersion: Nit: would be safer to check for `menuPosition !== undefined`, because 0 is a valid position.

ui-manchette/src/types.ts
```diff
@@ -10,7 +10,14 @@ export type Waypoint = {
...
+};
+
+export type WaypointMenuData = {
+  menu: React.ReactNode;
+  activeWaypointId?: string;
+  activeWaypointRef?: React.RefObject<HTMLDivElement>;
```
emersion: It's a bit cumbersome for the library user to grab the active waypoint ref from the `onClick` callback, then pass it back as a prop to the library. It would be nicer if this ref was entirely managed by the library itself.

To do so, we can make it so `Manchette` creates the ref, and passes it down to the active `Waypoint`. Basically,  we can:

- In `Waypoint`, add a new `waypointRef?: React.Ref<HTMLDivElement>` property
- In `WaypointList`, add a new `activeWaypointRef?: React.Ref<HTMLDivElement>` property and pass `waypointRef={activeWaypointId === waypoint.id ? activeWaypointRef : null}`
- In `Manchette`, create the `activeWaypointRef` ref and pass it to `WaypointList`

And then we can drop `activeWaypointRef` from the story/callback/etc.

ui-manchette/src/components/Manchette.tsx
```diff
@@ -1,15 +1,18 @@
...
 import { ZoomIn, ZoomOut } from '@osrd-project/ui-icons';
 import cx from 'classnames';
 
 import { INITIAL_OP_LIST_HEIGHT, MAX_ZOOM_Y, MIN_ZOOM_Y } from './consts';
 import WaypointList from './WaypointList';
-import type { InteractiveWaypoint } from '../types';
+import type { InteractiveWaypoint, WaypointMenuData } from '../types';
+
+const WAYPOINT_MENU_LEFT_OFFSET = 8;
+const WAYPOINT_MENU_TOP_OFFSET = -2;
```
emersion: Instead of having these consts, it would be nicer to move these into CSS. For instance:

```css
.menu-wrapper {
  left: 8px;
  margin-top: 30px;
}
```

emersion (commented):
We're getting close! A few new comments.

ui-manchette/src/components/Manchette.tsx
SharglutDev: Oh right, I was so focused on trying to handle the scrollY that I didn't think to do it that way.

ui-manchette/src/components/Manchette.tsx
SharglutDev: Still got to keep this one in mind, thank you

ui-manchette/src/components/Manchette.tsx
SharglutDev: Good point, done

ui-manchette/src/types.ts
SharglutDev: You're right it's better ! Done

ui-manchette/src/components/Manchette.tsx
SharglutDev: Done

emersion (approved):
LGTM, thanks a lot for your perseverance through this odyssey!

anisometropie (approved):
tested again, Good job !

anisometropie (commented):
just a bit unsure about disabling completely all the interactions outside of the menu.
it’s pretty natural to close the menu by clicking outside of it

SharglutDev:
> just a bit unsure about disabling completely all the interactions outside of the menu. it’s pretty natural to close the menu by clicking outside of it

It's just the story working like that. In osrd, you will be able to close the menu when clicking outside of it of course.

PR #669, opened 2024-10-25
ui-speedspacechart: use specific d3 sub-packages

emersion:
_See individual commits._

Instead of importing all of d3, only import the two sub-packages
that we actually need: d3-selection and d3-zoom.

This removes 28 unused d3 sub-packages and all of their dependencies.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4675,6 +4675,12 @@
         "@types/node": "*"
       }
     },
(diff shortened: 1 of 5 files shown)


SharglutDev (approved):
Lgtm and tested. Thank you for that :)

anisometropie (approved):
:pray:

PR #683, opened 2024-11-04
ui-core: remove hardcoded label size in checkbox and compensate with small right padding

### LINKED ISSUE #9463: I can tick all the trains from the list even from far away
### What happened?

With my cursor here i can tick the checkbox :
Image

I should only be able to tick the checkbox when the cursor is on it

.

Recette (SNCF)

Google Chrome

dffb4b0

Synar:
Close https://github.com/OpenRailAssociation/osrd/issues/9463

### DIFF
--- a/ui-core/src/styles/inputs/checkbox.css
+++ b/ui-core/src/styles/inputs/checkbox.css
@@ -13,11 +13,12 @@
   font-weight: 400;
   letter-spacing: 0;
   text-align: left;
+  width: fit-content;
 
   .label {
     line-height: 1.5rem;
     padding-top: 0.188rem;
-    width: 12.813rem;
+    padding-right: 0.35rem;
     @apply text-grey-80;
   }
 


ui-core/src/styles/inputs/checkbox.css
```diff
@@ -17,7 +17,7 @@
   .label {
     line-height: 1.5rem;
     padding-top: 0.188rem;
-    width: 12.813rem;
+    padding-right: 0.3rem;
```
SharglutDev: You can merge the two padding together

```suggestion
    padding-inline: 0.188rem 0.3rem;
```

ui-core/src/styles/inputs/checkbox.css
SharglutDev, line 1: This change doesn't seem to be enough. The width of the whole component seems to be taking the one from its parent.

Looks like adding a `width: fit-content;` to the `custom-checkbox` class does the job.

SharglutDev (changes requested):
Thank you for the fix, the bug isn't solved yet

ui-core/src/styles/inputs/checkbox.css
Synar: We are setting the padding top and right, not left and right

SharglutDev: Oh boy, my eyes aren't fixed yet, sorry about that

ui-core/src/styles/inputs/checkbox.css
Synar: Did the bug persist in the scenario page, or are you just talking about the storybook?

I assumed letting the parent control the width and display mode of its checkbox child (whether expanding like in the storybook which is in display block, or not expanding like in the scenario which is in display flex with no grow) was fine.

I don't mind changing this per se though, but this might impact other components like the settings in the spacetimechart, where I assume we want the checkbox to take all the space, although the component could override it

SharglutDev: Hmm maybe we should ask a PO for that. 

In scenario, the actual state of this PR solves the issue but as you said, it doesn't do that for the space time chart settings (see video below) and I think we want that as well.

https://github.com/user-attachments/assets/ec783239-6c8c-4b66-9003-1a16d1be30e9

maelysLeratRosso: I agree, in my opinion one should click inside the box to check it

axrolld: Same

SharglutDev: Only the box ? Not the label ?

Synar: Thibaut settled the issue, we should be able to click the label with a little padding and no more.

SharglutDev: @thibautsailly confirmed we want to be able to click on the label as well to check

SharglutDev (approved):
Lgtm and tested

PR #695, opened 2024-11-07
Fix clear input on blur when no match

achrafmohye:


emersion:
Sounds like a good idea to me.

RomainValls (approved):
LGTM, tested in osrd ✅

ui-core/src/components/inputs/ComboBox/ComboBox.tsx
```diff
@@ -198,8 +198,7 @@ const ComboBox = <T,>({
     } else if (!isInputInSuggestions && selectedOption) {
       setInputValue(getSuggestionLabel(selectedOption));
     } else if (!isInputInSuggestions) {
-      setInputValue('');
-      setSelectedOption(null);
+      clearInput();
```
emersion, line 201: `clearInput()` will `focusInput()` at the end. This is confusing for users: they've navigated away from the input (either by clicking outside, or with `Tab`, or something else), but then focus is set back to the input.

emersion:
The commit name has the "front: " prefix, but it should be "ui-core: ". Also it would be nice to mention that this is about the ComboBox component.

emersion:
Superseded by https://github.com/OpenRailAssociation/osrd-ui/pull/779

PR #696, opened 2024-11-08
speedspacechart: fix blinking in speed limit tags by importing images only once

### LINKED ISSUE #427: Speed space chart is blinking

Synar:
Fix #427

Last PR #633 approached the issue by reducing how often the layer was redrawn. As a result, blinking was a lot less frequent, but each redrawing still caused blinking.

Instead here we reduce the time it takes to draw the canvas, by removing the (async) image fetching from the drawing logic, and fetching them only once.

### DIFF
--- a/ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
+++ b/ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
@@ -13,22 +13,13 @@ import {
 } from '../../const';
 import {
   clearCanvas,
-  createSvgBlobUrl,
   drawLinearLayerBackground,
   drawRoundedRect,
   drawSeparatorLinearLayer,
   drawSvgImageWithColor,
-  loadSvgImage,
   maxPositionValue,
   positionOnGraphScale,
 } from '../../utils';
-
-const questionSvg =
-  '<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16"><path d="M0 8a8 8 0 1 1 16 0A8 8 0 0 1 0 8m8-6.5a6.5 6.5 0 1 0 0 13 6.5 6.5 0 0 0 0-13M6.92 6.085h.001a.75.75 0 1 1-1.342-.67c.169-.339.436-.701.849-.977C6.845 4.16 7.369 4 8 4a2.76 2.76 0 0 1 1.637.525c.503.377.863.965.863 1.725 0 .448-.115.83-.329 1.15-.205.307-.47.513-.692.662-.109.072-.22.138-.313.195l-.006.004a6 6 0 0 0-.26.16 1 1 0 0 0-.276.245.75.75 0 0 1-1.248-.832c.184-.264.42-.489.692-.661q.154-.1.313-.195l.007-.004c.1-.061.182-.11.258-.161a1 1 0 0 0 .277-.245C8.96 6.514 9 6.427 9 6.25a.61.61 0 0 0-.262-.525A1.27 1.27 0 0 0 8 5.5c-.369 0-.595.09-.74.187-.146.1-.263.238-.34.398M9 11a1 1 0 1 1-2 0 1 1 0 0 1 2 0"/></svg>';
-const alertFillSvg =
-  '<svg xmlns="http://www.w3.org/2000/svg" width = "16" height = "16" viewBox = "0 0 16 16" > <path d="M6.457 1.047c.659-1.234 2.427-1.234 3.086 0l6.082 11.378A1.75 1.75 0 0 1 14.082 15H1.918a1.75 1.75 0 0 1-1.543-2.575zM8 5a.75.75 0 0 0-.75.75v2.5a.75.75 0 0 0 1.5 0v-2.5A.75.75 0 0 0 8 5m1 6a1 1 0 1 0-2 0 1 1 0 0 0 2 0" /> </svg>';
-const questionBlobUrl = createSvgBlobUrl(questionSvg);
-const alertFillBlobUrl = createSvgBlobUrl(alertFillSvg);
 
 const RECT_HEIGHT = 17;
 const Y_POSITION = 12;
@@ -43,11 +34,12 @@ const TEXT_PADDING_TOP = 1;
 const ICON_BACKGROUND_WIDTH = 24;
 const ICON_BACKGROUND_HEIGHT = 24;
 
-export const drawSpeedLimitTags = async ({
+export const drawSpeedLimitTags = ({
   ctx,
   width,
   height: marginTop,
   store,
+  images,
 }: SpeedLimitTagsLayerDrawFunctionParams) => {
   const {
     speedLimitTags,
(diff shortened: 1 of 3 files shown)


ui-speedspacechart/src/components/helpers/drawElements/speedLimitTags.ts
```diff
@@ -43,11 +43,17 @@ const TEXT_PADDING_TOP = 1;
 const ICON_BACKGROUND_WIDTH = 24;
 const ICON_BACKGROUND_HEIGHT = 24;
 
-export const drawSpeedLimitTags = async ({
+export const loadSpeedLimitTagsImages = async () => ({
+  questionImage: await loadSvgImage(questionBlobUrl),
+  alertFillImage: await loadSvgImage(alertFillBlobUrl),
+});
```
emersion: Hm, I think loading these as a global variable will cause the images to get loaded at startup for any JS webapp which uses the speed limit tags somewhere in their code. For instance, the STDCM OSRD app would load them at startup even though these images are never used there.

Instead we can do something like this perhaps?

https://github.com/OpenRailAssociation/osrd-ui/blob/b1b380e749f5aeb72dd33e081d34a8d978fd3f26/ui-spacetimechart/src/components/WorkScheduleLayer.tsx#L18

Synar: The code you pointed to is only the export of the helper function, not the actual image loading, which is done by the component in a useeffect with a usestate.

The resulting structure is not too different from the code snippet you shared (we have the image instead of the imageurl as a dependency here, hence we need to have an additional check that the image hasn't been loaded yet, but that's the only difference I think).

Am I missing something?

Synar: Though moving the last remnants of the fetching logic from speedlimitags to speedlimittagslayer would probably make the code cleaner and easier to follow

emersion: Ehh, I missed the `() =>` and thought the `await` was executed when loading the JS file, but this is just a function. Sorry about that!

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
```diff
@@ -1,21 +1,39 @@
...
 const SpeedLimitTagsLayer = ({ width, marginTop, store }: SpeedLimitTagsLayerProps) => {
   const canvas = useRef<HTMLCanvasElement>(null);
   const tooltip = useRef<tooltipInfos | null>();
+  const [images, setImages] = useState<ImagesType>({ questionImage: null, alertFillImage: null });
+
+  useEffect(() => {
+    if (images.questionImage !== null && images.alertFillImage !== null) return;
+    const fetchImages = async () => {
+      const { questionImage, alertFillImage } = await loadSpeedLimitTagsImages();
+      setImages({ questionImage, alertFillImage });
```
emersion: Nit: no need to destructure here, we can store the whole object in a variable (`const images = await …`)

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
```diff
@@ -1,21 +1,39 @@
...
+    };
+    fetchImages();
+  }, [images]);
```
emersion: Hm, this is a bit surprising: this `useEffect()` depends on `images` and also calls `setImages()`. Am I missing something here?

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
```diff
@@ -1,21 +1,39 @@
...
+  computeTooltip,
+  loadSpeedLimitTagsImages,
+} from '../helpers/drawElements/speedLimitTags';
 
 type SpeedLimitTagsLayerProps = {
   width: number;
   marginTop: number;
   store: Store;
 };
+type ImagesType = {
```
emersion: Nit: in general I try to avoid putting "Type" in a type definition because it should already be clear from the naming that it's a type. Suggestion: `LoadedImages`?

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
Synar: The linter wanted me to add images to the dependency, I am not sure why, hence why I also explicitly added a check to skip if the images were already loaded.

The whole thing should disappear in the refactor though, it will then have imageurls as a dep, which will make more sense ^^

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
Synar: Done thanks!

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
Synar: Went further and put even the urls in the useeffect, with an empty dependence array, which should make the code flow as clear as possible

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
```diff
@@ -1,21 +1,42 @@
...
 
 const SpeedLimitTagsLayer = ({ width, marginTop, store }: SpeedLimitTagsLayerProps) => {
   const canvas = useRef<HTMLCanvasElement>(null);
   const tooltip = useRef<tooltipInfos | null>();
+  const [images, setImages] = useState<LoadedImages>({ questionImage: null, alertFillImage: null });
+
+  useEffect(() => {
+    const fetchImages = async () => {
+      const questionSvg =
+        '<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16"><path d="M0 8a8 8 0 1 1 16 0A8 8 0 0 1 0 8m8-6.5a6.5 6.5 0 1 0 0 13 6.5 6.5 0 0 0 0-13M6.92 6.085h.001a.75.75 0 1 1-1.342-.67c.169-.339.436-.701.849-.977C6.845 4.16 7.369 4 8 4a2.76 2.76 0 0 1 1.637.525c.503.377.863.965.863 1.725 0 .448-.115.83-.329 1.15-.205.307-.47.513-.692.662-.109.072-.22.138-.313.195l-.006.004a6 6 0 0 0-.26.16 1 1 0 0 0-.276.245.75.75 0 0 1-1.248-.832c.184-.264.42-.489.692-.661q.154-.1.313-.195l.007-.004c.1-.061.182-.11.258-.161a1 1 0 0 0 .277-.245C8.96 6.514 9 6.427 9 6.25a.61.61 0 0 0-.262-.525A1.27 1.27 0 0 0 8 5.5c-.369 0-.595.09-.74.187-.146.1-.263.238-.34.398M9 11a1 1 0 1 1-2 0 1 1 0 0 1 2 0"/></svg>';
```
emersion, line 29: Orthogonal to this PR: I've been meaning to stop putting SVG data into strings and instead embed them with e.g. `@rollup/plugin-image`.

emersion (approved):
Nice!

ui-speedspacechart/src/components/layers/SpeedLimitTagsLayer.tsx
Synar: I don't quite understand the implications of different image loading methods, but feel free to commit something if you think an improvement can be made in this area

PR #716, opened 2024-11-21
front: fix keyboard arrow scroll in combobox

### LINKED ISSUE #9721: stdcm - Using keyboard arrows don't scroll in selector
### What happened?

We I use the keyboard arrows to go down/up the propositions in an input field (traction engine, towed rolling stock, waypoint...), it does not scroll down/up in the list to show the highlighted line in the list.

I would expect to always see the highlighted line

As above

Dev (SNCF)

Google Chrome

1bd519d

theocrsb:
closes https://github.com/OpenRailAssociation/osrd/issues/9721


SharglutDev (approved):
Lgtm and tested, good job !

Akctarus (approved):
Tested, LGTM

PR #720, opened 2024-11-25
Use ROLLUP_WATCH instead of NODE_ENV

emersion:
We want to enable the livereload plugin when running in watch mode. We don't want to enable the livereload plugin when building, because that causes the build to hang forever.

NODE_ENV is the wrong tool for the job here: it can be set to "development" even when running `npm run build`, for instance if the user has set it globally (on a dev machine). In general, NODE_ENV is just supposed to indicate whether a Node.js app is running in production - nothing else.

Instead, use the ROLLUP_WATCH env var: it's set by rollup specifically when started in watch mode.

### DIFF
--- a/rollup-base.config.js
+++ b/rollup-base.config.js
@@ -7,7 +7,6 @@ import livereload from 'rollup-plugin-livereload';
 import postcss from 'rollup-plugin-postcss';
 
 const formats = ['esm'];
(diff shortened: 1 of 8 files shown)


Yohh (approved):
lgtm

Math-R (approved):
seems logic, lgtm !

PR #721, opened 2024-11-25
storybook: disable preserveSymlinks for vite resolve

emersion:
`preserveSymlinks` is for yarn only, it breaks NPM (keeps an old version of the dependencies around).

To test:

- `npm run watch` in ui-manchette and ui-manchette-with-spacetimechart
- `npm run storybook`
- Open the ui-manchette-with-spacetimechart story in the browser
- Add a `console.log()` in the `Waypoint` component

Before this patch, the log would never show up in the console. With this patch, it should live reload and show up.

Works on my machine™ as well as @SharglutDev's.

### DIFF
--- a/storybook/.storybook/main.ts
+++ b/storybook/.storybook/main.ts
@@ -1,5 +1,4 @@
 import type { StorybookConfig } from '@storybook/react-vite';
-import { mergeConfig } from 'vite';
 
 /** @type { import('@storybook/html-vite').StorybookConfig } */
 const config: StorybookConfig = {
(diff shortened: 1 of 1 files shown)


SharglutDev:
Lgtm for me but maybe it's better if other devs test it as well !

theocrsb (approved):
LGTM and tested

clarani:
It does not work on my machine :/

emersion:
It's not a 100% reliable solution, I've seen it fail on my machine as well. It's just that, before this PR it was never working, and now it's working at least _some_ of the time.

My guess is that there is something wrong with the watched files, and the reason it works sometimes is that `style.css` gets regenerated and triggers a rebuild/reload as a side effect, but it's racy.

SharglutDev:
Doesn't work anymore for me 🙃

emersion:
Oh well, I'll just keep it as a local patch until we figure out the complete solution.

PR #728, opened 2024-11-26
manchette: use height props

Wadjetz:
part of https://github.com/osrd-project/osrd-confidential/issues/872

Use the height prop to enable resizing.

### DIFF
--- a/ui-manchette/src/components/Manchette.tsx
+++ b/ui-manchette/src/components/Manchette.tsx
@@ -30,13 +30,10 @@ const Manchette = ({
   isProportional = true,
   toggleMode,
   children,
-  height,
+  height = INITIAL_OP_LIST_HEIGHT,
 }: ManchetteProps) => (
   <div className="manchette-container">
-    <div
-      className="bg-white-100 border-r border-grey-30"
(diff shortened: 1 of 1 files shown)


ui-manchette/src/components/Manchette.tsx
```diff
@@ -30,11 +30,12 @@ const Manchette = ({
   isProportional = true,
   toggleMode,
   children,
+  height,
 }: ManchetteProps) => (
   <div className="manchette-container">
     <div
       className="bg-white-100 border-r border-grey-30"
-      style={{ minHeight: `${INITIAL_OP_LIST_HEIGHT}px` }}
+      style={{ minHeight: `${INITIAL_OP_LIST_HEIGHT}px`, height }}
```
Math-R: as height can be undefined, i think you should go for 
`height : height | `${INITIAL_OP_LIST_HEIGHT}px``

Plus if we can change the height and make it lower than minHeight then we should remove minHeight

Math-R (changes requested):

Math-R (approved):
lgtm !

PR #745, opened 2024-12-05
ui-manchette: only use 0.5px line on HiDPI screens

### LINKED ISSUE #927: ui-manchette: some lines are invisible on Firefox
### What happened?

On Firefox, with a LoDPI screen:

Image

_No response_

.

Local Storybook, Deployed Storybook

Firefox

emersion:
0.5px lines don't show up reliably on LoDPI screens and Firefox.

Closes: https://github.com/OpenRailAssociation/osrd-ui/issues/927

### DIFF
--- a/ui-manchette/src/styles/waypoint.css
+++ b/ui-manchette/src/styles/waypoint.css
@@ -19,7 +19,10 @@
       bottom: 11px;
       left: 0;
       width: 100%;
-      height: 0.5px;
+      height: 1px;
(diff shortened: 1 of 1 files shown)


SharglutDev (commented):
Thank you for that ! 

With lines having 1px height, it now looks like this : 
Capture d’écran 2024-12-06 à 11 46 46

Lines from manchette and space time seems to be a little shifted.
Should change their position in space time for this kind of resolution or is it too much trouble ?

PR #756, opened 2024-12-10
slider + spacetime

anisometropie:
- new space time story with slider zoom
- adapt useManchetteWithSpaceTime hook to work with slider zoom and mouse scroll zoom

Behavior for https://github.com/OpenRailAssociation/osrd/issues/9723

main ticket https://github.com/osrd-project/osrd-confidential/issues/869

closes https://github.com/OpenRailAssociation/osrd-ui/pull/756

### DIFF
--- a/ui-core/src/components/inputs/Slider.tsx
+++ b/ui-core/src/components/inputs/Slider.tsx
@@ -6,6 +6,7 @@ import cx from 'classnames';
 export type SliderProps = InputHTMLAttributes<HTMLInputElement> & {
   width?: number;
   onChangeCommitted?: (e: MouseEvent<HTMLInputElement>) => void;
+  containerClassName?: string;
 };
 
 // onChange returns an event or number
@@ -19,7 +20,7 @@ const Slider = ({
   onChange,
   onChangeCommitted,
   disabled,
-  className,
+  containerClassName,
   ...rest
 }: SliderProps) => {
   const [value, setValue] = useState<number>(
@@ -43,7 +44,10 @@ const Slider = ({
   };
 
   return (
-    <div className={cx('range-wrapper', className, { disabled })} style={{ width: `${width}px` }}>
+    <div
+      className={cx('range-wrapper', containerClassName, { disabled })}
+      style={{ width: `${width}px` }}
+    >
       <input
         type="range"
         className="range-slider"
--- a/ui-core/src/components/inputs/Slider.tsx
+++ b/ui-core/src/components/inputs/Slider.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import type { ChangeEvent, MouseEvent, InputHTMLAttributes } from 'react';
 
 import cx from 'classnames';
@@ -19,11 +19,16 @@ const Slider = ({
   onChange,
   onChangeCommitted,
   disabled,
+  className,
   ...rest
 }: SliderProps) => {
   const [value, setValue] = useState<number>(
     initialValue !== undefined ? Number(initialValue) : Number(min)
   );
+
+  useEffect(() => {
+    setValue(Number(initialValue));
+  }, [initialValue]);
 
   const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
     const newValue = Number(e.target.value);
@@ -38,7 +43,7 @@ const Slider = ({
   };
 
   return (
-    <div className={cx('range-wrapper', { disabled })} style={{ width: `${width}px` }}>
+    <div className={cx('range-wrapper', className, { disabled })} style={{ width: `${width}px` }}>
       <input
         type="range"
         className="range-slider"
--- a/ui-manchette-with-spacetimechart/src/helpers.ts
+++ b/ui-manchette-with-spacetimechart/src/helpers.ts
@@ -3,12 +3,17 @@ import type {
   ProjectPathTrainResult,
   Waypoint,
 } from '@osrd-project/ui-manchette/dist/types';
-import type {
-  OperationalPoint,
-  SpaceTimeChartProps,
-} from '@osrd-project/ui-spacetimechart/dist/lib/types';
+import type { OperationalPoint } from '@osrd-project/ui-spacetimechart/dist/lib/types';
+import { clamp } from 'lodash';
 
-import { BASE_WAYPOINT_HEIGHT, MAX_TIME_WINDOW, MAX_ZOOM_X, MIN_ZOOM_X } from './consts';
+import {
+  BASE_WAYPOINT_HEIGHT,
+  MAX_TIME_WINDOW,
+  MAX_ZOOM_MS_PER_PX,
+  MAX_ZOOM_X,
+  MIN_ZOOM_MS_PER_PX,
+  MIN_ZOOM_X,
+} from './consts';
 import { calcTotalDistance, getHeightWithoutLastWaypoint, msToS } from './utils';
 
 type WaypointsOptions = { isProportional: boolean; yZoom: number; height: number };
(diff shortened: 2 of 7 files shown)


achrafmohye (approved):
LGTM and tested 👍

ui-core/src/components/inputs/Slider.tsx
```diff
@@ -19,12 +19,17 @@ const Slider = ({
...
+  className,
   ...rest
 }: SliderProps) => {
   const [value, setValue] = useState<number>(
     initialValue !== undefined ? Number(initialValue) : Number(min)
   );
 
+  useEffect(() => {
+    setValue(Number(initialValue));
+  }, [initialValue]);
```
emersion, line 32: Hm, it's a bit weird to have this internal state. Would be nicer to make this a controlled component like other form inputs, and have a single source of truth when it comes to the value.

But that sounds like a larger change.

ui-core/src/components/inputs/Slider.tsx
```diff
@@ -38,7 +43,7 @@ const Slider = ({
   };
 
   return (
-    <div className={cx('range-wrapper', { disabled })} style={{ width: `${width}px` }}>
+    <div className={cx('range-wrapper', className, { disabled })} style={{ width: `${width}px` }}>
```
emersion: Hm, it seems a bit arbitrary to apply part of the props to the wrapper `<div>` and the rest to the `<input>`… This is a bit surprising especially when the `className` field comes from the `HTMLInputElement` type. And it seems like this isn't used?

ui-spacetimechart/src/stories/horizontal-zoom.stories.tsx
```diff
@@ -0,0 +1,170 @@
...
+    zoomValue: timeScaleToZoomValue(DEFAULT_ZOOM_MS_PER_PX),
+    xOffset: offset,
+    yOffset: 0,
+    panning: null,
+  });
+  useEffect(() => {
+    setState((prev) => ({ ...prev, xOffset: offset }));
+  }, [offset]);
+  const handleZoom = (zoomValue: number, position = DEFAULT_WIDTH / 2) => {
+    if (MIN_ZOOM <= zoomValue && zoomValue <= MAX_ZOOM) {
```
emersion: Nit: would be slightly nicer to use `zoomValue = Math.min(MAX_ZOOM, Math.max(MIN_ZOOM, zoomValue))`, that way the zoom bounds are reached when using the mouse wheel  instead of being stuck on weird values close to the limit. (Or use `clamp()` from lodash.)

ui-spacetimechart/src/stories/horizontal-zoom.stories.tsx
```diff
@@ -0,0 +1,170 @@
...
+      const oldTimeScale = zoomValueToTimeScale(state.zoomValue);
+      const newTimeScale = zoomValueToTimeScale(zoomValue);
+      // by default zooming expands the graph around timeOrigin
+      // changing the offset to timeOriginShift alone keeps the left border in place.
+      // subtracting centerShift keeps the center in place
+      // (xOffset = how many px the timeOrigin is shifted to the right)
+      const timeOriginShift = (state.xOffset * oldTimeScale) / newTimeScale;
+      const centerShift = (position * oldTimeScale - position * newTimeScale) / newTimeScale;
+      const newOffset = timeOriginShift - centerShift;
+      setState((prev) => ({ ...prev, zoomValue, xOffset: newOffset }));
```
emersion: The `zoom()` function in `ui-spacetimechart/src/stories/lib/utils.ts` performs a similar task but it has simpler logic.

1. Could we simplify the logic here?
2. Could we use that function instead of re-writing it here?

ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -154,17 +159,32 @@ export const getScales = (
...
+export const timeScaleToZoomValue = (timeScale: number) =>
+  (100 * Math.log(timeScale / MIN_ZOOM_MS_PER_PX)) /
+  Math.log(MAX_ZOOM_MS_PER_PX / MIN_ZOOM_MS_PER_PX);
+
 /** Zoom on X axis and center on the mouse position */
 export const zoomX = (
   currentXZoom: number,
   currentXOffset: number,
-  { delta, position: { x } }: Parameters<NonNullable<SpaceTimeChartProps['onZoom']>>[0]
+  newXZoom: number,
```
emersion: Nit: since this function is named `zoomX`, it's probably fine to drop the `X` here.

ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -154,17 +159,32 @@ export const getScales = (
...
+  // subtracting centerShift keeps the center in place
+  // (xOffset = how many px the timeOrigin is shifted to the right)
+  const timeOriginShift = (currentXOffset * oldTimeScale) / newTimeScale;
+  const centerShift = (position * oldTimeScale - position * newTimeScale) / newTimeScale;
+  const newOffset = timeOriginShift - centerShift;
   return {
-    xZoom,
-    // Adjust zoom level relatively to the input delta value:
-    // These lines are here to center the zoom on the mouse position:
-    xOffset: x - ((x - currentXOffset) / currentXZoom) * xZoom,
```
emersion: Was this old logic not working somehow? Naively, I'd think it'd work fine with the new zoom (modulo some conversions)…

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -54,12 +62,6 @@ const useManchettesWithSpaceTimeChart = (
 
   const paths = usePaths(projectPathTrainResult, selectedTrain);
 
-  // Memoize timeWindow to avoid recalculation on each render
-  const timeWindow = useMemo(
-    () => computeTimeWindow(projectPathTrainResult),
```
emersion, line 59: Hm. Removing this no longer makes it so all trains are visible with the default zoom, I think? Is this intentional? (cc @clarani who implemented this feature IIRC)

emersion (commented):
Thanks for the patch, the storybook example works very well and the consts + exponential math stuff is easy to follow!

ui-core/src/components/inputs/Slider.tsx
anisometropie: This internal state allows to do the `onChangeCommitted` that only update after the user releases the mouse.

ui-core/src/components/inputs/Slider.tsx
anisometropie: The container allows to have a larger region to click on, and I need a way to access this container style to position it when I use it in OSRD.

We could offer two different props 
- `containerClassName` 
- `className` for the input itself

but we can always access with a selector from its parent

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
anisometropie: You’re right, They asked for a default value of 7500 ms/px, which means we only have 2h-2h30 visible range by default and when we reset the zoom.

What I can do though is rework the offset to be sure we’re not showing a completely empty region.

ui-manchette-with-spacetimechart/src/helpers.ts
anisometropie: You’re right, I just needed a few adjustment to make it work (timeScale has an inverse relationship to zoomValue, previously we didn’t have inverse relationship)

much better !

`const newOffset = position - ((position - state.xOffset) * oldTimeScale) / newTimeScale;`

ui-spacetimechart/src/stories/horizontal-zoom.stories.tsx
anisometropie: The slider provides a value in [0;100] as is, but `zoom` function expects a delta. It seems better to keep a clear mapping [0;100] => [625:600000] in this case.

I’ll use the simplification from other comment here too.

`const newOffset = position - ((position - state.xOffset) * oldTimeScale) / newTimeScale;`

ui-spacetimechart/src/stories/horizontal-zoom.stories.tsx
anisometropie: done, I went for clamp

ui-manchette-with-spacetimechart/src/helpers.ts
anisometropie: done, dropped all the xs in the function

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
anisometropie: seems like starting at offset zero is fine all the time, it’s the start of the earliest train.

ui-core/src/components/inputs/Slider.tsx
emersion: > className for the input itself

That's the current behavior FWIW: `className` is currently part of `...rest` and forwarded to the input.

> containerClassName

Sounds fine to me: if we need to control the class name of the wrapper from the outside, I think a property named differently like this is better.

> but we can always access with a selector from its parent

Right, if this doesn't make our life terrible in OSRD, this sounds good as well (and we don't need the new prop).

ui-core/src/components/inputs/Slider.tsx
emersion: Indeed, that's a good point.

ui-core/src/components/inputs/Slider.tsx
anisometropie: ok I went for the containerClassName

ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
emersion: @thibautsailly confirmed it's not desirable to fit the viewport to the selected train.

emersion, line 59: Nit: this was the only caller of `computeTimeWindow()`, so I think this function can be dropped? Could be done later I suppose.

emersion (approved):
LGTM!

PR #758, opened 2024-12-11
Fix max speed utils function

### LINKED ISSUE #9307: GEV speed limit too high to be displayed
### What happened?

When a train does not go to high speed, the maximum speed limit cannot be displayed considering the limited in height graph

Example with a train on a small itinerary:

Image

If i stretch the graph I can barely see it:

Image

that the whole vertical axis is based on the maximum speed limit and not the actual train speed

1. create itinerary with SBG - RDG
2. look at GEV

Recette (SNCF)

Firefox

dev 96c960e

flomonster:
close https://github.com/OpenRailAssociation/osrd/issues/9307

When the speed limits layer is enabled the max speed value to compute must be computed from the mrsp field.

**Render:**
image

image

### DIFF
--- a/ui-speedspacechart/src/__tests__/utils.spec.ts
+++ b/ui-speedspacechart/src/__tests__/utils.spec.ts
@@ -95,7 +95,7 @@ describe('getGraphOffsets', () => {
 
 describe('maxSpeedValue', () => {
   it('should return the correct maxSpeed', () => {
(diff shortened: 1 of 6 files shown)


clarani (approved):
LGTM :white_check_mark:

Yohh (approved):
lgtm

PR #761, opened 2024-12-13
ui-manchette/fix waypoint menu closing

### LINKED ISSUE #8629: Manchette : hide a waypoint by clicking on it
### Description and goal

In the manchette, we want to be able to hide a waypoint when clicking on it.

To work, this issue https://github.com/OpenRailAssociation/osrd-ui/issues/399 needs be done first.

### Acceptance criteria

- [x] When clicking on a waypoint, an options menu should be displayed
- [x] "Hide this waypoint" should be the only option visible for now
- [x] The menu's design should match the [sketch](https://www.sketch.com/s/a26d84be-5fb2-41e4-9e84-96d90f46a997/a/OeJV9yG) one 
- [x] Clicking on "Hide this waypoint" should close the menu and properly hide the waypoint
- [x] The space time chart should resize accordingly
- [x] The hidden waypoint should still be present in the waypoint list modal so it can't be displayed again
- [ ] When the menu is open, clicking on another waypoint or the same one just closes the menu and we need one more click to open another one
- [x] When the menu is open, the scroll in the manchette, the manchette toolbar and pan in space time chart are disabled
- [ ] If there was only two waypoints left, it shouldn't be possible to hide it (behavior needs to be checked with @thibautsailly)

SharglutDev:
See commits

close https://github.com/OpenRailAssociation/osrd/issues/8629 (see last comments for the issue)

### DIFF
--- a/ui-manchette-with-spacetimechart/src/stories/base-with-waypoint-menu.stories.tsx
+++ b/ui-manchette-with-spacetimechart/src/stories/base-with-waypoint-menu.stories.tsx
@@ -4,11 +4,11 @@ import { EyeClosed, Telescope } from '@osrd-project/ui-icons';
 import Manchette, { type ProjectPathTrainResult, type Waypoint } from '@osrd-project/ui-manchette';
 import { PathLayer, SpaceTimeChart } from '@osrd-project/ui-spacetimechart';
 import type { Meta } from '@storybook/react';
-import cx from 'classnames';
-
 import '@osrd-project/ui-core/dist/theme.css';
 import '@osrd-project/ui-manchette/dist/theme.css';
 import '@osrd-project/ui-manchette-with-spacetimechart/dist/theme.css';
+import cx from 'classnames';
+import { createPortal } from 'react-dom';
 
 import Menu, { type MenuItem } from './Menu';
 import { SAMPLE_WAYPOINTS, SAMPLE_PATHS_DATA } from '../assets/sampleData';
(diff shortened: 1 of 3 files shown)


ui-manchette-with-spacetimechart/src/stories/base-with-waypoint-menu.stories.tsx
```diff
@@ -69,6 +77,36 @@ const ManchetteWithSpaceTimeWrapper = ({
...
+        setIsClickOnWaypoint(true);
+      }
+      // Close the menu if the user clicks outside of it
+      if (!menuRef.current?.contains(event.target as Node)) {
+        setActiveWaypointId(undefined);
+      }
+    };
+
+    const toggleButton = document.querySelector('.toggle-mode div');
+    const zoomButtons = document.querySelectorAll('.zoom-buttons span');
```
emersion: This is not safe to do in a `useEffect()`.

emersion:
Hm, not a fan of these "clicked outside" event handlers with ad-hoc checks. Would prefer a transparent div covering the whole document (with a portal).

ui-manchette-with-spacetimechart/src/stories/base-with-waypoint-menu.stories.tsx
SharglutDev: Layout then ? 😖

SharglutDev:
> Hm, not a fan of these "clicked outside" event handlers with ad-hoc checks. Would prefer a transparent div covering the whole document (with a portal).

We need to discuss with @thibautsailly if having the waypoint menu displayed should block any first click interaction with the whole page (if yes we could maybe just use a modal ?) or just the manchette/spacetimechart area.

SharglutDev:
Added the `createPortal` logic :)

emersion (approved):
LGTM :)

theocrsb (approved):
LGTM and tested

PR #764, opened 2024-12-16
manchette space time zoom

### LINKED ISSUE #8179: Manchette: Zoom behavior needs improvement
### What happened?

Manchette is displaying interval [29; 56].
when zooming, it displays a completely disjoint interval [10.5; 12.6]

https://github.com/user-attachments/assets/b363fe29-4a19-405b-af94-3953a26ade06

zooming should not scroll, we should stay inside of the original [29; 56] (something like [35;45])

Local

chrome

anisometropie:
keep the scroll relative position when zooming.

closes https://github.com/OpenRailAssociation/osrd/issues/8179

### DIFF
--- a/ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
+++ b/ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
@@ -29,7 +29,10 @@ type State = {
   xZoom: number;
   yZoom: number;
   xOffset: number;
+  /** the current y-scroll of the view. always updates */
   yOffset: number;
+  /** only update after a zoom. used to update back the view scroll value */
+  scrollTo: number | null;
   panning: { initialOffset: { x: number; y: number } } | null;
   isProportional: boolean;
   waypointsChart: Waypoint[];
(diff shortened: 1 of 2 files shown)


ui-manchette-with-spacetimechart/src/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -78,15 +80,39 @@ const useManchettesWithSpaceTimeChart = (
...
+      const newYZoom = yZoom - ZOOM_Y_DELTA;
+      const newYOffset = yOffset * (newYZoom / yZoom);
+      setState((prev) => ({
+        ...prev,
+        yZoom: newYZoom,
+        yOffset: newYOffset,
+        scrollTo: newYOffset,
+      }));
+    }
+  }, [yZoom, yOffset]);
```
Yohh, line 106: what do you think about combining it in a single function like this?

```js
  const zoomY = useCallback(() => {
    let newYZoom = yZoom;
    if (yZoom < MAX_ZOOM_Y) {
      newYZoom = yZoom + ZOOM_Y_DELTA;
    }
    if (yZoom > MIN_ZOOM_Y) {
      newYZoom = yZoom - ZOOM_Y_DELTA;
    }
    const newYOffset = yOffset * (newYZoom / yZoom);

    setState((prev) => ({
      ...prev,
      yZoom: newYZoom,
      yOffset: newYOffset,
      scrollTo: newYOffset,
    }));
  }, [yZoom, yOffset]);
```

Yohh: nevermind, if we do that you will have to modify ui-manchette

Yohh (approved):

Math-R (approved):
lgtm

PR #773, opened 2024-12-18
ui-core: add close button to pickers

### LINKED ISSUE #693: Add closing `X` in `date_picker`, `time_picker`, and `tolerance_picker`
Image
Image

If https://github.com/osrd-project/osrd-confidential/issues/849 is being done at the same time:
Image

Otherwise:
Image

## AC
- when bumping the new release in osrd, don't forget to update the stdcm test to close the tolerance modal in `stdcm-page-model.ts` l.382

SharglutDev:
close #693 

[Sketch](https://www.sketch.com/s/ca1f66cf-046d-4992-ae63-62187269ba7e/p/C50E2825-1B58-4269-A53A-5CE04E024E50/canvas) with mockup on the date picker

### DIFF
--- a/ui-core/src/components/Modal.tsx
+++ b/ui-core/src/components/Modal.tsx
@@ -1,5 +1,7 @@
 import React, { useEffect, useRef } from 'react';
 
+import { X } from '@osrd-project/ui-icons';
+
 import { useModalPosition } from '../hooks/useModalPosition';
 import useOutsideClick from '../hooks/useOutsideClick';
 
@@ -31,6 +33,9 @@ const InputModal = ({ inputRef, isOpen, onClose, children }: ModalProps) => {
         className="modal-content"
         style={{ top: modalPosition.top, left: modalPosition.left }}
       >
+        <button className="close-button" onClick={onClose}>
+          <X size="lg" />
+        </button>
         {children}
       </div>
     </div>
--- a/ui-core/src/styles/inputs/datePicker.css
+++ b/ui-core/src/styles/inputs/datePicker.css
@@ -13,17 +13,10 @@
   }
 
   .calendar-picker {
-    position: absolute;
-    width: fit-content;
-    border-radius: 0.5rem;
-    box-shadow:
-      0 0.813rem 0.938rem -0.438rem rgba(31, 27, 23, 0.1),
-      0 0.188rem 0.375rem rgba(31, 27, 23, 0.2);
     display: flex;
     flex-direction: row;
     align-items: center;
     justify-content: center;
-    @apply bg-white-100;
 
     .calendar-navigation-btn {
       width: 2rem;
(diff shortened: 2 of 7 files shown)


ui-core/src/styles/inputs/datePicker.css
```diff
@@ -25,6 +25,21 @@
...
+
+      &:hover {
+        color: theme('colors.grey.90');
+      }
+
+      &:active {
+        color: theme('colors.grey.20');
+      }
+    }
+
```
Yohh: maybe we could add a `common.css` file for those kind of classes

ui-core/src/components/Modal.tsx
```diff
@@ -1,5 +1,7 @@
 import React, { useEffect, useRef } from 'react';
 
+import { X } from '@osrd-project/ui-icons';
+
```
Yohh, line 4: note for the future: we really need to figure out what's happening with those bloody imports

ui-core/src/styles/inputs/datePicker.css
SharglutDev: I can, but as `.close-button` won't be nested inside its parent class, we just need to be vigilant in the future that we don't create a similar class that overrides it. Or do you think it's safe anyway ?

Yohh: if you fear about that you can do 

```css
.calendar-picker,
.modal-content {
  .close-button {
    position: absolute;
    top: 16px;
    right: 16px;
    color: theme('colors.grey.50');

    &:hover {
      color: theme('colors.grey.90');
    }

    &:active {
      color: theme('colors.grey.20');
    }
  }
}
```

kmer2016: I think the right approach is to use the Modal component in the datePicker as well. This would naturally fix the duplication of the css, but it would also factorize the code.

Yohh: @kmer2016 what do you think about it?

SharglutDev: Hmm I wouldn't mind but seems it goes out of the scope of this PR

SharglutDev: Ok so I rebased in one commit and added a detailed description

ui-core/src/styles/inputs/datePicker.css
```diff
@@ -13,17 +13,25 @@
...
+      color: theme('colors.grey.50');
+
+      &:hover {
+        color: theme('colors.grey.90');
+      }
+
+      &:active {
+        color: theme('colors.grey.20');
+      }
+    }
```
kmer2016: No need for that bit of code since it's already managed in the modal's css, right?
```suggestion

```

kmer2016 (commented):
Just a suggestion for a little cleaning

ui-core/src/styles/inputs/datePicker.css
SharglutDev: Oops yes forgot about that, done

kmer2016 (approved):
Well done ! Thank you for the refacto.

Yohh (approved):

PR #776, opened 2024-12-19
front: ui-core input tooltip status messages

### LINKED ISSUE #770: Narrow display of text input messages
### Who would benefit from this feature?

End users

### What is this feature about?

- adding an alternate display for text input messages for situations where the layout doesn't leave enough room horizontally for the coloured background and the icon.

### Why is this feature valuable?

This feature matters because it makes it possible to display error messages in forms where text inputs are layed out closely, and all situations that are horizontally challenged estate wise, like on mobile screens.

### Anything else people should know?

Here is the mockup. Please refer to the Sketch document to get all the dimensions and colour specs : `Sketch > Librairies > OSRD Web > Form > Text input`.
Image

Wadjetz:
fix #770

Capture d’écran 2024-12-19 à 17 01 11

### DIFF
--- a/ui-core/src/components/inputs/Input.tsx
+++ b/ui-core/src/components/inputs/Input.tsx
@@ -71,6 +71,7 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
       withIcons = [],
       onKeyUp,
       onBlur,
+      onCloseStatusMessage,
       ...rest
     },
     ref
@@ -86,7 +87,9 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
         disabled={disabled}
         required={required}
         small={small}
+        statusIconPosition={statusWithMessage?.tooltip ? 'before-status-message' : undefined}
         className={cx('input-field-wrapper', inputFieldWrapperClassname)}
+        onCloseStatusMessage={onCloseStatusMessage}
       >
         <div
           className={cx('input-wrapper', {
--- a/ui-core/src/components/inputs/StatusMessage.tsx
+++ b/ui-core/src/components/inputs/StatusMessage.tsx
@@ -1,12 +1,15 @@
 import React from 'react';
 
+import { X } from '@osrd-project/ui-icons';
 import cx from 'classnames';
 
 import InputStatusIcon from './InputStatusIcon';
 
 export type Status = 'success' | 'info' | 'error' | 'warning' | 'loading';
+export type TooltipType = 'left' | 'right';
 
 export type StatusWithMessage = {
+  tooltip?: TooltipType;
   status: Status;
   message?: string;
 };
@@ -15,16 +18,30 @@ export type StatusMessageProps = {
   statusWithMessage: StatusWithMessage;
   showIcon?: boolean;
   small?: boolean;
+  onClose?: () => void;
 };
 
-const StatusMessage = ({ statusWithMessage, showIcon, small }: StatusMessageProps) => {
-  const { status, message } = statusWithMessage;
+const StatusMessage = ({ statusWithMessage, showIcon, small, onClose }: StatusMessageProps) => {
+  const { tooltip, status, message } = statusWithMessage;
   if (message === undefined) return null;
 
   return (
-    <div className="status-message-wrapper">
+    <div
+      className={cx({
+        'status-message-wrapper': !tooltip,
+        'status-message-wrapper-tooltip': tooltip,
+        'tooltip-left': tooltip === 'left',
+        'tooltip-right': tooltip === 'right',
+        [status]: status,
+      })}
+    >
       {showIcon && <InputStatusIcon status={status} small={small} />}
       <span className={cx('status-message', { [status]: status })}>{message}</span>
+      {status === 'info' && (
+        <button className="status-close" onClick={onClose}>
+          <X size="sm" />
+        </button>
+      )}
     </div>
   );
 };
--- a/ui-core/src/stories/Input.stories.tsx
+++ b/ui-core/src/stories/Input.stories.tsx
@@ -1,10 +1,11 @@
-import React from 'react';
+import React, { useState } from 'react';
 
 import { ChevronDown, X } from '@osrd-project/ui-icons';
 import type { Meta, StoryObj } from '@storybook/react';
 import '@osrd-project/ui-core/dist/theme.css';
 
 import Input from '../components/inputs/Input';
+import { type StatusWithMessage } from '../components/inputs/StatusMessage';
 
 const meta: Meta<typeof Input> = {
   component: Input,
@@ -171,6 +172,80 @@ export const ErrorInput: Story = {
       message: '“Michel Sardou” can’t be used',
     },
   },
+};
+
+export const TooltipErrorInput: Story = {
+  args: {
+    label: 'Name',
+    type: 'text',
+    required: true,
+    value: 'Michel Sardou',
+    statusWithMessage: {
+      tooltip: 'right',
+      status: 'error',
+      message: '“Michel Sardou” can’t be used',
+    },
+  },
+};
+
+export const TooltipInfoInput: Story = {
+  args: {
+    label: 'Name',
+    type: 'text',
+    required: true,
+    value: 'Michel Sardou',
+  },
+  decorators: [
+    function Component(Story, ctx) {
+      const [status, setStatus] = useState<StatusWithMessage | undefined>({
+        tooltip: 'right',
+        status: 'info',
+        message: '“Michel Sardou” can’t be used',
+      });
+      return (
+        <Story
+          args={{
+            ...ctx.args,
+            statusWithMessage: status,
+            onCloseStatusMessage: () => setStatus(undefined),
+          }}
+        />
+      );
+    },
+  ],
+};
+
+export const TwoTooltipErrorInput: Story = {
+  decorators: [
+    (Story) => (
+      <div style={{ display: 'flex', flexDirection: 'row', gap: '20px' }}>
+        <Story
+          args={{
+            label: 'Name',
+            type: 'text',
+            value: 'Michel Sardou',
+            statusWithMessage: {
+              tooltip: 'left',
+              status: 'error',
+              message: 'Michel Sardou can’t be used',
+            },
+          }}
+        />
+        <Story
+          args={{
+            label: 'Name',
+            type: 'text',
+            value: 'Jean-Michel Halleurt',
+            statusWithMessage: {
+              tooltip: 'right',
+              status: 'error',
+              message: 'Jean-Michel Halleurt can’t be used',
+            },
+          }}
+        />
+      </div>
+    ),
+  ],
 };
 
 export const ErrorWithoutMessageInput: Story = {
--- a/ui-core/src/styles/inputs/fieldWrapper.css
+++ b/ui-core/src/styles/inputs/fieldWrapper.css
@@ -1,17 +1,25 @@
 .feed-back {
+  --input-wrapper-padding: 3px;
+  --field-wrapper-padding-bottom: 16px;
+  --status-message-wrapper-tooltip--offset: 4px;
+
   border-radius: 0.5rem;
   position: relative;
   padding: 0.625rem 0.813rem 1rem 1rem;
 
-  &.info {
+  &.success:not(:has(.status-message-wrapper-tooltip)) {
+    @apply bg-success-5;
+  }
+
+  &.info:not(:has(.status-message-wrapper-tooltip)) {
     @apply bg-info-5;
   }
 
-  &.warning {
+  &.warning:not(:has(.status-message-wrapper-tooltip)) {
     @apply bg-warning-5;
   }
 
-  &.error {
+  &.error:not(:has(.status-message-wrapper-tooltip)) {
     @apply bg-error-5;
   }
 
--- a/ui-core/src/styles/inputs/statusMessage.css
+++ b/ui-core/src/styles/inputs/statusMessage.css
@@ -31,3 +31,84 @@
     }
   }
 }
+
+.status-message-wrapper-tooltip {
+  position: absolute;
+  display: flex;
+  align-items: center;
+  width: 320px;
+  top: calc(
+    100% - var(--input-wrapper-padding) - var(--field-wrapper-padding-bottom) -
+      var(--status-message-wrapper-tooltip--offset)
+  );
+  gap: 12px;
+  border-radius: 8px;
+  box-shadow:
+    0 10px 20px 0 rgba(0, 0, 0, 0.2),
+    0 1px 0 rgba(255, 255, 255, 0.6) inset;
+  padding-inline: 12px;
+  z-index: 10;
+
+  .status-close {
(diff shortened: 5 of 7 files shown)


ui-core/src/styles/inputs/fieldWrapper.css
```diff
@@ -28,5 +32,53 @@
         margin-left: 0.563rem;
       }
     }
+
+    input:focus:has(.status-message-wrapper-tooltip) .status-message-wrapper-tooltip {
+      z-index: 2;
+    }
+
+    .status-message-wrapper-tooltip {
```
SharglutDev: You can put all this style in `statusMessage.css`

ui-core/src/styles/inputs/fieldWrapper.css
SharglutDev, line 1: This tooltip message won't be used only for errors. We need to be able to display it for info/warning etc.
There will also be a special case when the status is `info` where we need to have a close button (mockup should inc today)

ui-core/src/stories/Input.stories.tsx
SharglutDev, line 1: It would be nice if you could have access to `statusWithMessage` props in the story to play with it, also the `small` props to see how it behaves.

ui-core/src/styles/inputs/fieldWrapper.css
```diff
@@ -28,5 +32,53 @@
...
+      padding: 8px 12px;
+      min-height: 64px;
+
+      z-index: 10;
+      transition: var(--tooltip-transition-duration) ease;
+
+      .status-message {
+        opacity: 1;
+        color: theme('colors.error.60');
+        font-size: 16px;
```
SharglutDev: That's the default font-size
```suggestion
```

ui-core/src/styles/inputs/fieldWrapper.css
```diff
@@ -28,5 +32,53 @@
...
+        font-weight: 600;
+        font-style: SemiBold;
```
SharglutDev: You can remove this, `font-weight` does the job already
```suggestion
```

ui-core/src/styles/inputs/fieldWrapper.css
```diff
@@ -28,5 +32,53 @@
...
+        letter-spacing: 0px;
+        text-align: left;
```
SharglutDev: Not needed
```suggestion
```

ui-core/src/styles/inputs/fieldWrapper.css
```diff
@@ -28,5 +32,53 @@
...
+      background-color: theme('colors.error.5');
+      opacity: 1;
+      padding: 8px 12px;
+      min-height: 64px;
+
+      z-index: 10;
+      transition: var(--tooltip-transition-duration) ease;
+
+      .status-message {
+        opacity: 1;
```
SharglutDev: Not needed
```suggestion
```

ui-core/src/styles/inputs/fieldWrapper.css
```diff
@@ -28,5 +32,53 @@
...
+      align-items: center;
+      width: 320px;
+      top: 60px;
+      gap: 12px;
+      border-radius: 8px;
+      box-shadow:
+        0px 10px 20px 0px rgba(0, 0, 0, 0.2),
+        0px 1px 0px rgba(255, 255, 255, 0.6) inset;
+      background-color: theme('colors.error.5');
+      opacity: 1;
```
SharglutDev:
```suggestion
      box-shadow:
        0 10px 20px 0 rgba(0, 0, 0, 0.2),
        0 1px 0 rgba(255, 255, 255, 0.6) inset;
      background-color: theme('colors.error.5');
```

ui-core/src/styles/inputs/fieldWrapper.css
```diff
@@ -28,5 +32,53 @@
...
+      padding: 8px 12px;
+      min-height: 64px;
```
SharglutDev: Not sure we need this ? If the message takes only one line maybe it doesn't have to be that high.
@thibautsailly is working on a mockup to define this case (hopefully should be up today)

ui-core/src/styles/inputs/fieldWrapper.css
```diff
@@ -28,5 +32,53 @@
...
+      width: 320px;
+      top: 60px;
+      gap: 12px;
+      border-radius: 8px;
+      box-shadow:
+        0px 10px 20px 0px rgba(0, 0, 0, 0.2),
+        0px 1px 0px rgba(255, 255, 255, 0.6) inset;
+      background-color: theme('colors.error.5');
+      opacity: 1;
+      padding: 8px 12px;
```
SharglutDev: Not quite the values in mockup : 
Capture d’écran 2024-12-20 à 09 24 22

```suggestion
      padding: 6px 12px 10px;
```

ui-core/src/styles/inputs/fieldWrapper.css
```diff
@@ -28,5 +32,53 @@
...
+    input:focus:has(.status-message-wrapper-tooltip) .status-message-wrapper-tooltip {
+      z-index: 2;
+    }
+
+    .status-message-wrapper-tooltip {
+      position: absolute;
+      display: flex;
+      align-items: center;
+      width: 320px;
+      top: 60px;
```
SharglutDev: To follow easier the mockup behavior (4px offset), maybe we can use this (the tooltip is too high atm)
```suggestion
      top: calc(100% - inputMargin - fieldMargin - 4px);
```

ui-core/src/components/inputs/Input.tsx
```diff
@@ -86,6 +86,7 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
         disabled={disabled}
         required={required}
         small={small}
+        statusIconPosition="before-status-message"
```
SharglutDev: This is already in 'next-to-field' by default in `FieldWrapper.tsx`, we shouldn't add this

ui-core/src/components/inputs/StatusMessage.tsx
```diff
@@ -18,11 +20,19 @@ export type StatusMessageProps = {
...
 
   return (
-    <div className="status-message-wrapper">
+    <div
+      className={cx({
+        'status-message-wrapper': !tooltip,
+        'status-message-wrapper-tooltip': tooltip,
+        'tooltip-left': tooltip === 'left',
+        'tooltip-right': tooltip === 'right',
+        [status]: status,
```
SharglutDev, line 35: This doesn't seem to be enough, the text and background color change only for the error status

ui-core/src/stories/Input.stories.tsx
```diff
@@ -173,6 +173,56 @@ export const ErrorInput: Story = {
...
+    label: 'Name',
+    type: 'text',
+    required: true,
+    value: 'Michel Sardou',
+    statusWithMessage: {
+      tooltip: 'right',
+      status: 'error',
+      message: '“Michel Sardou” can’t be used',
+    },
+    statusIconPosition: 'before-status-message',
```
SharglutDev: Control this props here means I can make an error when importing the component. Maybe we should have so check in osrd-ui side like "if tooltip is defined, I want to have `statusIconPosition='before-status-message'`"

ui-core/src/styles/inputs/fieldWrapper.css
```diff
@@ -28,5 +32,48 @@
...
+        0px 10px 20px 0px rgba(0, 0, 0, 0.2),
+        0px 1px 0px rgba(255, 255, 255, 0.6) inset;
+      padding: 6px 12px 10px;
+      min-height: 64px;
+      z-index: 10;
+      transition: var(--tooltip-transition-duration) ease;
+
+      .status-message {
+        font-weight: 600;
+        text-align: left;
```
SharglutDev:
```suggestion
```

ui-core/src/styles/inputs/fieldWrapper.css
SharglutDev: still missing the `0px` to `0` in the `box-shadow` :)

ui-core/src/styles/inputs/statusMessage.css
```diff
@@ -31,3 +31,60 @@
...
+  }
+
+  &.info {
+    background-color: theme('colors.info.5');
+  }
+
+  &.warning {
+    background-color: theme('colors.warning.5');
+  }
+
```
SharglutDev, line 85: You can also add success (same for `.status-message` `color.60`)
```suggestion
  &.success {
    background-color: theme('colors.success.5');
  }
```

ui-core/src/styles/inputs/statusMessage.css
```diff
@@ -31,3 +31,60 @@
...
+  &.tooltip-left {
+    left: -5px;
+  }
+
+  &.tooltip-right {
+    right: -5px;
+  }
```
SharglutDev: The offset is from the text field edges 

Capture d’écran 2024-12-20 à 14 39 10

```suggestion
  &.tooltip-left {
    left: 8px;
  }

  &.tooltip-right {
    right: 8px;
  }
```

ui-core/src/components/inputs/StatusMessage.tsx
SharglutDev, line 1: There is an issue in `InputStatusIcon.tsx`, the `Info` icon should be the `variant="fill"`

ui-core/src/components/inputs/Input.tsx
```diff
@@ -86,6 +86,11 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
         disabled={disabled}
         required={required}
         small={small}
+        statusIconPosition={
+          statusWithMessage?.tooltip === 'left' || statusWithMessage?.tooltip === 'right'
+            ? 'before-status-message'
+            : undefined
+        }
```
SharglutDev: This should be enough
```suggestion
        statusIconPosition={statusWithMessage?.tooltip ? 'before-status-message' : undefined}
```

ui-core/src/styles/inputs/statusMessage.css
```diff
@@ -31,3 +31,60 @@
...
+  box-shadow:
+    0 10px 20px 0 rgba(0, 0, 0, 0.2),
+    0 1px 0 rgba(255, 255, 255, 0.6) inset;
+  padding: 6px 12px 10px;
+  min-height: 64px;
+  z-index: 10;
+
+  &.small {
+    top: 58px;
+  }
```
SharglutDev: Sorry to insist, but if `top` was calculated with `top: calc(100% - 16px - 3px - 4px);`, you shouldn't need to add this `.small` style here and the `small` props to `StatusMessage`. (16px and 3px are `padding-bottom` from `.feed-pack` and `.input-wrapper`).
We could even add in main.css. to improve maintainability :
```
:root {
  --input-wrapper-padding: 3px;
  --field-wrapper-padding-bottom: 16px;
  --status-message-wrapper-tooltip--offset: 4px;
}
```
and use them in for `top` in `status-message.css` (gonna ask for opinions in #ui channel)

SharglutDev:
Lgtm, except the last comment where we need @thibautsailly 's mockup

thibautsailly:
Sorry, I haven't been able to finish this before the holidays. The Sketch library "OSRD Web" is updated with the requested informations.

image

ui-core/src/components/inputs/StatusMessage.tsx
```diff
@@ -1,30 +1,50 @@
...
+        'status-message-wrapper-tooltip': tooltip,
+        'tooltip-left': tooltip === 'left',
+        'tooltip-right': tooltip === 'right',
+        'tooltip-close': showClose && close,
+        [status]: status,
+      })}
+    >
       {showIcon && <InputStatusIcon status={status} small={small} />}
       <span className={cx('status-message', { [status]: status })}>{message}</span>
+      {showClose && (
```
SharglutDev: As we already have access to `statusWithMessage` in this component, I don't think we need the `showClose` prop, we could just do

```suggestion
      {statusWithMessage.status === 'info' && (
```

ui-core/src/components/inputs/StatusMessage.tsx
```diff
@@ -1,30 +1,50 @@
...
 
   return (
-    <div className="status-message-wrapper">
+    <div
+      className={cx({
+        'status-message-wrapper': !tooltip,
+        'status-message-wrapper-tooltip': tooltip,
+        'tooltip-left': tooltip === 'left',
+        'tooltip-right': tooltip === 'right',
+        'tooltip-close': showClose && close,
```
SharglutDev: Having a `display: none` means the element isn't remove from the DOM when we click on the close button which isn't ideal I think.

```suggestion
```

We could just add `if (close) return null;` after the other `if` l.28

ui-core/src/components/inputs/StatusMessage.tsx
```diff
@@ -1,30 +1,50 @@
 import React from 'react';
 
+import { X } from '@osrd-project/ui-icons';
+
```
SharglutDev: You need to remove this extra line to get rid off the warning
```suggestion
```

ui-core/src/components/inputs/StatusMessage.tsx
SharglutDev, line 1: We still have one issue : when we close the tooltip when info with the close button, the input is still styled as if the tooltip was there which isn't what we want I think (even though this case isn't reproduce on the mockup).

Wadjetz: Yes, it happens because the component doesn’t have control over the status. I added an onClose callback to remove the status.

SharglutDev: Ok that's fine by me !

ui-core/src/styles/inputs/statusMessage.css
```diff
@@ -31,3 +31,83 @@
...
+  top: calc(
+    100% - var(--input-wrapper-padding) - var(--field-wrapper-padding-bottom) -
+      var(--status-message-wrapper-tooltip--offset)
+  );
+  gap: 12px;
+  border-radius: 8px;
+  box-shadow:
+    0 10px 20px 0 rgba(0, 0, 0, 0.2),
+    0 1px 0 rgba(255, 255, 255, 0.6) inset;
+  padding-inline: 12px;
```
SharglutDev, line 49: Small issue : 

Capture d’écran 2025-01-03 à 15 27 02

In some cases, the text overlaps the close button. There should be 8px padding left of the close button, see mockup : 
Capture d’écran 2025-01-03 à 15 29 17

So padding-right here should be 8 (the gap between the close btn and the text) + 16 (the btn width) + 8 (the btn right offset) > 32px only when the status is `info`

SharglutDev (commented):
Lgtm except one last comment :)

SharglutDev:
> [!NOTE]
> We should keep in mind for your implementation PR what we want to do if the info tooltip is displayed and the user changes the value of the related input

SharglutDev (approved):
Lgtm and tested, great job !

PR #779, opened 2024-12-23
core: refacto combo box

### LINKED ISSUE #778: Refacto combo box
### Description and goal

Refacto the combo box component and solve the following bugs:
- https://github.com/OpenRailAssociation/osrd/issues/8932
- https://github.com/OpenRailAssociation/osrd/issues/9296

### Acceptance criteria

The combo box works correctly and the code is clear

clarani:
closes #778 

Here are the main changes:
- the combo box takes now a value of type T and not a string anymore
- the combo box has a default behavior (filters on the labels) but a custom behaviour can be used
- the hook useOutsideClick replaces the previous handleParentDivBlur function
- a new story has been added to test the custom behavior

### DIFF
--- a/ui-core/src/components/inputs/ComboBox/ComboBox.tsx
+++ b/ui-core/src/components/inputs/ComboBox/ComboBox.tsx
@@ -7,157 +7,100 @@ import React, {
   useMemo,
   useRef,
   useState,
+  useCallback,
 } from 'react';
 
 import { ChevronDown, XCircle } from '@osrd-project/ui-icons';
 import cx from 'classnames';
 
-import { normalizeString } from './utils';
+import useOutsideClick from '../../../hooks/useOutsideClick';
 import Input, { type InputProps } from '../Input';
 
-export type ComboBoxProps<T> = InputProps & {
+export type ComboBoxProps<T> = Omit<InputProps, 'value'> & {
+  value?: T;
   suggestions: Array<T>;
-  getSuggestionLabel: (option: T) => string;
   customLabel?: ReactNode;
   numberOfSuggestionsToShow?: number;
-  exactSearch?: boolean;
-  value?: string;
-  onSelectSuggestion?: (option: T | undefined) => void;
-  disableDefaultFilter?: boolean;
+  getSuggestionLabel: (option: T) => string;
+  onSelectSuggestion: (option: T | undefined) => void;
+  resetSuggestions: () => void;
 };
 
+/**
+ * ComboBox component
+ *
+ * You can use the hook useDefaultComboBox to get the default behavior.
+ * See the stories.
+ */
 const ComboBox = <T,>({
   suggestions,
-  onChange,
-  getSuggestionLabel,
   customLabel,
   numberOfSuggestionsToShow = 5,
-  exactSearch = false,
-  value = '',
+  value,
   small,
+  getSuggestionLabel,
+  onChange,
   onSelectSuggestion,
-  disableDefaultFilter = false,
+  resetSuggestions,
   ...inputProps
 }: ComboBoxProps<T>) => {
-  const [filteredSuggestions, setFilteredSuggestions] = useState<T[]>([]);
   const [activeSuggestionIndex, setActiveSuggestionIndex] = useState(-1);
-  const [inputValue, setInputValue] = useState(value);
-  const [selectedOption, setSelectedOption] = useState<T | null>(null);
+  const [inputValue, setInputValue] = useState('');
   const [isInputFocused, setIsInputFocused] = useState(false);
   const suggestionRefs = useRef<(HTMLLIElement | null)[]>([]);
 
   const inputRef = useRef<HTMLInputElement>(null);
+  const wrapperRef = useRef<HTMLInputElement>(null);
 
-  const sortedSuggestions = useMemo(
-    () =>
-      !disableDefaultFilter
-        ? [...suggestions].sort((a, b) =>
-            getSuggestionLabel(a).localeCompare(getSuggestionLabel(b))
-          )
-        : suggestions,
-    [suggestions, getSuggestionLabel, disableDefaultFilter]
-  );
+  const focusInput = useCallback(() => {
+    inputRef.current?.focus();
+  }, [inputRef]);
 
-  const showSuggestions = isInputFocused && filteredSuggestions.length > 0 && !inputProps.disabled;
-
-  const focusInput = () => inputRef.current?.focus();
-
-  const clearInput = () => {
-    setInputValue('');
-    // Immediately clear the input's value in the DOM to prevent inconsistencies with handleInputFocus
-    if (inputRef.current) {
-      inputRef.current.value = '';
-    }
-    setSelectedOption(null);
-    onSelectSuggestion?.(undefined);
-    const syntheticEvent = {
-      target: {
-        value: '',
-      },
-      currentTarget: {
-        value: '',
-      },
-    } as React.ChangeEvent<HTMLInputElement>;
-
-    // Call the parent's onChange handler with the synthetic event
-    onChange?.(syntheticEvent);
-    focusInput();
-  };
-
-  const icons = [
-    ...(selectedOption || suggestions.some((suggestion) => getSuggestionLabel(suggestion) === value)
-      ? [
-          {
-            icon: <XCircle variant="fill" />,
-            action: clearInput,
-            className: 'clear-icon',
-          },
-        ]
-      : []), // Conditionally include the clear icon only when inputValue is not empty
-    ...(sortedSuggestions.length > 0
-      ? [
-          {
-            icon: <ChevronDown size={small ? 'sm' : 'lg'} />,
-            action: focusInput,
-            className: cx('chevron-icon', {
-              disabled: inputProps.disabled,
-            }),
-          },
-        ]
-      : []), // Conditionally include the chevron icon only when suggestions are not empty
-  ];
-
-  /* eslint-disable react-hooks/exhaustive-deps */
-  useEffect(() => {
-    if (value) {
-      setInputValue(value);
-    } else {
-      setInputValue('');
-      setSelectedOption(null);
-    }
-  }, [value]);
-
-  useEffect(() => {
-    setFilteredSuggestions(sortedSuggestions);
-  }, [sortedSuggestions]);
-
-  const handleInputChange: ChangeEventHandler<HTMLInputElement> = (e) => {
-    const userInput = normalizeString(e.currentTarget.value).trim();
-    setInputValue(e.currentTarget.value);
-    onChange?.(e);
-
-    if (userInput.trim() === '') {
-      setFilteredSuggestions([]);
-      setSelectedOption(null);
-      return;
-    }
-
-    const filtered = sortedSuggestions.filter((suggestion) => {
-      const suggestionLabel = normalizeString(getSuggestionLabel(suggestion).toLowerCase());
-      return exactSearch
-        ? suggestionLabel.startsWith(userInput.toLowerCase())
-        : suggestionLabel.includes(userInput.toLowerCase());
-    });
-    setFilteredSuggestions(filtered);
-  };
-
-  const selectSuggestion = (index: number) => {
-    const selectedSuggestion = filteredSuggestions[index];
-    const suggestionLabel = getSuggestionLabel(selectedSuggestion);
-    setInputValue(suggestionLabel);
-    setSelectedOption(selectedSuggestion);
-    onSelectSuggestion?.(selectedSuggestion);
-    setFilteredSuggestions([]);
+  const removeFocus = () => {
+    setIsInputFocused(false);
     setActiveSuggestionIndex(-1);
     setTimeout(() => {
       inputRef.current?.blur();
     }, 0);
+    resetSuggestions();
+  };
+
+  /* eslint-disable react-hooks/exhaustive-deps */
+  useEffect(() => {
+    if (value) {
+      setInputValue(getSuggestionLabel(value));
+    } else {
+      setInputValue('');
+    }
+  }, [value]);
+
+  const showSuggestions = useMemo(
+    () => isInputFocused && suggestions.length > 0 && !inputProps.disabled,
+    [isInputFocused, suggestions.length, inputProps.disabled]
+  );
+
+  // behavior
+  const handleInputChange: ChangeEventHandler<HTMLInputElement> = (e) => {
+    onChange?.(e);
+    setInputValue(e.currentTarget.value);
+  };
+
+  const selectSuggestion = (index: number) => {
+    const selectedSuggestion = suggestions.at(index)!;
+    onSelectSuggestion(selectedSuggestion);
+    setInputValue(getSuggestionLabel(selectedSuggestion));
+    removeFocus();
+  };
+
+  const closeSuggestions = () => {
+    setInputValue(value ? getSuggestionLabel(value) : '');
+    removeFocus();
   };
 
   const handleKeyDown: KeyboardEventHandler<HTMLInputElement> = (e) => {
     if (e.key === 'ArrowDown') {
       setActiveSuggestionIndex((prev) => {
-        const newIndex = prev < filteredSuggestions.length - 1 ? prev + 1 : prev;
+        const newIndex = prev < suggestions.length - 1 ? prev + 1 : prev;
         if (suggestionRefs.current[newIndex]) {
           (suggestionRefs.current[newIndex] as HTMLLIElement).scrollIntoView({
             block: 'nearest',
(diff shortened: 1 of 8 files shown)


ui-core/src/components/inputs/ComboBox/ComboBox.tsx
```diff
@@ -7,151 +7,139 @@ import React, {
...
-    setInputValue(e.currentTarget.value);
-    onChange?.(e);
+  useEffect(() => {
+    setFilteredSuggestions(suggestions);
+  }, [suggestions]);
 
-    if (userInput.trim() === '') {
-      setFilteredSuggestions([]);
-      setSelectedOption(null);
+  const filterSuggestions = useCallback((input: string) => {
```
emersion: Can't we use `useMemo()` here instead of these `useEffect()` calls?

ui-core/src/components/inputs/ComboBox/ComboBox.tsx
```diff
@@ -7,151 +7,139 @@ import React, {
...
-  getSuggestionLabel: (option: T) => string;
   customLabel?: ReactNode;
   numberOfSuggestionsToShow?: number;
   exactSearch?: boolean;
-  value?: string;
-  onSelectSuggestion?: (option: T | undefined) => void;
   disableDefaultFilter?: boolean;
+  getSuggestionLabel: (option: T) => string;
+  onSelectSuggestion: (option: T | undefined) => void;
+  customOnInputChange?: (inputValue: string) => void;
```
emersion: This feels a bit awkward when there is already an `onChange` prop that users can set if they need to react to text input changes…

It seems like in OSRD we want to pre-filter suggestions before we pass them to the combobox?

Alternative proposals:

- A `filterSuggestions: boolean` prop to enable/disable the combobox built-in filtering.
- A `filterSuggestions: (suggestions: T[], query: string) => T[]` prop to tweak the combobox built-in filtering. Can be set to `(suggestions, query) => suggestions` to disable filtering. Supersedes `exactSearch`.

ui-core/src/components/inputs/ComboBox/ComboBox.tsx
```diff
@@ -7,151 +7,139 @@ import React, {
...
       return exactSearch
-        ? suggestionLabel.startsWith(userInput.toLowerCase())
-        : suggestionLabel.includes(userInput.toLowerCase());
+        ? suggestionLabel.startsWith(input.toLowerCase())
+        : suggestionLabel.includes(input.toLowerCase());
     });
     setFilteredSuggestions(filtered);
+  }, []);
+
+  // behaviors
```
emersion: Nit: "behavior" is uncountable in this context

ui-core/src/components/inputs/ComboBox/ComboBox.tsx
```diff
@@ -7,151 +7,139 @@ import React, {
...
+  };
+
+  const closeSuggestions = () => {
+    setIsInputFocused(false);
+
+    // if only 1 suggestion is left, we select it
+    if (filteredSuggestions.length === 1) {
+      selectSuggestion(0);
+      return;
+    }
```
emersion: That's quite surprising. I'd expect the "Escape" key to cancel everything, instead of selecting the first result. But it seems like the old code was doing that too.

ui-core/src/components/inputs/ComboBox/ComboBox.tsx
clarani: Done :white_check_mark:

ui-core/src/components/inputs/ComboBox/ComboBox.tsx
clarani: Agree, but it was asked by Thibaut I think

ui-core/src/components/inputs/ComboBox/useDefaultComboBox.ts
```diff
@@ -0,0 +1,40 @@
...
+  query: string
+) => {
+  const input = normalizeString(query).trim().toLowerCase();
+  if (!input) {
+    return suggestions;
+  }
+
+  return suggestions.filter((suggestion) => {
+    const suggestionLabel = normalizeString(getSuggestionLabel(suggestion).toLowerCase());
+    return suggestionLabel.startsWith(input.toLowerCase());
```
emersion: Nit: `input` is already lower-case at this point.

ui-core/src/stories/ComboBoxCustom.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+  { id: '4', firstname: 'Manolo', lastname: 'Fernandez' },
+  { id: '5', firstname: 'Maria', lastname: 'Lopez' },
+  { id: '6', firstname: 'Jose', lastname: 'Martinez' },
+  { id: '7', firstname: 'Ana', lastname: 'Sanchez' },
+  { id: '8', firstname: 'Pedro', lastname: 'Perez' },
+  { id: '9', firstname: 'Lucia', lastname: 'Gomez' },
+  { id: '10', firstname: 'Carlos', lastname: 'Martin' },
+  { id: '11', firstname: 'Elena', lastname: 'Jimenez' },
+  { id: '12', firstname: 'Miguel', lastname: 'Ruiz' },
+] as Suggestion[];
```
emersion: I'd prefer to set the type via `const suggestions: Suggestion[] = …` instead so that we get a type error if one of the elements doesn't conform to the type.

ui-core/src/stories/ComboBoxCustom.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  const getSuggestionLabel = (suggestion: Suggestion) =>
+    `${suggestion.firstname} ${suggestion.lastname}`;
+
+  const onSelectSuggestion = (suggestion?: Suggestion) => {
+    setValue(suggestion);
+  };
+
+  const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const inputValue = e.target.value;
```
emersion: Probably we want to `toLowerCase()` this value too?

ui-core/src/components/inputs/ComboBox/ComboBox.tsx
```diff
@@ -7,157 +7,95 @@ import React, {
...
-  getSuggestionLabel: (option: T) => string;
   customLabel?: ReactNode;
   numberOfSuggestionsToShow?: number;
-  exactSearch?: boolean;
-  value?: string;
-  onSelectSuggestion?: (option: T | undefined) => void;
-  disableDefaultFilter?: boolean;
+  getSuggestionLabel: (option: T) => string;
+  onSelectSuggestion: (option: T | undefined) => void;
+  resetSuggestions: () => void;
```
emersion, line 26: Nit: maybe `onClearFilter` would be a slightly better name?

- Makes it clearer that this is supposed to restore all suggestions, as opposed to emptying the suggestion list.
- Makes it clear that this is an event handler, triggered when the user clicks on a button.

ui-core/src/components/inputs/ComboBox/ComboBox.tsx
```diff
@@ -7,157 +7,95 @@ import React, {
...
-
-    const filtered = sortedSuggestions.filter((suggestion) => {
-      const suggestionLabel = normalizeString(getSuggestionLabel(suggestion).toLowerCase());
-      return exactSearch
-        ? suggestionLabel.startsWith(userInput.toLowerCase())
-        : suggestionLabel.includes(userInput.toLowerCase());
-    });
-    setFilteredSuggestions(filtered);
+    const newInputValue = e.currentTarget.value;
+    setInputValue(newInputValue);
```
emersion: Nit: no need for the intermediate variable

ui-core/src/stories/ComboBox.stories.tsx
```diff
@@ -22,24 +23,44 @@ const suggestions = [
...
+const ComboBoxStory = (props: { small?: boolean; disabled?: boolean; readOnly?: boolean }) => {
+  const [value, setValue] = useState<Suggestion>();
+
+  const getSuggestionLabel = (suggestion: Suggestion) => suggestion.label;
+
+  const onSelectSuggestion = (suggestion?: Suggestion) => {
+    setValue(suggestion);
+  };
+
+  const { filteredSuggestions, onChange, resetSuggestions } = useDefaultComboBox(
```
emersion: Nit: no need for destructuring here, we could spread the object returned by `useDefaultComboBox()` in `<ComboBox>` like butter. That way, the code is slightly shorter and any new prop returned by `useDefaultComboBox()` in the future will be passed through without any change here.

ui-core/src/components/inputs/ComboBox/useDefaultComboBox.ts
```diff
@@ -0,0 +1,40 @@
...
+  });
+};
+
+const useDefaultComboBox = <T>(suggestions: T[], getSuggestionLabel: (suggestion: T) => string) => {
+  const [filteredSuggestions, setFilteredSuggestions] = useState<T[]>(suggestions);
```
emersion: If the caller updates the suggestions, `filteredSuggestions` will still be stuck to the old value.

I would recommend to make the current query a state instead of making the filtered suggestion a state:

```ts
const [query, setQuery] = useState<String>('');
const filteredSuggestions = useMemo(() => defaultFilterSuggestions(…), [query, suggestions]);
```

That way, the data flow is clearer as well: the filtered suggestions are _derived_ from the full suggestion list.

emersion (commented):
This is so much better than the old code, thanks a bunch!

Here are a few comments.

ui-core/src/components/inputs/ComboBox/ComboBox.tsx
```diff
@@ -178,57 +116,52 @@ const ComboBox = <T,>({
...
+      // Conditionally include the clear icon only when input is not empty
+      ...(inputValue !== ''
+        ? [
+            {
+              icon: <XCircle variant="fill" />,
+              action: clearInput,
+              className: 'clear-icon',
+            },
+          ]
+        : []),
```
SharglutDev: When I start typing in the field, I see the clear button.
Capture d’écran 2025-01-07 à 14 01 14
In the mockup, this button shows only when a suggestion has been selected.

ui-core/src/components/inputs/ComboBox/useDefaultComboBox.ts
SharglutDev: In the default behavior, we also want to return the suggestions which contains the inputs, but sorted after the "startsWith". See this [issue](https://github.com/OpenRailAssociation/osrd/issues/9356).

ui-core/src/stories/ComboBox.stories.tsx
SharglutDev: This is the default behavior but we will still need to import this hook from osrd-ui each we use the combobox ? Can't it be integrated in the component and overridden when we want a custom behavior ?

ui-core/src/components/inputs/ComboBox/useDefaultComboBox.ts
clarani: Done

ui-core/src/stories/ComboBoxCustom.stories.tsx
clarani: Done

ui-core/src/stories/ComboBoxCustom.stories.tsx
clarani: Done

ui-core/src/components/inputs/ComboBox/ComboBox.tsx
clarani: Done

ui-core/src/stories/ComboBox.stories.tsx
clarani: Done

ui-core/src/components/inputs/ComboBox/useDefaultComboBox.ts
clarani: Hum I'm not sure I can do this, how can I handle the reset function ?
I've added a state value for the query and added a useEffect

emersion: `setQuery('')`?

ui-core/src/components/inputs/ComboBox/ComboBox.tsx
clarani: Done

ui-core/src/components/inputs/ComboBox/useDefaultComboBox.ts
clarani: Done

ui-core/src/stories/ComboBox.stories.tsx
clarani: Hum @emersion ?

I would rather not have the default behavior in the component, but we can re-discuss it.
Previous comment related : https://github.com/OpenRailAssociation/osrd-ui/pull/779#discussion_r1895769205

ui-core/src/components/inputs/ComboBox/ComboBox.tsx
clarani: resetSuggestions is not only called when the user clicks the clear button :/ I'm sure it's a better name

ui-core/src/stories/ComboBox.stories.tsx
emersion: Yeah, I don't see a good way to automatically do the right thing here.

If really invoking the hook is too much boilerplate, we could have a `FilteredComboBox` (or w/e we want to call it) which invokes the hook and forwards everything else to a `ComboBox`.

ui-core/src/components/inputs/ComboBox/useDefaultComboBox.ts
emersion: Gentle ping

ui-core/src/components/inputs/ComboBox/useDefaultComboBox.ts
```diff
@@ -0,0 +1,60 @@
...
+  query: string
+) => {
+  const input = normalizeString(query).trim().toLowerCase();
+  if (!input) {
+    return suggestions;
+  }
+
+  const { startingWithInput, containingInput } = suggestions.reduce<{
+    startingWithInput: T[];
+    containingInput: T[];
```
emersion: Hm, would be nicer IMHO to assign a score to suggestions

```ts
const getSuggestionScore = (suggestion) => {
  const suggestionLabel = normalizeString(getSuggestionLabel(suggestion).toLowerCase());
  if (suggestionLabel.startsWith(input)) {
    return 2;
  }
  if (suggestionLabel.includes(input)) {
    return 1;
  }
  return 0;
};

suggestions
  .map((suggestion) => { suggestion, score: getSuggestionScore(suggestion) })
  .filter(({ score }) => score > 0)
  .sort(({ score: scoreA }, { score: scoreB }) => scoreB - scoreA)
  .map(({ suggestion }) => suggestion)
```

SharglutDev: Chaining 4 loops are better than having only one ? Is it because of the arrays concatenations ? Or it doesn't change much for perf and is just about clarity ?

ui-core/src/stories/ComboBox.stories.tsx
SharglutDev: I'm just afraid that most of dev (except the 3 of us now) won't know about this hook and will always be tempted to create a new custom function. We will need to be vigilant in reviews :)

clarani: Agree !!

ui-core/src/components/inputs/ComboBox/useDefaultComboBox.ts
clarani: Ah yes you're right, sorry I'm already tired !! 😅

ui-core/src/components/inputs/ComboBox/useDefaultComboBox.ts
emersion: I don't think performance matters here, even with long lists. (If it ends up mattering, the above can easily be converted to a `for` loop.) It's just about clarity.

ui-core/src/stories/ComboBox.stories.tsx
emersion: That's a good point. Maybe docs would help? (as a doc-comment above the props referencing the hook)

ui-core/src/components/inputs/ComboBox/ComboBox.tsx
emersion: Alright, I'm not feeling strongly about this anyways.

emersion (approved):
LGTM :)

ui-core/src/stories/ComboBox.stories.tsx
SharglutDev: For the hook and also the `ComboBox` component maybe ? Like "You can use the component with the default custom hook or use a custom one etc"

SharglutDev (approved):
Lgtm and tested, great job !

PR #806, opened 2025-01-07
Clear combobox value when invalid after suggestions update

### LINKED ISSUE #802: Dynamic Suggestions: Selected Value Persists Even When Not in new updated Suggestions
### What happened?

When the suggestions list is updated dynamically, the input value persists even if it no longer exists in the updated suggestions. This results in invalid values being displayed in the input field.

The Combobox should clear or validate the value against the updated suggestions.

- Have a dynamic suggestions updated by the parent
- Have a previously selected input that does't exist in the updated suggestions
- See the value that doesn't exist in the updated suggestions

_No response_

_No response_

achrafmohye:
Closes #802

### DIFF
--- a/ui-core/src/components/inputs/ComboBox/ComboBox.tsx
+++ b/ui-core/src/components/inputs/ComboBox/ComboBox.tsx
@@ -119,6 +119,17 @@ const ComboBox = <T,>({
 
   useEffect(() => {
     setFilteredSuggestions(sortedSuggestions);
+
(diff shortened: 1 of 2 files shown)


emersion:
Might conflict and be superseded by https://github.com/OpenRailAssociation/osrd-ui/pull/779

emersion:
I believe this is superseded by https://github.com/OpenRailAssociation/osrd-ui/pull/779

PR #808, opened 2025-01-08
ui-core: fix time picker component's properties

### LINKED ISSUE #807: core: TimePicker - value attribut
### What happened?

The component `TimePicker` needs the following properties hours, minutes  &  seconds to works (which are the values of the input).

But we can also give it a `value` attribut (due to the property extends React.InputHTMLAttributes) which override the value in the input.

`TimePicker` should not allow `value` property
It should be the same for 'onClick' ,  'onChange' & `type`

This is valid at compilation, but then you can't see the change you do on the time (override by the value)

```tsx
<TimePicker
    hour={13}
    minutes={56}
   value={"10:40"} />
```

_No response_

_No response_

sim51:
fix #807 

This commit fix the allowed properties on the component, to avoid confusion for example between `value` and `hours` & `minutes`.

Also refacto :
- use `useCallback` when necessary
- static function shoud be outside component definition
- `otherProps` at top level, so even if a prop is defined and the component also use it, it will be overriden by the component

### DIFF
--- a/ui-core/src/components/inputs/TimePicker.tsx
+++ b/ui-core/src/components/inputs/TimePicker.tsx
@@ -1,11 +1,14 @@
-import React, { useState, useRef } from 'react';
+import React, { useState, useRef, useCallback } from 'react';
 
 import cx from 'classnames';
 
 import Input, { type InputProps } from './Input';
 import InputModal from '../Modal';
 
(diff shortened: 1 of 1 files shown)


emersion (approved):
LGTM, thanks for the cleanup!

PR #809, opened 2025-01-08
front: fix exported function

Akctarus:


anisometropie (approved):

ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -1,6 +1,6 @@
 import type { InteractiveWaypoint, Waypoint } from '@osrd-project/ui-manchette/dist/types';
 import type { OperationalPoint } from '@osrd-project/ui-spacetimechart/dist/lib/types';
-import { filterVisibleElements } from '@osrd-project/ui-speedspacechart/src/components/utils';
+import { filterVisibleElements } from '@osrd-project/ui-speedspacechart';
```
anisometropie: we should probably avoid creating a dependency to speed space chart altogether here

Akctarus: do you recommend creating the same function for each component?

Yohh: > do you recommend creating the same function for each component?

I agree with that

I think we should have a talk about our chart modules, finding a way to use similar utils or anything else, but in a dedicated refacto, some things to talk later

ui-speedspacechart/src/components/utils.ts
```diff
@@ -333,32 +333,6 @@ export const interpolate = (x1: number, y1: number, x2: number, y2: number, x: n
 export const clamp = (value: number, min: number, max: number) =>
   Math.min(Math.max(value, min), max);
 
-export const filterVisibleElements = <T>({
```
anisometropie: keep the function, it’s still has a purpose to isolate all this inside a separate block and give it a name

just copy the function in a manchette-with-space-time helper file

Akctarus: done

emersion (approved):

PR #813, opened 2025-01-13
ui-core: add narrow input style

### LINKED ISSUE #821: Add narrow input style
### Description and goal

By default in the design system, inputs are wrapped with some padding like this : 
Image

This is to enable the use of required input and prevent some UI reposition : 
Image

Or the use of inputs with status message : 
Image

But in the stdcm form, the mockup requires inputs to be very close from each another, not allowing any space to display a "required" input.
@thibautsailly doesn't plan to add any "required" icon in the form as it's gonna be a form use multiple times each day. The user already knows which fields are required.
Moreover, in stdcm we don't display status message with the standard design system but with a new one using tooltips : 
Image

### Acceptance criteria

- [ ] Add a new way to use inputs for narrow forms where required fields are not used and status message are displayed with tooltips only
- [ ] Required, small, statusWithMessage styles keep working in both mode

SharglutDev:
close #821

### DIFF
--- a/ui-core/src/components/inputs/FieldWrapper.tsx
+++ b/ui-core/src/components/inputs/FieldWrapper.tsx
@@ -15,6 +15,11 @@ export type FieldWrapperProps = {
   disabled?: boolean;
   statusWithMessage?: StatusWithMessage;
   statusIconPosition?: 'next-to-field' | 'before-status-message';
+  /**
+   * Input without any wrapper padding.
+   * Should not be used with required or statusWithMessage with no tooltip.
+   */
+  narrow?: boolean;
   small?: boolean;
   children?: React.ReactNode;
   className?: string;
@@ -29,15 +34,21 @@ const FieldWrapper = ({
   disabled,
   statusWithMessage,
   statusIconPosition = 'next-to-field',
+  narrow = false,
   small = false,
   className,
   children,
   onCloseStatusMessage,
 }: FieldWrapperProps) => {
   const statusClassname = statusWithMessage ? { [statusWithMessage.status]: true } : {};
+  const defaultTooltip: StatusWithMessage['tooltip'] = narrow ? 'left' : undefined;
+
+  if (narrow && required) {
+    throw new Error('narrow should not be used with required for now. This breaks the input UI.');
+  }
 
   return (
-    <div className={cx('feed-back', statusClassname, className, { small: small })}>
+    <div className={cx('feed-back', statusClassname, className, { small, narrow })}>
       <div className="custom-field">
         {/* LABEL AND HINT */}
         {label && (
@@ -68,7 +79,10 @@ const FieldWrapper = ({
         {statusWithMessage && (
           <StatusMessage
             small={small}
-            statusWithMessage={statusWithMessage}
+            statusWithMessage={{
+              ...statusWithMessage,
+              tooltip: statusWithMessage?.tooltip ?? defaultTooltip,
+            }}
             showIcon={statusIconPosition === 'before-status-message'}
             onClose={onCloseStatusMessage}
           />
--- a/ui-core/src/components/inputs/Input.tsx
+++ b/ui-core/src/components/inputs/Input.tsx
@@ -66,6 +66,7 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
       disabled = false,
       readOnly = false,
       statusWithMessage,
+      narrow,
       inputFieldWrapperClassname = '',
       small = false,
       withIcons = [],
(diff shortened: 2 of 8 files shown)


SarahBellaha (approved):
LGTM

ui-core/src/components/inputs/FieldWrapper.tsx
```diff
@@ -15,6 +15,11 @@ export type FieldWrapperProps = {
   disabled?: boolean;
   statusWithMessage?: StatusWithMessage;
   statusIconPosition?: 'next-to-field' | 'before-status-message';
+  /**
+   * Input without any wrapper padding.
+   * Should not be used with required or statusWithMessage with no tooltip.
+   */
+  narrow?: boolean;
```
SarahBellaha, line 22: Maybe too much, but we could probably have a type that would prevent having “narrow” and “required” at the same time

ui-core/src/components/inputs/FieldWrapper.tsx
```diff
@@ -15,6 +15,11 @@ export type FieldWrapperProps = {
   disabled?: boolean;
   statusWithMessage?: StatusWithMessage;
   statusIconPosition?: 'next-to-field' | 'before-status-message';
+  /**
+   * Input without any wrapper padding.
+   * Should not be used with required or statusWithMessage with no tooltip.
```
emersion, line 20: Since `narrow` without tooltip results in a broken input message, should we forbid or prevent this from happening? One way would be to add a default value for `tooltip` for `narrow` inputs, e.g.

```tsx
const defaultTooltip = narrow ? 'left' : undefined;

statusWithMessage={({
  ...statusWithMessage,
  tooltip: statusWithMessage.tooltip ?? defaultTooltip,
})}
```

In the future, when we switch to the new anchor API, we will be able to get rid of left/right and only expose a boolean to enable/disable the tooltip.

ui-core/src/components/inputs/FieldWrapper.tsx
SharglutDev: You're totally right, I tried for a long time to find a good way to do this and ended up with a bunch of type errors in some components where `Input` is used. But I'll give it another try :)

ui-core/src/components/inputs/FieldWrapper.tsx
SharglutDev: Done

emersion (approved):

SharglutDev:
@SarahBellaha @emersion, I added another commit for Sarah's request to prevent having `required` and `narrow` defined at the same time. You can test it by trying to set both props at true in any component calling `Input` component. Type is a bit messy but it works. What do you think ? Do we keep it or no ?

ui-core/src/components/inputs/ComboBox/ComboBox.tsx
```diff
@@ -16,7 +16,13 @@ import cx from 'classnames';
...
 import Input, { type InputProps } from '../Input';
 
-export type ComboBoxProps<T> = Omit<InputProps, 'value'> & {
+export type ComboBoxProps<T> = (
+  | (Omit<InputProps, 'value' | 'children' | 'required'> & {
+      required: boolean;
+      narrow?: undefined;
+    })
+  | (Omit<InputProps, 'value' | 'children' | 'narrow'> & { narrow: boolean; required?: undefined })
+) & {
```
emersion: Why do we need to change this type? I think `InputProps` already has the correct `required` and `narrow` fields?

ui-core/src/components/inputs/Input.tsx
```diff
@@ -46,7 +46,10 @@ type IconConfig = {
 };
 
 export type InputProps = React.InputHTMLAttributes<HTMLInputElement> &
-  Omit<FieldWrapperProps, 'children'> & {
+  (
+    | ({ narrow: boolean; required?: undefined } & Omit<FieldWrapperProps, 'children' | 'required'>)
+    | ({ required: boolean; narrow?: undefined } & Omit<FieldWrapperProps, 'children' | 'narrow'>)
+  ) & {
```
emersion: Same here, I think we can just leave this type unchanged and it should work?

ui-core/src/components/inputs/FieldWrapper.tsx
```diff
@@ -11,15 +11,27 @@ export type FieldWrapperProps = {
...
+   */
+} & (
+  | {
+      required: boolean;
+      narrow: undefined;
+    }
+  | {
+      narrow: boolean;
+      required: undefined;
+    }
```
emersion: This forbids things such as:

```ts
{ required: false, narrow: true }
{ narrow: false, required: false }
```

I think these should be allowed?

Proposal:

```ts
  | {
      required: true;
      narrow?: false;
    }
  | {
      narrow: true;
      required?: false;
    }
  | {
      narrow?: false,
      required?: false,
    }
```

SharglutDev:
After discussions, removed the commit for type safety with required/narrow. Instead, throw an error if both are true at the same time.

ui-core/src/components/inputs/FieldWrapper.tsx
emersion: We tried again, and it got pretty messy so we'd prefer to keep it simple for now :P

emersion (approved):

SarahBellaha (approved):

PR #823, opened 2025-01-14
Fix unresolved dependencies 

emersion:
_See individual commits._

We don't want to include our dependencies (react, maplibre, and so
on) in the osrd-ui library bundle: because library users download these as
transitive deps we want to leave it up to them to resolve the
imports and include them in the final app bundle. For instance,
OSRD is responsible for inlining maplibre imports in ui-warped-map's
bundle.

On the other hand, relative imports local to the package should be
bundled.

By default, rollup only resolves relative imports, and doesn't
resolve global imports. rollup prints an error when it encounters
a global import that it can't resolve. We don't want to just
ignore unresolved global imports, because that would silently
allow broken non-existing imports.

The @rollup/plugin-node-resolve plugin allows rollup to resolve
global imports like node does. However, by default it includes
all global imports in the bundle. This isn't what we want: we
want to exclude dependencies from the bundle.

To fix this, follow the recommendations in the [rollup docs][1]
and use the "external" option to check whether an import points
to the "node_modules" directory or a "dist" directory of any
osrd-ui package. In that case, ask rollup to not bundle the
import.

We don't need to manually specify a list of external imports
from each package anymore.

[1]: https://rollupjs.org/configuration-options/#external

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,6 +20,7 @@
       ],
       "devDependencies": {
         "@rollup/plugin-eslint": "^9.0.5",
+        "@rollup/plugin-node-resolve": "^16.0.0",
         "@rollup/plugin-terser": "^0.4.4",
         "@rollup/plugin-typescript": "^12.1.0",
         "@types/lodash": "^4.17.13",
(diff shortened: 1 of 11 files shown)


Yohh (approved):
nice, thanks

SharglutDev (approved):
Thanks !

PR #824, opened 2025-01-14
ui-core: fix input status info close button

SharglutDev:
In tooltip status, a close button is present to close the tooltip but it shouldn't be there if the status isn't a tooltip.

Before
Capture d’écran 2025-01-14 à 14 20 17

After
Capture d’écran 2025-01-14 à 14 25 07


kmer2016 (approved):
LGTM ! Thank you for the fix

PR #826, opened 2025-01-14
Fix modal position to stick to input for pickers

achrafmohye:
Related to https://github.com/OpenRailAssociation/osrd/issues/10378


clarani (commented):
I tried to test in osrd but the modal appears at a weird place

image
image

achrafmohye:
@clarani added a new story where the height is scrollable it might be handy for easier testing regardless of the Osrd integration i think there is a problem here too, will test the integration part :)

clarani (approved):
LGTM :white_check_mark:

PR #832, opened 2025-01-20
build: build ui-icons before ui-core

emersion:
ui-core depends on ui-icons. Ensure it gets build before to fix unresolved module errors.

PR #833, opened 2025-01-20
build: use rollup --failAfterWarnings

emersion:
We've got to a point where we've eliminated all build errors. We
previously did not notice these because they only show up as
warnings in CI, without making CI fail.

Use rollup --failAfterWarnings so that CI fails when a build error
is emitted.

Depends on #832

### DIFF
--- a/ui-spacetimechart/package.json
+++ b/ui-spacetimechart/package.json
@@ -27,9 +27,8 @@
     }
   },
   "scripts": {
-    "rollup": "rollup -c",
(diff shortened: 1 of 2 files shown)


emersion:
This is blocked by a cyclic dependency between ui-spacetimechart and ui-trackoccupancydiagram.

emersion:
Now that all charts packages have been consolidated in ui-charts, `--failAfterWarnings` is enabled everywhere.

PR #834, opened 2025-01-20
fix partially hidden stories + typo

### LINKED ISSUE #559: Speed-Space chart is partially hidden in story book + awkward blacks margins
### What happened?

https://github.com/user-attachments/assets/6ba725be-7632-4f06-8bf6-d80a27e664de

I can't see the left part, it's eaten by the box. I think this weird behavior is reflected in OSRD: when zooming in or out, the component's width doesn't change
Also, it feels like the black borders shouldn't be there

I expect the component to take up the width of the box and not flow out.

1. Go to https://openrailassociation.github.io/osrd-ui/?path=/docs/speedspacechart-rendering--docs
2. Witness

Deployed Storybook

Firefox

Synar:
Close #559

### DIFF
--- a/storybook/.storybook/main.ts
+++ b/storybook/.storybook/main.ts
@@ -5,7 +5,7 @@ import { mergeConfig } from 'vite';
 /** @type { import('@storybook/html-vite').StorybookConfig } */
 const config: StorybookConfig = {
   stories: [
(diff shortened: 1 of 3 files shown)


Caracol3 (approved):
lgtm !

Math-R (approved):
Lgtm

PR #836, opened 2025-01-20
ui-manchette: always display arrival operational point in manchette

clarani:
related to https://github.com/OpenRailAssociation/osrd/issues/9538

### DIFF
--- a/ui-manchette-with-spacetimechart/src/helpers.ts
+++ b/ui-manchette-with-spacetimechart/src/helpers.ts
@@ -26,8 +26,15 @@ export const filterVisibleElements = <T>({
   getWeight,
   minSpace,
 }: VisibilityFilterOptions<T>): T[] => {
+  const firstElement = elements.at(0);
+  const lastElement = elements.at(-1);
+  if (!firstElement || !lastElement) return elements;
+
   const sortedElements = [...elements].sort((a, b) => (getWeight(b) ?? 0) - (getWeight(a) ?? 0));
-  const displayedElements: { element: T; position: number }[] = [];
+  const displayedElements: { element: T; position: number }[] = [
+    { element: firstElement, position: getPosition(firstElement) },
+    { element: lastElement, position: getPosition(lastElement) },
+  ];
 
   for (const element of sortedElements) {
     const position = getPosition(element);


ui-manchette-with-spacetimechart/src/helpers.ts
```diff
@@ -26,8 +26,15 @@ export const filterVisibleElements = <T>({
   getWeight,
   minSpace,
 }: VisibilityFilterOptions<T>): T[] => {
+  const firstElement = elements.at(0);
+  const lastElement = elements.at(-1);
+  if (!firstElement || !lastElement) return elements;
```
SharglutDev, line 31: What's the point of continuing the function if `elements` is an empty array ?
Also where `filterVisibleElements` is called, there is a `waypoints.length === 0` check so we know they will be defined.

ui-manchette-with-spacetimechart/src/helpers.ts
SharglutDev, line 1: nit : not from your PR but `filterVisibleElements` is exported but not used outside of this file, we could remove the `export`

Akctarus (approved):
tested, works well in infras without weighted OP

ui-manchette-with-spacetimechart/src/helpers.ts
clarani: > What's the point of continuing the function if elements is an empty array ?

If it's an empty array, the function will stop at the test l31 :) 

> Also where filterVisibleElements is called, there is a waypoints.length === 0 check so we know they will be defined.

Hum this does not handle the case where waypoints contain only 1 element

ui-manchette-with-spacetimechart/src/helpers.ts
clarani: No I have to keep it, because I have some tests :)

SharglutDev: The tests are using `filterVisibleElements` from `speedspacechart` repo, not `manchettewith...`, but it's not very important :)

ui-manchette-with-spacetimechart/src/helpers.ts
SharglutDev: The manchette can't exist if there is only one element in it, but that's also not very important :)

SharglutDev (approved):
Lgtm and tested

PR #837, opened 2025-01-20
all: stop importing CSS files from library code

emersion:
We were importing CSS files directly from library code. This causes issues for downstream users, because they are forced to configure their build system to support CSS imports (these aren't standard JS, e.g. webpack doesn't support this by default). Moreover, this makes it impossible to use the library without a build system (e.g. importing it directly from pure HTML/JS files).

Make users responsible for including the CSS one way or another. Our stories do this using CSS imports already, but other downstream consumers might prefer to use a `<link>` or similar.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18601,6 +18601,10 @@
     },
     "storybook": {
       "name": "@osrd-project/storybook",
+      "dependencies": {
+        "@osrd-project/ui-core": "0.0.1-dev",
+        "@osrd-project/ui-trackoccupancydiagram": "0.0.1-dev"
+      },
       "devDependencies": {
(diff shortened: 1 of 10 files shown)


emersion:
Hm so… Turns out this is trickier than anticipated.

`dist/theme.css` gets generated by rollup-plugin-postcss, as a side effect of importing a CSS file from JS. Dropping the imports makes it so rollup-plugin-postcss no longer collects any CSS file as input.

However, the issue isn't the `./styles/main.css` import that we have: that one gets stripped away from the final JS. The issue is the `@osrd-project/ui-core/dist/theme.css` import which doesn't get stripped.

I've found a better solution to this problem: https://github.com/OpenRailAssociation/osrd-ui/pull/852

PR #838, opened 2025-01-20
fix path label positioning in spacetimechart

### LINKED ISSUE #468: Train name isn't aligned with subpieces of its path
### What happened?

image
The train name isn't centered on the path that's visible. If there were some pieces of another train next to this one, one wouldn't be able to tell them apart

I expect the name to not be completely out of the piece of path. For example, centered on the piece if it's too small

1. View a time table with many trains, some partially projected on the current path

_No response_

Firefox

### LINKED ISSUE #529: Space-time chart: train name doesn't stick with the curve when scrolling the space axis
### What happened?

https://github.com/user-attachments/assets/3f59e468-9717-49ec-904c-7409c26ee65e

When scrolling along the time axis, the train name follows the curve. When scrolling the space axis, we lose the name of the train

I expect the name of the train to stay visible as long as the curve is in the window.

1. Go to the deployed storybook
2. Zoom on the space axis, so there's room to scroll
3. Scroll on space axis
4. Witness

Deployed Storybook

Firefox

Synar:
Close #468
Close #529

### DIFF
--- a/ui-spacetimechart/src/components/PathLayer.tsx
+++ b/ui-spacetimechart/src/components/PathLayer.tsx
@@ -225,19 +225,22 @@ export const PathLayer = ({
       if (firstPointOnScreenIndex === 0) {
         if (next) angle = Math.atan2(next.y - curr.y, next.x - curr.x);
       } else {
-        if (!swapAxis) {
+        const slope = (curr.y - prev.y) / (curr.x - prev.x);
+        const yOnYAxisIntersect = curr.y - curr.x * slope;
+        const xOnXAxisIntersect = curr.x - curr.y / slope;
+        if (yOnYAxisIntersect >= 0) {
           position = {
             x: 0,
-            y: curr.y - (curr.x * (curr.y - prev.y)) / (curr.x - prev.x),
+            y: yOnYAxisIntersect,
           };
         } else {
           position = {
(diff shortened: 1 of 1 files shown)


ui-spacetimechart/src/components/PathLayer.tsx
```diff
@@ -338,9 +369,21 @@ export const PathLayer = ({
       drawExtremities(ctx, stcContext);
 
       // Draw label:
-      if (!stcContext.hidePathsLabels) drawLabel(ctx, stcContext, path.label, color, segments);
+      if (!stcContext.hidePathsLabels) {
+        const pathLenght = computePathLength(stcContext.operationalPoints, segments);
```
clarani: Typo `pathLenght` => `pathLength` :wink:

clarani (approved):
LGTM, can you just rename your commits ? You're fixing the spacetimechart, and not the speedspacechart ;)

ui-spacetimechart/src/components/PathLayer.tsx
Synar: Thanks!

Synar:
Oops, will fix! '^^

kmer2016 (approved):

PR #848, opened 2025-01-21
ci: upgrade NodeJS to version 23

emersion:
We don't need anything new from this release, but it's better to keep up with new versions to get bugfixes (older NodeJS versions get less support) and a heads-up for new deprecations.


Khoyo (approved):

emersion:
The second commit can be dropped if https://github.com/OpenRailAssociation/osrd-ui/pull/849 is merged.

PR #853, opened 2025-01-22
build: type-check package stories from storybook

emersion:


emersion:
So… It turns out making this work is quite a pain. It's a lot of juggling to exclude the right subdirs from the right TS and ESLint configs. Right now it's just infinite looping.

Let's just move all stories to `storybook/` instead.

PR #858, opened 2025-01-27
ui-spacetimechart: fix blank stories

emersion:
The space time chart stories were sometimes not loading properly, showing a blank page.

Two issues:

- All stories were passing className='absolute inset-0' to SpaceTimeChart. However, SpaceTimeChart was combining these classes with 'relative'. As a result, the browser would roll a dice and pick either 'absolute' or 'relative'. When 'relative' is picked first, 'inset-0' has no effect because…
- All stories except one were using a wrapper div with className="inset-0". This had no effect because this div is neither absolute nor relative.

As a result, half the time the chart would have a zero height.

Fix this by:

- Adding absolute positioning to the wrapper div.
- Setting the height of the SpaceTimeChart instead of mixing up conflicting positioning classes.

### DIFF
--- a/ui-spacetimechart/src/stories/additional-data.stories.tsx
+++ b/ui-spacetimechart/src/stories/additional-data.stories.tsx
@@ -188,9 +188,9 @@ const Wrapper = ({ swapAxis, spaceScaleType }: WrapperProps) => {
   });
 
   return (
-    <div className="inset-0">
+    <div className="absolute inset-0">
       <SpaceTimeChart
-        className={cx('inset-0 absolute p-0 m-0', state.panning && 'cursor-grabbing')}
+        className={cx('h-full p-0 m-0', state.panning && 'cursor-grabbing')}
         operationalPoints={OPERATIONAL_POINTS}
         swapAxis={swapAxis}
         spaceOrigin={0}
--- a/ui-spacetimechart/src/stories/base-rendering.stories.tsx
+++ b/ui-spacetimechart/src/stories/base-rendering.stories.tsx
@@ -50,20 +50,22 @@ const Wrapper = ({
     : PATHS;
 
   return (
-    <SpaceTimeChart
-      className="inset-0 absolute"
-      operationalPoints={operationalPoints}
-      spaceOrigin={0}
-      spaceScales={spaceScales}
-      timeOrigin={+new Date('2024/04/02')}
-      timeScale={60000 / xZoomLevel}
-      xOffset={xOffset}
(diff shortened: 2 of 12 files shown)


ui-spacetimechart/src/stories/additional-data.stories.tsx
```diff
@@ -188,9 +188,9 @@ const Wrapper = ({ swapAxis, spaceScaleType }: WrapperProps) => {
   });
 
   return (
-    <div className="inset-0">
+    <div className="absolute inset-0">
       <SpaceTimeChart
-        className={cx('inset-0 absolute p-0 m-0', state.panning && 'cursor-grabbing')}
+        className={cx('h-full p-0 m-0', state.panning && 'cursor-grabbing')}
```
Synar, line 193: Since we add h-full in every single story (and we also do in osrd), should we move it to the base class of the component?

Synar (approved):
Overall LGTM. This bug was extremely annoying when working on the component, so thanks a lot!

Tangent to this PR, but in the osrd repo we also use <className="inset-0 absolute h-full">, should we remove the absolute and inset-0?
More generally there seem to be a lot of conflicting positioning class uses in this component and its wrappers, which could use a cleanup.

emersion:
> Tangent to this PR, but in the osrd repo we also use <className="inset-0 absolute h-full">, should we remove the absolute and inset-0?

Yes, I think so. We specify a fixed `height` attribute so all of these classes can be dropped.

ui-spacetimechart/src/stories/additional-data.stories.tsx
emersion: I don't think it's right to use it in OSRD, since we don't want the chart to take up the full height. (All stories want the chart to take the full height.)

anisometropie (approved):
tested, thanks!

clarani (approved):
LGTM :white_check_mark:

ui-spacetimechart/src/stories/additional-data.stories.tsx
clarani: But I think it would be nice to create a `BaseStory` utils, which could be used in every story. We could move the h-full classname in this component if we create it :)

emersion: Not a fan of that, because this would make it more difficult for users to figure out how to use the component. They'd need to look at more files to see how the components are rendered.

(Resolving because this is orthogonal to this PR, and can be discussed separately.)

PR #870, opened 2025-02-03
ui-spacetimechart: zoom rectangle [1/3]

### LINKED ISSUE #10569: [storybook] draw rectangle from 2 points on a canvas
Be able to draw rectangle from 2 points using the mouse with reasonable behavior.
 
[main ticket](https://github.com/osrd-project/osrd-confidential/issues/528)

anisometropie:
1st step to implement rectangle zoom on spacetime, only showcasing the behavior drawing with the mouse.

- SimpleRect component
- space time chart story Draw Rectangle

https://www.sketch.com/s/bcbac4ff-7f74-4f1a-9e10-08d115c47ed8/a/nQorEMz

closes https://github.com/OpenRailAssociation/osrd/issues/10569

### DIFF
--- a/ui-charts/src/spaceTimeChart/components/PatternRect.tsx
+++ b/ui-charts/src/spaceTimeChart/components/PatternRect.tsx
@@ -2,55 +2,27 @@ import { useCallback } from 'react';
 
 import { useDraw } from '../hooks/useCanvas';
 import { type DrawingFunction } from '../lib/types';
+import { fillRect, type CanvasRect } from '../utils/canvas';
 
-export type PatternRectProps = {
-  timeStart: Date;
-  timeEnd: Date;
-  spaceStart: number; // mm
-  spaceEnd: number; // mm
+export type PatternRectProps = CanvasRect & {
   imageElement: HTMLImageElement;
 };
 
 /**
  * draws a repeating pattern in the space time chart
  */
-export const PatternRect = ({
-  timeStart,
-  timeEnd,
-  spaceStart,
-  spaceEnd,
-  imageElement,
-}: PatternRectProps) => {
-  const drawRegion = useCallback<DrawingFunction>(
-    (ctx, { getSpacePixel, getTimePixel, spaceAxis }) => {
-      const timeStartPixel = getTimePixel(Number(timeStart));
-      const endTimePixel = getTimePixel(Number(timeEnd));
-      const spaceStartPixel = getSpacePixel(spaceStart);
-      const spaceEndPixel = getSpacePixel(spaceEnd);
-
-      const areaSpaceSize = spaceEndPixel - spaceStartPixel;
-      const areaTimeSize = endTimePixel - timeStartPixel;
-      if (!areaSpaceSize || !areaTimeSize) return;
-
+export const PatternRect = ({ imageElement, ...rect }: PatternRectProps) => {
+  const drawPatternRect = useCallback<DrawingFunction>(
+    (ctx, context) => {
       const pattern = ctx.createPattern(imageElement, 'repeat');
-      if (!pattern) {
-        return;
+      if (pattern) {
+        ctx.fillStyle = pattern;
+        fillRect(ctx, rect, context);
       }
-
-      ctx.save();
-      ctx.fillStyle = pattern;
-      if (spaceAxis === 'x') {
-        ctx.translate(spaceStartPixel, timeStartPixel);
-        ctx.fillRect(0, 0, areaSpaceSize, areaTimeSize);
-      } else {
-        ctx.translate(timeStartPixel, spaceStartPixel);
-        ctx.fillRect(0, 0, areaTimeSize, areaSpaceSize);
-      }
-      ctx.restore();
     },
-    [timeStart, timeEnd, spaceStart, spaceEnd, imageElement]
+    [imageElement, rect]
   );
-  useDraw('background', drawRegion);
+  useDraw('background', drawPatternRect);
 
   return null;
 };
--- a/ui-charts/src/spaceTimeChart/components/ZoomRect.tsx
+++ b/ui-charts/src/spaceTimeChart/components/ZoomRect.tsx
@@ -1,0 +1,53 @@
+import { useCallback } from 'react';
+
+import { useDraw } from '../hooks/useCanvas';
+import { type DrawingFunction } from '../lib/types';
+import { fillRect, type CanvasRect } from '../utils/canvas';
+
+/**
+ * radius 1 black dot with radius 3 white region around it
+ */
+function squareDot(ctx: CanvasRenderingContext2D, cx: number, cy: number) {
+  ctx.save();
+  ctx.fillStyle = 'white';
+  ctx.fillRect(cx - 1, cy - 1, 3, 3);
+  ctx.fill();
+
+  ctx.fillStyle = 'black';
+  ctx.fillRect(cx, cy, 1, 1);
+  ctx.fill();
+  ctx.restore();
+}
+
+const LINE_WIDTH = 1;
+const SPACING = 4;
+
+export const ZoomRect = (rect: CanvasRect) => {
+  const drawZoomRect = useCallback<DrawingFunction>(
+    (ctx, context) => {
+      ctx.save();
+      ctx.fillStyle = '#0000000D'; /* black5 */
+      const { areaTimeSize, areaSpaceSize } = fillRect(ctx, rect, context);
+      if (areaTimeSize && areaSpaceSize) {
+        const width = context.timeAxis === 'x' ? areaTimeSize : areaSpaceSize;
+        const height = context.timeAxis === 'x' ? areaSpaceSize : areaTimeSize;
+
+        ctx.lineWidth = LINE_WIDTH;
+
+        for (let i = 0; Math.abs(i) < Math.abs(width); i += SPACING * Math.sign(width)) {
+          squareDot(ctx, i, 0);
+          squareDot(ctx, i, 0 + height);
+        }
+        for (let i = 0; Math.abs(i) < Math.abs(height); i += SPACING * Math.sign(height)) {
+          squareDot(ctx, 0, i);
+          squareDot(ctx, 0 + width, i);
+        }
+      }
+      ctx.restore();
+    },
+    [rect]
+  );
+  useDraw('background', drawZoomRect);
+
+  return null;
+};
--- a/ui-charts/src/spaceTimeChart/stories/draw-rectangle.stories.tsx
+++ b/ui-charts/src/spaceTimeChart/stories/draw-rectangle.stories.tsx
@@ -1,0 +1,211 @@
+import React, { useEffect, useMemo, useState } from 'react';
+
+import { Button } from '@osrd-project/ui-core';
+import type { Meta } from '@storybook/react';
+import { clamp } from 'lodash';
+
+import '@osrd-project/ui-core/dist/theme.css';
+
+import { OPERATIONAL_POINTS, PATHS } from './lib/paths';
+import { PathLayer } from '../components/PathLayer';
+import { SpaceTimeChart } from '../components/SpaceTimeChart';
+import { ZoomRect } from '../components/ZoomRect';
+import { type Point, type PathData, type OperationalPoint } from '../lib/types';
+import { getDiff } from '../utils/vectors';
+import { MouseTracker } from './lib/components';
+
+const DEFAULT_WIDTH = 900;
+const DEFAULT_HEIGHT = 450;
+
+const MIN_ZOOM = 0;
+const MAX_ZOOM = 100;
+const MIN_ZOOM_MS_PER_PX = 600000;
+const MAX_ZOOM_MS_PER_PX = 625;
+const DEFAULT_ZOOM_MS_PER_PX = 7500;
+type SpaceTimeHorizontalZoomWrapperProps = {
+  swapAxes: boolean;
+  keepDrawing: boolean;
+  spaceOrigin: number;
+  xOffset: number;
+  yOffset: number;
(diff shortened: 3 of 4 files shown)


ui-spacetimechart/src/styles/main.css
```diff
@@ -46,3 +50,100 @@
...
+  border: none; /* Remove default border */
+}
+  
+.simple-rect-border-box::after {
+  content: "";
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
```
RomainValls: Would using 'inset : 0' work here ?

ui-spacetimechart/src/styles/main.css
```diff
@@ -46,3 +50,100 @@
...
+    bottom 1px left 1px / 100% 1px no-repeat,
+    /* LEFT BORDER */
+    repeating-linear-gradient(
+      0deg,
+      theme('colors.primary.90') 0px,
+      theme('colors.primary.90') 1px,
+      transparent 1px,
+      transparent 4px
+    )
+    left 1px bottom 1px / 1px 100% no-repeat;
```
RomainValls: I tried using : border: 1px dashed theme('colors.primary.90'); instead of all this. Seems quite similar, i don't know if we could use it instead ?

RomainValls: If yes, we could do something similar with the ::before

RomainValls (commented):
Tested, nice work ! Left some comments, not sure they can be applied.

ui-spacetimechart/src/styles/main.css
anisometropie: `1px dashed` or `1px dotted` don’t exactly match what’s asked in the mockup. we can’t adjust the space between two dots.

`border-image` and canvas solutions I tried weren’t satisfying either.

https://www.sketch.com/s/bcbac4ff-7f74-4f1a-9e10-08d115c47ed8/a/nQorEMz

ui-spacetimechart/src/styles/main.css
anisometropie: thanks, didn’t know that shortcut

ui-spacetimechart/src/components/SimpleRect.tsx
```diff
@@ -0,0 +1,93 @@
...
+          overlayRef.current.style.width = `${borderTimeSize}px`;
+          overlayRef.current.style.height = `${borderSpaceSize}px`;
+        }
+      }
+    },
+    [timeStart, timeEnd, spaceStart, spaceEnd, fillStyle, border]
+  );
+  useDraw('background', drawRegion);
+
+  return <div ref={overlayRef} className="simple-rect-border-box" />;
```
emersion: Any reason why half of the rectangle (the background) is rendered with a canvas, and the other half (the border) is rendered with CSS? I think this is surprising in terms of z-position (the border will always be on top of everything else) and in terms of performance (the div will not behave well when instanced many times).

ui-spacetimechart/src/components/SimpleRect.tsx
```diff
@@ -0,0 +1,93 @@
...
+export type SimpleRectProps = {
+  timeStart: Date;
+  timeEnd: Date;
+  spaceStart: number; // mm
+  spaceEnd: number; // mm
+  fillStyle: FillStyle;
+  border?: boolean;
+};
+
+export const SimpleRect = ({
```
emersion: I'm not a fan of the term "SimpleRect": I'm not sure it conveys what this component does very well. Plus, the border style is hardcoded - this is surprising for a low-level component like this.

But do we need to have the border and the rect as a single component? This mixes up a bit two separate things IMHO. Maybe we could have a separate component for just the border, applied after the pattern rect?

Taking another step back, do we want to be using a generic rect component for the zoom rectangle? The actual functionality that we're interesting in here is just `translate()` and `fillRect()` - I'd prefer to duplicate that small bit instead of making the existing component more generic. Since the zoom rectangle has a particular style, could we have a component dedicated to drawing that style - just like we have `ConflictLayer`, `OccupancyBlockLayer`, `WorkScheduleLayer`, and so on?

ui-spacetimechart/src/stories/draw-rectangle.stories.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+import { Button } from '@osrd-project/ui-core';
+import type { Meta } from '@storybook/react';
+import { clamp } from 'lodash';
+
+import '@osrd-project/ui-core/dist/theme.css';
+
+import { OPERATIONAL_POINTS, PATHS } from './lib/paths';
+import { PathLayer } from '../components/PathLayer';
+import { SimpleRect } from '../components/SimpleRect';
```
emersion: This story imports some low-level drawing helpers to render the zoom rectangle. What's the long term plan once the feature is complete: do we expect users to implement the actual zoom functionality themselves, or will we expose helpers to handle zoom rectangle interactions?

anisometropie: Once we have a story to showcase the zoom behavior using a rectangle (PR [2/3]) 

I’m planning to have a wrapper like `MouseTracker` that you just put inside your space time chart, and probably some helper hook/function to handle the logic behind

ui-spacetimechart/src/components/SimpleRect.tsx
anisometropie: I couldn’t get any satisfying rendering with the canvas, it’s very blurry.

ui-spacetimechart/src/components/SimpleRect.tsx
anisometropie: just “Rectangle” ? or `Rect` ?

Yes, the border probably could go in a separate component.

I don’t really see the point to duplicate, it’s not just translate and fillRect, we have all the coordinates and dimensions to compute.

component composition seems just fine here
`SimpleRect` takes a `fillStyle` and draws it in a given 2d region.
`PatternRect` is a particular case of `SimpleRect` where the `fillStyle` is a canvas pattern.
`BorderRect` could be SimpleRect + border prop

ui-spacetimechart/src/components/SimpleRect.tsx
emersion: One needs to be careful wrt. pixel grid alignment when rendering lines on the canvas. See e.g. https://github.com/OpenRailAssociation/osrd-ui/pull/914

But I really don't think using mixed canvas and HTML is a good idea here.

anisometropie: I retried canvas, seems much better even without any trying to align pixels.

RomainValls (approved):
There seems to be some unused variables left but except for that LGTM 👍

ui-spacetimechart/src/components/SimpleRect.tsx
emersion: Nice!

ui-spacetimechart/src/components/CanvasRect.tsx
```diff
@@ -0,0 +1,75 @@
...
+      if (timeAxis === 'x') {
+        ctx.translate(timeStartPixel, spaceStartPixel);
+        ctx.fillRect(0, 0, areaTimeSize, areaSpaceSize);
+      } else {
+        ctx.translate(spaceStartPixel, timeStartPixel);
+        ctx.fillRect(0, 0, areaSpaceSize, areaTimeSize);
+      }
+
+      if (drawBorder) {
+        drawBorder(ctx, timeAxis, areaTimeSize, areaSpaceSize);
```
emersion: Do we really need this call to be inside this component? Can't the caller do it themselves?

ui-spacetimechart/src/components/SimpleRect.tsx
emersion: Still not a fan of the `CanvasRect`/`PatternRect` split TBH. This sounds like introducing abstractions for little gain.

ui-spacetimechart/src/stories/draw-rectangle.stories.tsx
emersion: Hm, okay, I suppose we can keep things as-is for now and decide later!

ui-spacetimechart/src/components/SimpleRect.tsx
anisometropie: Hm, I think it’s a reasonable tradeoff. 
- we avoid repeating too much code. all this stuff about computing x, y and size + translate is use in three cases, rectangle, rectangle with border, pattern rectangle.
```
      const timeStartPixel = getTimePixel(Number(timeStart));
      const endTimePixel = getTimePixel(Number(timeEnd));
      const spaceStartPixel = getSpacePixel(spaceStart);
      const spaceEndPixel = getSpacePixel(spaceEnd);

      const areaSpaceSize = spaceEndPixel - spaceStartPixel;
      const areaTimeSize = endTimePixel - timeStartPixel;
      if (!areaSpaceSize || !areaTimeSize) return;
///
      if (timeAxis === 'x') {
        ctx.translate(timeStartPixel, spaceStartPixel);
        ctx.fillRect(0, 0, areaTimeSize, areaSpaceSize);
      } else {
        ctx.translate(spaceStartPixel, timeStartPixel);
        ctx.fillRect(0, 0, areaSpaceSize, areaTimeSize);
      }
```

- We avoid having a bloated component doing rectangle + border + pattern, concern which you brought up initially
> But do we need to have the border and the rect as a single component? This mixes up a bit two separate things IMHO. Maybe we could have a separate component for just the border, applied after the pattern rect?

each component does very little and is easy to reuse.

ui-spacetimechart/src/components/CanvasRect.tsx
anisometropie: Here the wrapper `DottedBorderRect` doesn’t have `ctx` in its scope, so we pass `drawBorder` as props, which receives `ctx`

Otherwise we could have another `useDraw` call inside `DottedBorderRect` to expose `ctx`.

But I think I prefer not to have to have another `useDraw` every time we want to use `CanvasRect` with a custom border. `drawBorder`

emersion: I would prefer to _compose_ simple, independent components instead of introducing flexible components with more options and inter-dependencies. `fillStyle` and `drawBorder` are overly generic IMHO: passing callbacks and options introduces more complexity than applying simple, basic components in sequence. Repetition of a few lines of simple code is more maintainable than adding abstractions and passing callbacks around, even if it's slightly more boilerplate.

When repetition is too much, isolating only the concrete repeated code (e.g. the code responsible for filling a rect given a canvas context) is more maintainable than making components more generic. Here for instance, if we decided that the `getTimePixel()` + `ctx.translate()` + `ctx.fillRect()` are too much to repeat, we could introduce a shared `fillRect(ctx, { timeStart, timeEnd, spaceStart, spaceEnd })` function called from `PatternRect` and a component drawing the selection dotted rectangle, instead of introducing a flexible `CanvasRect` with options to accommodate for all use-cases.

What do I mean by "maintainable"? I refer specifically to _software engineering_, which is what happens when to programming when you add time and other programmers. Lots of other people will need to read the code we write, and these people won't have the context we have when we're writing it. In fact, a piece of code is read a lot more times than it's written/edited. When reviewing, debugging, touching other code calling into or called by the code we've written, others will need to understand how it works. That's why I try to optimize for _readability_ when I write code. If I write some code and then tell to myself, "that's smart", then the alarm bells start ringing: others will need to unfold the smartness to understand the code (note, "others" include future-me).

Sorry for the tangent - I'm trying to take a step back and explain the motivations behind the kind of comments that I often throw at you :P

ui-spacetimechart/src/components/SimpleRect.tsx
emersion: I've replied below: https://github.com/OpenRailAssociation/osrd-ui/pull/870#discussion_r1975214988

> concern which you brought up initially


That comment was made before I realized that we don't actually need to draw a pattern for the selection rect.

emersion:
Just to make things a bit more explicit, here's how I would've designed the components:

- A `ZoomRect` which takes only `{ timeStart, timeEnd, spaceStart, spaceEnd }` as props and draws the rect (border + background) according to the mockup. (Don't leak details such as `fillStyle`/`lineWidth`/`spacing` in the API so that we can adjust the style without breaking our API.)
- Optionally, a shared `fillRect(ctx, { timeStart, timeEnd, spaceStart, spaceEnd })` util used by both `ZoomRect` and `PatternRect`.
- No `CanvasRect`.

All of that of course up for discussion.

ui-spacetimechart/src/components/CanvasRect.tsx
anisometropie: Ok, I see now, where you’re coming from every time I see «not a fan» popping in my PR :p

Thanks ! it’s always nice to spell out one’s reasoning and what motivates one’s decisions. I do appreciate your taking the time to articulate yours.

I’ll go with your solution with ZoomRect, fillRect, looks very neat

anisometropie:
> Just to make things a bit more explicit, here's how I would've designed the components:
> 
> * A `ZoomRect` which takes only `{ timeStart, timeEnd, spaceStart, spaceEnd }` as props and draws the rect (border + background) according to the mockup. (Don't leak details such as `fillStyle`/`lineWidth`/`spacing` in the API so that we can adjust the style without breaking our API.)
> * Optionally, a shared `fillRect(ctx, { timeStart, timeEnd, spaceStart, spaceEnd })` util used by both `ZoomRect` and `PatternRect`.
> * No `CanvasRect`.
> 
> All of that of course up for discussion.

I did the changes

emersion (approved):
LGTM, thank you!

PR #883, opened 2025-02-05
ui-spacetimechart: fix time labels, graduations and captions

### LINKED ISSUE #10688: time graduation off by 1 or 2 hour
Time axis should show 00 06 12. 01, 07, 13 make no sense
Image

largest vertical line should be midnight

Image

### LINKED ISSUE #10689: Space time chart time caption is off
Image

anisometropie:
closes https://github.com/OpenRailAssociation/osrd/issues/10688
closes https://github.com/OpenRailAssociation/osrd/issues/10689

### DIFF
--- a/ui-spacetimechart/src/stories/lib/components.tsx
+++ b/ui-spacetimechart/src/stories/lib/components.tsx
@@ -83,7 +83,7 @@ const DataLabel = ({
         </>
       ) : (
         <>
-          <div>Time: {new Date(data.time).toLocaleTimeString('en-GB', { timeZone: 'UTC' })}</div>
+          <div>Time: {new Date(data.time).toLocaleTimeString()}</div>
           <div>Distance: {round(data.position).toLocaleString()} m</div>
         </>
       )}
--- a/ui-spacetimechart/src/utils/canvas.ts
+++ b/ui-spacetimechart/src/utils/canvas.ts
@@ -1,4 +1,4 @@
-import { clamp } from 'lodash';
+import { clamp, identity } from 'lodash';
 
 import { type PathEnd, type Point, type RGBAColor, type RGBColor } from '../lib/types';
 
@@ -272,4 +272,41 @@ export function drawPathExtremity(
   } else {
     drawPathStopExtremity(ctx, timePixel, spacePixel, swapAxis);
   }
+}
+
+/**
+ *
+ * @param minT number timestamp
+ * @param maxT number timestamp
+ * @param timeRanges time frames (24h, 12h, 6h, …)
+ * @param gridlinesLevels width of the lines for each time frame
+ * @param formatter function to format de values inside the output object
+ * @returns Record<number, number>
+ * Keys are times in ms
+ * Values are the highest level on each time
+ */
+export function computeVisibleTimeMarkers<T>(
+  minT: number,
+  maxT: number,
+  timeRanges: number[],
+  levels: number[],
+  formatter: (level: number, i: number) => T = identity
+) {
+  const result: Record<number, T> = {};
+  const minTLocalOffset = new Date(minT).getTimezoneOffset() * 60 * 1000;
+
+  timeRanges.forEach((range, i) => {
+    const gridlinesLevel = levels[i];
+
+    if (!gridlinesLevel) return;
+
+    let t = Math.floor((minT - minTLocalOffset) / range) * range + minTLocalOffset;
+    while (t <= maxT) {
+      if (t >= minT) {
+        result[t] = formatter(gridlinesLevel, i);
+      }
+      t += range;
+    }
+  });
+  return result;
 }
--- a/ui-spacetimechart/src/components/TimeCaptions.tsx
+++ b/ui-spacetimechart/src/components/TimeCaptions.tsx
@@ -3,6 +3,7 @@ import { useCallback } from 'react';
 import { useDraw } from '../hooks/useCanvas';
 import { MINUTE } from '../lib/consts';
 import { type DrawingFunction } from '../lib/types';
+import { computeVisibleTimeMarkers } from '../utils/canvas';
 
 const MINUTES_FORMATTER = (t: number) => `:${new Date(t).getMinutes().toString().padStart(2, '0')}`;
 const HOURS_FORMATTER = (t: number, pixelsPerMinute: number) => {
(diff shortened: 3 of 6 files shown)


ui-spacetimechart/src/utils/canvas.ts
```diff
@@ -273,3 +273,39 @@ export function drawPathExtremity(
...
+
+export function displayElementsBasedOnZoom<T>(
+  minT: number,
+  maxT: number,
+  timeRanges: number[],
+  levels: number[],
+  formatter: (level: number, i: number) => T = identity
+) {
+  const result: Record<number, T> = {};
+  const localOffset = -new Date(minT).getTimezoneOffset() * 60 * 1000;
```
emersion: In general I'm not a fan of using `getTimezoneOffset()`, because it behaves wrongly on timezone offset change. For instance if daylight savings are applied on the left side of the chart but aren't on the right side, the chart will be wrong.

But it's not the end of the world here, since it's just about graduations.

I'd suggest applying the offset only to `minT` since it's the only timestamp it's valid for:

```ts
const localOffset = new Date(minT).getTimezoneOffset() * 60 * 1000;
let t = Math.floor((minT - localOffset) / range) * range + localOffset;
```

then the condition inside the loop can be dropped, and all other `localOffset` operations can be removed.

(By the way, maybe `minTLocalOffset` may be a more descriptive name to make it clear that it's only valid for `minT`.)

ui-spacetimechart/src/utils/canvas.ts
```diff
@@ -273,3 +273,39 @@ export function drawPathExtremity(
...
+ * Keys are times in ms
+ * Values are the highest level on each time
+ */
+
+export function displayElementsBasedOnZoom<T>(
+  minT: number,
+  maxT: number,
+  timeRanges: number[],
+  levels: number[],
+  formatter: (level: number, i: number) => T = identity
```
emersion, line 293: I'd prefer to apply this as a second step (ie, the caller can do it) to reduce the complexity of that function. (The docs seem outdated.)

ui-spacetimechart/src/utils/canvas.ts
```diff
@@ -273,3 +273,39 @@ export function drawPathExtremity(
...
+ * @param minT number timestamp
+ * @param maxT number timestamp
+ * @param timeRanges time frames (24h, 12h, 6h, …)
+ * @param gridlinesLevels width of the lines for each time frame
+ * @returns Record<number, number>
+ * Keys are times in ms
+ * Values are the highest level on each time
+ */
+
+export function displayElementsBasedOnZoom<T>(
```
emersion: I wonder if we could find a better name of this function.

The function itself doesn't deal with the zoom so it sounds a bit weird to mention it there.

Maybe `computeMarksFromTimeLevels()`?

ui-spacetimechart/src/stories/lib/components.tsx
```diff
@@ -83,7 +83,7 @@ const DataLabel = ({
         </>
       ) : (
         <>
-          <div>Time: {new Date(data.time).toLocaleTimeString('en-GB', { timeZone: 'UTC' })}</div>
+          <div>Time: {new Date(data.time).toLocaleTimeString()}</div>
```
emersion, line 86: Good catch!

ui-spacetimechart/src/utils/canvas.ts
anisometropie: `computeVisibleTimeMarkers` ?

ui-spacetimechart/src/utils/canvas.ts
anisometropie: that issue existed already, we have days with only 23h or 25h, everything becomes shifted, because we add 86400 s each time.

thanks, it’s neater like this.
Why do we need the offset inside the Math.floor ?
I think `let t = Math.floor(minT / range) * range - minTLocalOffset;` is enough.

ui-spacetimechart/src/utils/canvas.ts
anisometropie: we have to use Object.entries / re map over the object / Object.fromEntries
this seems way more cumbersome than passing a simple formatter

ui-spacetimechart/src/utils/canvas.ts
emersion: Works for me!

ui-spacetimechart/src/utils/canvas.ts
emersion: My reasoning is that all timestamps are Unix time without any offset. If we apply the offset once without undoing it, then it's not Unix time anymore, or am I missing something?

Maybe the `Math.floor()` makes it so we get the correct number by chance if we apply the offset only once? However I think it'd break on other `range` and offset values.

ui-spacetimechart/src/utils/canvas.ts
emersion: `mapValues` from lodash makes this less cumbersome, FWIW.

(But I'm not feeling strongly about this.)

ui-spacetimechart/src/utils/canvas.ts
anisometropie: We’re still getting a valid unix time. But instead of getting the timestamp of 0h UTC we get the timestamp of 22h UTC (so that it shows 0h Paris and not 2h Paris like before)

Not sure why it works with the offset applied twice to be quite frank

image_50353153

ui-spacetimechart/src/utils/canvas.ts
anisometropie: think i’m gonna stick with the original idea, `mapValues` doesn’t keep the indices

ui-spacetimechart/src/utils/canvas.ts
emersion: Sorry, I still don't believe it's correct. `minT` is the earliest timestamp in the time window. We want to find the first graduation marker. We need to snap `minT` to a `range`-spaced grid shifted by the offset. Everything needs to be UTC apart from the snapping.

Your solution works as long as `minTLocalOffset` is negative, because the loop keeps going until it reaches `maxT` (it'll just run longer). However, it breaks with positive timezone offsets. You can try e.g. `minTLocalOffset = 6 * 60 * 60 * 1000` and notice missing grid lines on the left.

anisometropie: Ok I see the issue, we get `t` snapped to 0h local time in both case;
but we sometimes get 0h off by one day, so we miss some of the range, thanks for digging that up ! 

I’ve written a test to take that case into account

emersion: Sweet, thanks for adding a test!

emersion (approved):
LGTM!

emersion:
Typo: "up-spacetimechart: fix time graduation" (should be "ui-", not "up-")

PR #888, opened 2025-02-06
Enable import/no-extraneous-dependencies lint

emersion:
_See individual commits._

Ensure all packages have dependencies properly defined. This is
especially important for a monorepo because dependencies might get
pulled in by another package by chance.

Disable the check for stories, because these depend on the package
they're in.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -90,7 +90,13 @@
         "tsx": "never",
       },
     ],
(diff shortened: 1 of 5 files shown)


kmer2016 (approved):
LGTM !

Caracol3 (approved):
LGTM

PR #898, opened 2025-02-12
Centralize dev dependencies

emersion:
_See individual commits._

osrd-ui's root package provides some re-usable build tool configurations, for instance postcss-base.config.cjs and rollup-base.config.js.

Make it so the root package contains the devDependencies required to run these configuration files instead of duplicating them in each sub-package. That way, we don't need to keep all of these in sync when adding/removing a plugin.

While at it, drop unused dependencies such as rollup-plugin-serve.


Synar (approved):
LGTM ^^

Math-R (approved):
lgtm ! thanks !

PR #900, opened 2025-02-13
ui-manchette-with-space-time-chart: return xZoom

clarani:
This props was removed accidently [here](https://github.com/OpenRailAssociation/osrd-ui/commit/ab5aa4e8c2f352482d8c5b44d74aee43b0dae8d5).
Since it is missing, we can't bump osrd-ui in osrd.


anisometropie (approved):

Math-R (approved):
lgtm

PR #913, opened 2025-02-18
ui-spacetimechart: fixes panning with dpr < 1

### LINKED ISSUE #912: space time chart: glitches when devicePixelRatio < 1
### What happened?

https://github.com/user-attachments/assets/7b020576-31df-4aac-817e-3186d6c6ace2

When panning the space time chart with a `devicePixelRatio` less than 1 (appears when users zoom out, for instance), glitches appear in the right and in the bottom of the chart.

The component should behave as it does when the `devicePixelRatio` is 1 or more.

1. Go to [this page](https://openrailassociation.github.io/osrd-ui/?path=/story/spacetimechart-measuring-times-and-distances--default-args)
2. Zoom out in your browser (with `Ctrl` + `-` for instance)
3. Pan the space time chart

Local Storybook

Firefox

jacomyal:
This commit fixes #912.

It updates stage clearings to compensate the scaling done to fit the device's pixel ratio.

### DIFF
--- a/ui-spacetimechart/src/hooks/useCanvas.ts
+++ b/ui-spacetimechart/src/hooks/useCanvas.ts
@@ -45,8 +45,18 @@ export function useCanvas(
   const stcContextRef = useRef(inputStcContext);
   const scheduledRef = useRef<null | { frameId: number }>(null);
 
-  const size = useSize(dom);
   const [hoveredItem, setHoveredItem] = useState<HoveredItem | null>(null);
+
+  const size = useSize(dom);
+  const sizeRef = useRef<typeof size>(size);
+
+  /**
+   * Keep the sizeRef value up to date (to allow rendering functions not being reset anytime size is
+   * updated):
+   */
+  useEffect(() => {
+    sizeRef.current = size;
+  }, [size]);
 
   /**
    * This function renders all picking layers:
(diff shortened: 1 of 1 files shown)


SharglutDev (approved):
As far as I know, lgtm and tested :)

ui-spacetimechart/src/hooks/useCanvas.ts
```diff
@@ -61,7 +61,9 @@ export function useCanvas(
         const set = pickingFunctions.current[layer];
 
         if (ctx) {
-          ctx.clearRect(0, 0, ctx.canvas.width, ctx.canvas.height);
+          const scale = window.devicePixelRatio || 1;
+          ctx.clearRect(0, 0, ctx.canvas.width / scale, ctx.canvas.height / scale);
```
emersion: Could we use `size` instead here? This would avoid duplicating the `devicePixelRatio` logic in case we update how the size is computed in the future.

Alternatively, we could reset the transform:

```ts
ctx.save();
ctx.setTransform(1, 0, 0, 1, 0, 0); // identity
ctx.clearRect(0, 0, canvas.width, canvas.height);
ctx.restore();
```

jacomyal: I went with the `size` version (felt more readable to me than adding more `transform` / `scale` shenanigans...). I had to store `size` in a new ref, though, or I'd have to update the lifecycles of most callbacks and effects, and I fear it might have had too much impact.

I also realized that the picking layer is broken similarly with lower DPRs, but I couldn't find a simple fix yet. I'll open a new ticket about it.

emersion (approved):

ui-spacetimechart/src/hooks/useCanvas.ts
emersion: WFM

PR #914, opened 2025-02-19
ui-spacetimechart: fixes blurry graduations

### LINKED ISSUE #10662: front: different resolution for space time chart and manchette
### Description and goal

The resolution is not the same for the space time chart and the manchette. This should not happen

Image

Here you can see that the space time chart is a bit blurred

### Acceptance criteria

The space time chart should not be blurred

jacomyal:
This commit fixes OpenRailAssociation/osrd#10662.

It ensures vertical and horizontal graduation lines are as thin as possible, so that they are as crisp as they can be.

### DIFF
--- a/ui-spacetimechart/src/components/PathLayer.tsx
+++ b/ui-spacetimechart/src/components/PathLayer.tsx
@@ -14,7 +14,12 @@ import {
   type Point,
   type SpaceTimeChartContextType,
 } from '../lib/types';
-import { drawAliasedDisc, drawAliasedLine, drawPathExtremity } from '../utils/canvas';
+import {
+  drawAliasedDisc,
+  drawAliasedLine,
+  drawPathExtremity,
+  getCrispLineCoordinate,
+} from '../utils/canvas';
 import { indexToColor, hexToRgb } from '../utils/colors';
 import { getSpaceBreakpoints } from '../utils/scales';
 
@@ -170,7 +175,7 @@ export const PathLayer = ({
         if (i) {
           const { position: prevPosition, time: prevTime } = a[i - 1];
           if (prevPosition === position && stopPositions.has(position)) {
-            const spacePixel = getSpacePixel(position);
+            const spacePixel = getCrispLineCoordinate(getSpacePixel(position), ctx.lineWidth);
             ctx.beginPath();
             if (!swapAxis) {
               ctx.moveTo(getTimePixel(prevTime), spacePixel);
--- a/ui-spacetimechart/src/utils/canvas.ts
+++ b/ui-spacetimechart/src/utils/canvas.ts
@@ -289,14 +289,14 @@ export function computeVisibleTimeMarkers<T>(
   minT: number,
   maxT: number,
   timeRanges: number[],
-  levels: number[],
+  gridlinesLevels: number[],
   formatter: (level: number, i: number) => T = identity
 ) {
   const result: Record<number, T> = {};
   const minTLocalOffset = new Date(minT).getTimezoneOffset() * 60 * 1000;
 
   timeRanges.forEach((range, i) => {
-    const gridlinesLevel = levels[i];
+    const gridlinesLevel = gridlinesLevels[i];
 
     if (!gridlinesLevel) return;
 
@@ -309,4 +309,25 @@ export function computeVisibleTimeMarkers<T>(
     }
   });
   return result;
+}
+
+/**
+ * To get crisp horizontal or vertical lines on a canvas, we must draw them as thin as possible, in
+ * terms of actual pixels on screen.
+ * The best way for this is:
+ * - To center lines 1, 3, 5... pixels wide in the middle of a pixel
+ * - To center lines 2, 4, 6... pixels wide between two pixels
+ * @param rawCoordinate Any input coordinate to fix
+ * @param lineWidth The width of the line to draw
+ * @param devicePixelRatio
+ */
+export function getCrispLineCoordinate(
+  rawCoordinate: number,
+  lineWidth: number,
+  devicePixelRatio = window.devicePixelRatio || 1
+): number {
+  const centerOffset = lineWidth / 2;
+  return (
+    Math.round((rawCoordinate - centerOffset) * devicePixelRatio) / devicePixelRatio + centerOffset
+  );
 }
--- a/ui-spacetimechart/src/__tests__/canvas.spec.ts
+++ b/ui-spacetimechart/src/__tests__/canvas.spec.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'vitest';
 
-import { getAliasedDiscShape } from '../utils/canvas';
+import { getAliasedDiscShape, getCrispLineCoordinate } from '../utils/canvas';
 
 describe('getAliasedDiscShape', () => {
   it('should return return the expected flat matrices', () => {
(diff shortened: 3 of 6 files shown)


ui-spacetimechart/src/components/PathLayer.tsx
```diff
@@ -170,7 +175,7 @@ export const PathLayer = ({
         if (i) {
           const { position: prevPosition, time: prevTime } = a[i - 1];
           if (prevPosition === position && stopPositions.has(position)) {
-            const spacePixel = getSpacePixel(position);
+            const spacePixel = getCrispLineCoordinate(getSpacePixel(position), ctx.lineWidth);
```
emersion, line 178: Should we adjust the space pixel for paths?

The path in general isn't a straight vertical line, does it make sense to center the curve x positions vertically?

Why is the time axis handled differently?

ui-spacetimechart/src/utils/canvas.ts
```diff
@@ -310,3 +310,17 @@ export function computeVisibleTimeMarkers<T>(
...
+ * terms of actual pixels on screen.
+ * The best way for this is:
+ * - To center lines 1, 3, 5... pixels wide in the middle of a pixel
+ * - To center lines 2, 4, 6... pixels wide between two pixels
+ * @param rawCoordinate Any input coordinate to fix
+ * @param lineWidth The width of the line to draw
+ */
+export function getCrispLineCoordinate(rawCoordinate: number, lineWidth: number): number {
+  const centerOffset = (Math.ceil(lineWidth) % 2) / 2;
+  return Math.round(rawCoordinate - centerOffset) + centerOffset;
```
emersion: Took me a bit to double-check this formula, it seems to work perfectly fine on LoDPI screens :)

I'm not super sure about HiDPI though. On HiDPI screens, the numbers will be implicitly multiplied by the scale before drawing on screen. Ideally, on a `devicePixelRatio = 2` setup, a 0.5 large line at rawCoordinate 0 would be painted at coordinate 0.25, right? Also, on a HiDPI screen it's not necessary to align a 2px large line between two pixels: painting at x = 0.5 will result in a crisp image, I think?

Should we read `window.devicePixelRatio` here, and change our behavior based on that number? (Mainly how rounding is performed)

Or should we give up and passthrough any number which isn't an integer? Or do you see another solution?

ui-spacetimechart/src/components/PathLayer.tsx
jacomyal: This specific code is just to render the lighter and thicker part in the background, when trains pause, not the path itself. And paths always pause with constant space, not constant time.
image

ui-spacetimechart/src/utils/canvas.ts
jacomyal: Oh good point indeed... Let me check that again :)

ui-spacetimechart/src/components/PathLayer.tsx
emersion: Oh yeah that makes perfect sense!

ui-spacetimechart/src/utils/canvas.ts
jacomyal: Well, I ran some tests on my 3.5 DPR Android phone, and:
- I couldn't get pixel perfect lines (when I check my screen captures zoomed on GIMP at least)
- They appear very acceptable on screen to me, I wouldn't have guessed it wasn't pixel perfect
- The resolution is such that I couldn't see any difference from afar, in all the tests I ran

To give a bit more context, the solution that made the most sense to me was to update the formula as:

```typescript
export function getCrispLineCoordinate(rawCoordinate: number, lineWidth: number): number {
  const centerOffset = (Math.ceil(lineWidth * devicePixelRatio) % 2) / 2;
  return Math.round(rawCoordinate - centerOffset) + centerOffset;
}
```

Then, for a 2px wide line, on my phone, it would be 7px wide effectively, and thus with a `centerOffset` of 0.5. I checked with `console.log`, and that's what I got. And here is how it was rendered:

image

Zoomed in, it's clearly not crisp, and it's not even symmetric. But on my phone, with the actual screen, it looks very crisp. So, at that point, I really don't know how to do better :sweat_smile: Also, it looks like that last formula breaks with lower DPIs.

I'd say we go with this for now, at least if we agree that the lines are crisp enough on high DPIs.

emersion: I've written a bunch of tests to more easily play with the code and ensure it does what I'd expect:

<details>

```ts
describe('getCrispLineCoordinate', () => {
  it.each([1, 3, 99])('should align a %ipx line on a LoDPI device', (lineWidth) => {
    const devicePixelRatio = 1;
    expect(getCrispLineCoordinate(0, lineWidth, devicePixelRatio)).toEqual(0.5);
    expect(getCrispLineCoordinate(0.1, lineWidth, devicePixelRatio)).toEqual(0.5);
    expect(getCrispLineCoordinate(0.5, lineWidth, devicePixelRatio)).toEqual(0.5);
    expect(getCrispLineCoordinate(0.9, lineWidth, devicePixelRatio)).toEqual(0.5);
    expect(getCrispLineCoordinate(1, lineWidth, devicePixelRatio)).toEqual(1.5);
    expect(getCrispLineCoordinate(42, lineWidth, devicePixelRatio)).toEqual(42.5);
    expect(getCrispLineCoordinate(-0.4, lineWidth, devicePixelRatio)).toEqual(-0.5);
    expect(getCrispLineCoordinate(-0.7, lineWidth, devicePixelRatio)).toEqual(-0.5);
  });

  it.each([2, 4, 64])('should align a 2px line on a LoDPI device', (lineWidth) => {
    const devicePixelRatio = 1;
    expect(getCrispLineCoordinate(-0.4, lineWidth, devicePixelRatio)).toEqual(0);
    expect(getCrispLineCoordinate(0, lineWidth, devicePixelRatio)).toEqual(0);
    expect(getCrispLineCoordinate(0.4, lineWidth, devicePixelRatio)).toEqual(0);
    expect(getCrispLineCoordinate(0.5, lineWidth, devicePixelRatio)).toEqual(1);
    expect(getCrispLineCoordinate(1.5, lineWidth, devicePixelRatio)).toEqual(2);
    expect(getCrispLineCoordinate(42, lineWidth, devicePixelRatio)).toEqual(42);
    expect(getCrispLineCoordinate(-0.7, lineWidth, devicePixelRatio)).toEqual(-1);
  });

  it('should align a 2px line on a HiDPI device', () => {
    const devicePixelRatio = 2;
    const lineWidth = 2;
    expect(getCrispLineCoordinate(0.5, lineWidth, devicePixelRatio)).toEqual(0.5);
  });

  it('should align a 0.5px line on a HiDPI device', () => {
    const devicePixelRatio = 2;
    const lineWidth = 0.5;
    expect(getCrispLineCoordinate(0, lineWidth, devicePixelRatio)).toEqual(0.25);
  });
});
```

</details>

I've simplified your code to the following:

```ts
export function getCrispLineCoordinate(rawCoordinate: number, lineWidth: number): number {
  const centerOffset = lineWidth / 2;
  return Math.round(rawCoordinate - centerOffset) + centerOffset;
}
```

and then I've extended it to handle HiDPI:

```ts
export function getCrispLineCoordinate(
  rawCoordinate: number,
  lineWidth: number,
  devicePixelRatio = window.devicePixelRatio || 1
): number {
  const centerOffset = lineWidth / 2;
  return (
    Math.round((rawCoordinate - centerOffset) * devicePixelRatio) / devicePixelRatio + centerOffset
  );
}
```

Do you think this is a version we could use?

jacomyal: I tested it, and it looks good to me. I pushed your modifications to the PR.

emersion (commented):
LGTM!

emersion (approved):

PR #917, opened 2025-02-19
Upgrade storybook to v8.5.8

emersion:
_See individual commits._

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2270,12 +2270,13 @@
       }
     },
     "node_modules/@joshwooding/vite-plugin-react-docgen-typescript": {
-      "version": "0.4.2",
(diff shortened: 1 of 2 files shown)


Uriel-Sautron (approved):
LGTM tested

SharglutDev (approved):
We don't need to update the `package.json` as well ?

emersion:
Here it's a bugfix version upgrade, so we don't need to (the code in this PR works fine with both the old and new versions of storybook). Side note, dependabot also doesn't update `package.json` in that case.

During a minor version upgrade, it's indeed best to bump the version in `package.json`, to indicate that the app might be broken with older versions, e.g. because features from the new version start being used (dependabot does this, but `npm update` doesn't).

PR #918, opened 2025-02-19
ui-spacetimechart: handle devicePixelRatio changes

emersion:
Introduce a small hook to listen to devicePixelRatio changes. Useful to react to browser zoom or switching a window between HiDPI/LoDPI screens (and still render a crisp frame in these cases).

### DIFF
--- a/ui-spacetimechart/src/hooks/useCanvas.ts
+++ b/ui-spacetimechart/src/hooks/useCanvas.ts
@@ -2,6 +2,7 @@ import { useCallback, useContext, useEffect, useMemo, useRef, useState } from 'r
 
 import { isEqual } from 'lodash';
 
+import { useDevicePixelRatio } from './useDevicePixelRatio';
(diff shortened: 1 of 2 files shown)


jacomyal:
I'm not sure it's that useful, since I can't imagine a case where DPR is changed without a resize trigger, but the code is real clean and it can't hurt. I didn't know about that smart method to detect DPR changes either, that's really nice.
LGTM!

jacomyal (approved):

emersion:
Here are two situations on my machine where `devicePixelRatio` changes while the `ResizeObserver` doesn't (checked with a `console.log()` in `updateSize()`):

- Set a fixed width/height on the `.space-time-chart` HTML element, then zoom/dezoom.
- On a multi-monitor setup with one HiDPI screen and one LoDPI screen, make the browser window floating and drag it between monitors.

kmer2016 (approved):
LGTM

PR #919, opened 2025-02-21
ui-spacetimechart: implements "splitting"

jacomyal:
This pull request addresses issue osrd-project/osrd-confidential#870.

It allows giving a SpaceTimeChart **flat steps**, i.e. `SpaceScales` with a `to` value equal to the previous one's (where the chart is effectively split). Then, it adds a story that combines flat steps with a custom layer, to overdraw flat steps with a custom component.

Details:
- Adds back-and-forth trains in stories
- Improves and fixes `getPathDirection`, to handle trains that change directions
- Updates `getSpaceBreakpoints` to no more deduplicate repeated values, which allows easily detecting flat steps
- Updates `getNormalizedScaleAtPosition` and `getSpaceToPixel` to allow retrieving both sides of the flat steps
- Fixes `getPathSegments` with flat steps
- Updates space graduations and path pauses rendering, so that they are rendered on both sides of the flat steps
- Adds a story to showcase all this

### DIFF
--- a/ui-spacetimechart/src/components/PathLayer.tsx
+++ b/ui-spacetimechart/src/components/PathLayer.tsx
@@ -21,27 +21,8 @@ import {
   getCrispLineCoordinate,
 } from '../utils/canvas';
 import { indexToColor, hexToRgb } from '../utils/colors';
+import { getPathDirection } from '../utils/paths';
 import { getSpaceBreakpoints } from '../utils/scales';
-
-function getDirection({ points }: PathData, reverse?: boolean): 'up' | 'down' {
-  if (points.length < 2) return 'down';
-
-  if (!reverse) {
-    for (let i = 1, l = Math.min(3, points.length); i < l; i++) {
-      const diff = points[i].position - points[i - 1].position;
-      if (diff > 0) return 'up';
-      if (diff < 0) return 'down';
-    }
-  } else {
-    for (let i = points.length - 2, l = Math.max(points.length - 4, points.length); i > l; i--) {
-      const diff = points[i].position - points[i + 1].position;
-      if (diff > 0) return 'up';
-      if (diff < 0) return 'down';
-    }
-  }
-
-  return 'down';
-}
 
 const DEFAULT_PICKING_TOLERANCE = 5;
 const PAUSE_THICKNESS = 7;
@@ -284,7 +265,6 @@ export const PathLayer = ({
     (ctx, { getTimePixel, getSpacePixel, swapAxis }) => {
       if (!path.points.length) return;
 
-      const pathDirection = getDirection(path);
       const from = path.points[0];
       const fromEnd = path.fromEnd || DEFAULT_PATH_END;
       const to = last(path.points) as DataPoint;
@@ -296,7 +276,7 @@ export const PathLayer = ({
         getSpacePixel(from.position),
         swapAxis,
         'from',
-        pathDirection,
+        getPathDirection(path, 0),
         fromEnd
       );
       drawPathExtremity(
@@ -305,7 +285,7 @@ export const PathLayer = ({
         getSpacePixel(to.position),
         swapAxis,
         'to',
-        pathDirection,
+        getPathDirection(path, path.points.length - 1, true),
         toEnd
       );
     },
--- a/ui-spacetimechart/src/components/PathLayer.tsx
+++ b/ui-spacetimechart/src/components/PathLayer.tsx
@@ -21,7 +21,7 @@ import {
   getCrispLineCoordinate,
 } from '../utils/canvas';
 import { indexToColor, hexToRgb } from '../utils/colors';
-import { getPathDirection } from '../utils/paths';
+import { getPathDirection, getSpacePixels } from '../utils/paths';
 import { getSpaceBreakpoints } from '../utils/scales';
 
 const DEFAULT_PICKING_TOLERANCE = 5;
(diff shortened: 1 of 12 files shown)


ui-spacetimechart/src/stories/split.stories.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  // For this story, we split the chart on "City C" and "City E:
+  const splitPoints = useMemo(() => {
+    const operationalPointsDict = keyBy(OPERATIONAL_POINTS, 'id');
+    return [
+      {
+        position: operationalPointsDict['city-c'].position,
+        label: operationalPointsDict['city-c'].label,
+        height: 100,
+      },
```
Yohh:
```suggestion
      {
        position: operationalPointsDict['city-c'].position,
        label: operationalPointsDict['city-c'].label,
        height: 100 * state.yZoomLevel,
      },
```

if it's on purpose, you should add a comment

ui-spacetimechart/src/components/PathLayer.tsx
```diff
@@ -175,16 +169,26 @@ export const PathLayer = ({
...
+            // Detect flat steps, and draw two graduations if any (one on each side of the step):
+            const spacePixelFromStart = getSpacePixel(position);
+            const spacePixelFromEnd = getSpacePixel(position, true);
+            const spacePixels =
+              spacePixelFromStart === spacePixelFromEnd
+                ? [spacePixelFromStart]
+                : [spacePixelFromStart, spacePixelFromEnd];
+
+            spacePixels.forEach((rawPixel) => {
+              const spacePixel = getCrispLineCoordinate(rawPixel, ctx.lineWidth);
```
Yohh: duplicated in `SpaceGraduations.tsx`, could be in a utils function

ui-spacetimechart/src/stories/split.stories.tsx
jacomyal: It was on purpose, to showcase that it is possible to add both steps with constant (in pixel) heights, and steps with a height that scales with the zoom. I'll add a comment while moving the breakpoints to the story arguments.

jacomyal: I just pushed new arguments for the story, including the height of the steps, and a flag to specify if the steps should scale with the zoom or not.

ui-spacetimechart/src/components/PathLayer.tsx
jacomyal: Fixed

ui-spacetimechart/src/stories/split.stories.tsx
```diff
@@ -0,0 +1,204 @@
...
+  },
+} as Meta<typeof SplitSpaceTimeChartWrapper>;
+
+export const Default = {
+  name: 'Default arguments',
+  args: {
+    splitPoints: 'A,C,E',
+    splitHeight: 100,
+    scaleWithZoom: false,
+    swapAxis: false,
```
Yohh, line 202: maybe a little bit overkill, but we can add a `scale` boolean arg, if `true` it could multiply `spliHeight`by `state.yZoomLevel` to display the two possible behaviors

jacomyal: That's what the `scaleWithZoom` does exactly :)

Yohh: oh sorry!

Yohh (approved):

theocrsb (approved):
LGTM and tested

ui-spacetimechart/src/stories/lib/paths.ts
```diff
@@ -137,9 +155,31 @@ export const PATHS: (PathData & { color: string })[] = [
...
+  // Back and forth trains:
+  ...getPaths(
+    'back-and-forth',
+    BACK_AND_FORTH_POINTS,
+    10 * MIN,
+    30 * MIN,
+    (80 * KM) / (60 * MIN),
+    2,
+    +START_DATE + 15 * MIN,
+    { color: '#286109', toEnd: 'out' }
```
Math-R, line 175: nit : Maybe you can put the colors in a dictionnary as the cities, or we could put a TODO to remember to change the hardcoded colors to colors that will be shared via a dictionary in the [new ui-chart module](https://github.com/OpenRailAssociation/osrd-ui/pull/931)

ui-spacetimechart/src/stories/split.stories.tsx
```diff
@@ -0,0 +1,204 @@
...
+          },
+          {
+            to: position,
+            size: height,
+          },
+        ])
+        .concat({
+          to: OPERATIONAL_POINTS.at(-1)!.position,
+          coefficient: COEFFICIENT / state.yZoomLevel,
+        }),
```
Math-R, line 104: I'm surely missing something but why don't use a map instead of the flatMap and return an array of

```
{
      to: position,
      coefficient: COEFFICIENT / state.yZoomLevel,
      size: height,
 }
```
And then concat ?

Akctarus (approved):

ui-spacetimechart/src/utils/scales.ts
```diff
@@ -107,17 +107,26 @@ export function spaceScalesToBinaryTree(
  *
  * Also, if the position is lower than the tree's min (tree.from), then the first leaf is returned
  * and if it is higher than the max (tree.to), the last leaf is returned.
+ *
+ * Finally, if pickLast is truthy, then it returns the last leaf node that contains that position
+ * instead of the first one. It is very important when there are flat sections (
```
theocrsb: My comment isn't very helpful, but there's an extra parenthesis
```suggestion
 * instead of the first one. It is very important when there are flat sections
```

ui-spacetimechart/src/stories/split.stories.tsx
jacomyal: Hum, the `map` wouldn't work. What we want here is to "join" constant scales (the ones with the `coefficient`) with flat steps, basically. So, if we have one split point at `50` for instance, on a chart that goes from 0 to 120, we want:

```typescript
const spaceScales = [
  {
    to: 50,
    coefficient: CONSTANT_COEFF,
  },
  {
    to: 50, // Since the "to" is the same as the previous one, it's a flat step
    height: SPLIT_HEIGHT,
  },
  {
    to: 120, // Another constant "classic" step
    coefficient: CONSTANT_COEFF,
  },
]
```

And this will split the space in three different segments:
- The first one goes from 0 to 50, with a constant scale
- The second one is a flat step, and we can display other custom components on top, such as a track occupancy diagram, for instance
- The last one starts at 50, and goes to 120, with the same scale as the first one

ui-spacetimechart/src/stories/lib/paths.ts
jacomyal: Good point, I just pushed a `TODO` comment.

ui-spacetimechart/src/utils/scales.ts
jacomyal: Thanks! Fixed.

ui-spacetimechart/src/stories/split.stories.tsx
Math-R: Ok i see, thanks !

Math-R (approved):
Lgtm !

PR #924, opened 2025-02-21
front: remove input-wrapper class from input component

### LINKED ISSUE #921: Remove input-wapper from ui-core/input
### Description and goal

Currently, there is a div wrapper surrounding the input with a 3px padding. ([mockup](https://www.sketch.com/s/ca1f66cf-046d-4992-ae63-62187269ba7e/p/C50E2825-1B58-4269-A53A-5CE04E024E50/canvas))
This was probably done to simplify to focus behavior where we need to have a blueish border around the input.

Image

This adds complexity when we need to integrate ui-core/input in complex mockups like in stdcm because the distances between various elements are hard to calculate and maintain because of this.

We could remove this wrapper and display the focus effect with a pseudo-element in position absolute.

### Acceptance criteria

- [ ] the div containing `input-wrapper` class has been removed
- [ ] the `input-wrapper` class has been removed
- [ ] the input label remains 3px on top of the input
- [ ] focusing the input with tab still displays the proper style
- [ ] other inputs styles (leading/trailing content, statusWithIcon, tooltips...) work the same way

SharglutDev:
This div/class was used to easily display the focus style around the input when focusing it with tab. The focus style is now displayed via a pseudo element.

close #921 

Maybe there was other way to change this behavior, it's totally debatable.

### DIFF
--- a/ui-core/src/components/inputs/Input.tsx
+++ b/ui-core/src/components/inputs/Input.tsx
@@ -92,62 +92,61 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
           statusWithMessage?.tooltip || narrow ? 'before-status-message' : undefined
         }
         narrow={narrow}
-        className={cx('input-field-wrapper', inputFieldWrapperClassname)}
+        className={cx('input-field-wrapper', inputFieldWrapperClassname, {
+          small,
+        })}
         onCloseStatusMessage={onCloseStatusMessage}
       >
+        {leadingContent && (
+          <InputAffix
+            value={leadingContent}
+            type="leading"
+            disabled={disabled}
+            readOnly={readOnly}
+          />
+        )}
         <div
-          className={cx('input-wrapper', {
-            small,
+          className={cx('input-container', {
             'focused-by-tab': isFocusByTab,
           })}
         >
-          {leadingContent && (
-            <InputAffix
-              value={leadingContent}
-              type="leading"
-              disabled={disabled}
-              readOnly={readOnly}
-            />
-          )}
-          <div className="input-container">
-            <input
-              ref={ref}
-              className={cx('input', {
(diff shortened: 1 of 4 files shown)


ui-core/src/styles/inputs/input.css
```diff
@@ -26,6 +18,18 @@
     position: relative;
     width: 100%;
 
+    &.focused-by-tab {
+      &::before {
```
emersion: Usually keyboard focus is indicated via an [`outline`](https://developer.mozilla.org/en-US/docs/Web/CSS/outline). The distance between the border and the output can be set via `outline-offset`. Do you think we could use that instead of the pseudo-element?

SharglutDev: Didn't know that, thanks ! Done.

emersion (approved):
LGTM!

ui-core/src/components/inputs/Input.tsx
```diff
@@ -92,62 +92,62 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
           statusWithMessage?.tooltip || narrow ? 'before-status-message' : undefined
         }
         narrow={narrow}
-        className={cx('input-field-wrapper', inputFieldWrapperClassname)}
+        className={cx('input-field-wrapper', {
+          inputFieldWrapperClassname,
```
anisometropie: not sure what you’re trying to do.
you’re telling cx that the string `'inputFieldWrapperClassname'`  is a class name

emersion: Ow, good catch. It should be moved outside of the curly braces.

SharglutDev: Oops, nice catch yes ! Fixed.

anisometropie (approved):
thanks! (just reviewed)

PR #925, opened 2025-02-21
ui-charts: add zoom rectangle (part 3)

### LINKED ISSUE #10960: [storybook] manchette + Space-time rectangle zoom
The story has to meet the same [requirements] as the final result on OSRD (https://github.com/osrd-project/osrd-confidential/issues/528)

- [x] L'utilisateur clique sur un bouton en haut à droite du GET pour passer en mode zoom rectangulaire
  - [x] Le curseur change (curseur zoom)
  - [x] L'utilisateur peut re-cliquer sur le bouton pour déactiver le mode zoom rectangulaire
- [x] **En mode zoom rectangulaire**, l’utilisateur peut sélectionner une zone sur laquelle zoomer avec un clic maintenu enfoncé (clic simple = zoom par incrément) et un drag
  - [x] La zone pré-sélectionnée est grisée (voir maquette)
  - [x] Si l'utilisateur sort de la zone du GET, l'interface ne défile pas: seule la zone visible à l'écran est sélectionnable
  - [x] L'utilisateur ne peut pas zoomer avec la molette ou se déplacer pendant la sélection de la zone
- [x] Lorsque l'utilisateur relâche sa souris:
  - [x] Il est zoomé sur la zone sélectionnée
  - [x] Le mode zoom rectangulaire est désactivé
  - [x] la réglette en bas à droite se met à jour avec le bon niveau de zoom
  - [x] Si la zone sélectionnée correspond à un niveau de zoom supérieur au zoom autorisé, il est zoomé au maximum autorisé et centré sur la zone sélectionnée

anisometropie:
closes https://github.com/OpenRailAssociation/osrd/issues/10960

### y scale requirements
- **min zoom => the entire path is shown** 
  the scale min zoom metre/px value depends on the length of the path (which is not the case with the time axis, which are constants)
- **max zoom => the entire height of the manchette is 500 meter.**

### map yZoom -> mm/px scale
Previously, `yZoom` was used as a multiplier. `yScale` was just changed by increments of `0.5`  (using the + or - buttons) and the  height in pixel was just multiplied by `yScale`. This approach is not enough as we need to be able to navigate arbitrary zoom levels and to know how much the size of a rectangle means in term of yZoom.
To do that we need functions to map back and forth between the `yZoom` level and the value of the scale in mm/px. 
this is what `zoomValueToSpaceScale` `spaceScaleToZoomValue` do.

seeing that with the min zoom level we need to display the entire path, min Zoom scale value also depends on the length of the path. min and max scale values are given by `getExtremaScales`

#### `getExtremaScales` example:

Screenshot_20250312_120237-1

Min zoom level is given by the **entire path length** being represented between some top and bottom padding (here we have 489 px between them), which gives: `168 km/489 px = 343672 mm/px`

for max zoom level, we keep only the top padding (to keep the beginning of the path in place, while the arrival exits the view), which gives `500 m/505 px = 990 mm/px`

| zoom level | mm/px |
| ----- | ----- |
| 1 | 343672 |
| 10,5 | 990 |

https://www.desmos.com/calculator/ayz6dhx24a

#### non proportional mode
in non proportional mode (linear, every stop has the same height on the manchette), we’re not using a mm/px scale. We directly convert rectangle px height (extended to the full chart height) means in term of stop height.

### DIFF
--- a/ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
+++ b/ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
@@ -1,60 +1,141 @@
 import { useCallback, useEffect, useMemo, useState } from 'react';
 
+import { clamp } from 'lodash';
+
 import usePaths from './usePaths';
 import type { SpaceScale, SpaceTimeChartProps } from '../../spaceTimeChart';
-import { MAX_ZOOM_Y, MIN_ZOOM_Y, ZOOM_Y_DELTA, DEFAULT_ZOOM_MS_PER_PX } from '../consts';
+import {
+  MAX_ZOOM_Y,
+  MIN_ZOOM_Y,
+  ZOOM_Y_DELTA,
+  DEFAULT_ZOOM_MS_PER_PX,
+  MAX_ZOOM_MS_PER_PX,
+  MIN_ZOOM_MS_PER_PX,
+  BASE_WAYPOINT_HEIGHT,
+  FOOTER_HEIGHT,
+  WAYPOINT_LINE_HEIGHT,
+} from '../consts';
 import type { ProjectPathTrainResult, Waypoint } from '../types';
-import { getDistance } from '../utils';
+import { getDistance, calcTotalDistance } from '../utils';
 import {
   computeWaypointsToDisplay,
   getScales,
   zoomX,
   zoomValueToTimeScale,
   timeScaleToZoomValue,
+  spaceScaleToZoomValue,
+  getExtremaScales,
+  zoomValueToSpaceScale,
 } from '../utils/helpers';
 
 type State = {
   xZoom: number;
   yZoom: number;
+  timeOrigin: number;
+  spaceOrigin: number;
+  /** current x PIXEL offset from x origin */
   xOffset: number;
-  /** the current y-scroll of the view. always updates */
+  /** current y PIXEL offset from y origin (the current y-scroll of the view. always updates) */
   yOffset: number;
   /** only update after a zoom. used to update back the view scroll value */
   scrollTo: number | null;
   panning: { initialOffset: { x: number; y: number } } | null;
+  zoomMode: boolean;
+  rect: {
+    timeStart: Date;
+    timeEnd: Date;
+    spaceStart: number; // mm
+    spaceEnd: number; // mm
+  } | null;
+  pixelRect: {
+    xStart: number;
+    xEnd: number;
+    yStart: number;
+    yEnd: number;
+  } | null;
   isProportional: boolean;
   waypointsChart: Waypoint[];
   scales: SpaceScale[];
 };
 
-const useManchettesWithSpaceTimeChart = (
-  waypoints: Waypoint[],
-  projectPathTrainResult: ProjectPathTrainResult[],
-  manchetteWithSpaceTimeChartContainer: React.RefObject<HTMLDivElement>,
-  selectedTrain?: number,
+const useManchetteWithSpaceTimeChart = ({
+  waypoints,
+  projectPathTrainResult,
+  manchetteWithSpaceTimeChartRef,
+  selectedTrain,
   height = 561,
-  spaceTimeChartRef?: React.RefObject<HTMLDivElement>
-) => {
+  spaceTimeChartRef,
+  defaultTimeOrigin = 0,
+  defaultSpaceOrigin = 0,
+}: {
+  waypoints: Waypoint[];
+  projectPathTrainResult: ProjectPathTrainResult[];
+  manchetteWithSpaceTimeChartRef: React.RefObject<HTMLDivElement>;
+  selectedTrain?: number;
+  height?: number;
+  spaceTimeChartRef?: React.RefObject<HTMLDivElement>;
+  defaultTimeOrigin?: number;
+  defaultSpaceOrigin?: number;
+}) => {
   const [isShiftPressed, setIsShiftPressed] = useState(false);
   const [state, setState] = useState<State>({
     xZoom: timeScaleToZoomValue(DEFAULT_ZOOM_MS_PER_PX),
     yZoom: 1,
+    timeOrigin: defaultTimeOrigin,
+    spaceOrigin: defaultSpaceOrigin,
     xOffset: 0,
     yOffset: 0,
     scrollTo: null,
     panning: null,
+    zoomMode: false,
+    rect: null,
+    pixelRect: null,
     isProportional: true,
     waypointsChart: [],
     scales: [],
   });
 
-  const { xZoom, yZoom, xOffset, yOffset, scrollTo, panning, isProportional } = state;
+  const {
+    xZoom,
+    yZoom,
+    timeOrigin,
+    spaceOrigin,
+    xOffset,
+    yOffset,
+    scrollTo,
+    panning,
+    zoomMode,
+    rect,
+    pixelRect,
+    isProportional,
+  } = state;
+
+  /** used when we change the train dataset for example to center the chart on the new data */
+  const setTimeOrigin = useCallback((newTimeOrigin: number) => {
+    setState((prev) => ({ ...prev, timeOrigin: newTimeOrigin }));
+  }, []);
 
   const paths = usePaths(projectPathTrainResult, selectedTrain);
+  const canvasDrawingHeight = height - FOOTER_HEIGHT; // 521
+  const drawingHeightWithoutTopPadding = canvasDrawingHeight - BASE_WAYPOINT_HEIGHT / 2; // 505
+  const drawingHeightWithoutBothPadding = canvasDrawingHeight - BASE_WAYPOINT_HEIGHT; // 489
+  const totalDistance = calcTotalDistance(waypoints);
+
+  const { minZoomMillimeterPerPx, maxZoomMillimeterPerPx } = getExtremaScales(
+    drawingHeightWithoutTopPadding,
+    drawingHeightWithoutBothPadding,
+    totalDistance
+  );
 
   const waypointsToDisplay = useMemo(
-    () => computeWaypointsToDisplay(waypoints, { height, isProportional, yZoom }),
-    [waypoints, height, isProportional, yZoom]
+    () =>
+      computeWaypointsToDisplay(
+        waypoints,
+        { height, isProportional, yZoom },
+        minZoomMillimeterPerPx,
+        maxZoomMillimeterPerPx
+      ),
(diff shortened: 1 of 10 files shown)


ui-charts/src/spaceTimeChart/index.ts
```diff
@@ -18,3 +18,6 @@ export * from './components/ConflictTooltip';
 export * from './components/OccupancyBlockLayer';
 export * from './components/WorkScheduleLayer';
 export * from './components/PatternRect';
+export * from './components/ZoomRect';
+export * from './components/TimeCaptions';
+export * from './stories/lib/components';
```
emersion, line 25: Hm, I don't think we should export anything from stories in the library. (All stories will soon be moved to `storybook/`.)

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -1,60 +1,140 @@
...
+  spaceTimeChartRef,
+  defaultTimeOrigin = 0,
+  defaultSpaceOrigin = 0,
+}: {
+  waypoints: Waypoint[];
+  projectPathTrainResult: ProjectPathTrainResult[];
+  manchetteWithSpaceTimeChartContainer: React.RefObject<HTMLDivElement>;
+  selectedTrain?: number;
+  height?: number;
+  spaceTimeChartRef?: React.RefObject<HTMLDivElement>;
```
kmer2016, line 76: If we rename `manchetteWithSpaceTimeChartRef` to `manchetteWithSpaceTimeChartContainer`, we might also consider renaming `spaceTimeChartRef` for consistency.

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -1,60 +1,140 @@
...
+    setState((prev) => ({ ...prev, timeOrigin: newTimeOrigin }));
+  }, []);
 
   const paths = usePaths(projectPathTrainResult, selectedTrain);
+  const canvasDrawingHeight = height - FOOTER_HEIGHT; // 521
+  const drawingHeightWithoutTopPadding = canvasDrawingHeight - BASE_WAYPOINT_HEIGHT / 2; // 505
+  const drawingHeightWithoutBothPadding = canvasDrawingHeight - BASE_WAYPOINT_HEIGHT; // 489
+  const totalDistance = calcTotalDistance(waypoints);
+
+  const { minZoomMillimetrePerPx, maxZoomMillimetrePerPx } = getExtremaScales(
```
kmer2016: small typo. ;)
```suggestion
  const { minZoomMillimeterPerPx, maxZoomMillimeterPerPx } = getExtremaScales(
```

anisometropie: metre is the standard spelling

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -1,60 +1,140 @@
...
+    yOffset,
+    scrollTo,
+    panning,
+    zoomMode,
+    rect,
+    pixelRect,
+    isProportional,
+  } = state;
+
+  const setTimeOrigin = useCallback((newTimeOrigin: number) => {
```
SharglutDev, line 114: Since this function isn't used in the story, it would be nice to have a comment explaining why we need it

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
SharglutDev, line 1: If I change the `DEFAULT_HEIGHT` to `761` for example (meaning in osrd, we resized the space time chart) and then use the rectangle zoom, the height of the space time chart becomes way more than 761px.

ui-charts/src/manchette/utils/helpers.ts
```diff
@@ -7,10 +7,76 @@ import {
...
+export const getExtremaScales = (
+  drawingHeightWithoutTopPadding: number,
+  drawingHeightWithoutBothPadding: number,
+  pathLengthMillimeter: number
+) => ({
+  minZoomMillimetrePerPx: pathLengthMillimeter / drawingHeightWithoutBothPadding,
+  maxZoomMillimetrePerPx: MAX_ZOOM_MANCHETTE_HEIGHT_MILLIMETER / drawingHeightWithoutTopPadding,
+});
+
+// export const getScaleFromRectangle = ()
```
SharglutDev:
```suggestion
```

ui-charts/src/manchette/utils/helpers.ts
```diff
@@ -67,30 +136,36 @@ export const computeWaypointsToDisplay = (
...
-            }px`,
+        height: `${Math.round(waypointHeight)}px`,
       },
     };
   });
 };
 
+/**
+ * 2 modes for space scales
+ * km (isProportional): { coefficient: gives a scale in metre/pixel }
```
SharglutDev:
```suggestion
 * km (isProportional): { coefficient: gives a scale in meter/pixel }
```

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -180,24 +376,71 @@ const useManchettesWithSpaceTimeChart = (
...
+            const spaceStart = clamp(initialData.position, minPoint.position, maxPoint.position);
+            const spaceEnd = clamp(data.position, minPoint.position, maxPoint.position);
+            const newRect = {
+              timeStart: new Date(timeStart),
+              timeEnd: new Date(timeEnd),
+              spaceStart,
+              spaceEnd,
+            };
+
+            let newPixelRect = null;
```
SharglutDev: For type safety

```suggestion
            const newRect: State['rect'] = {
              timeStart: new Date(timeStart),
              timeEnd: new Date(timeEnd),
              spaceStart,
              spaceEnd,
            };

            let newPixelRect: State['pixelRect'] = null;
```

ui-charts/src/manchette/utils/helpers.ts
anisometropie: metre is the standard spelling

SharglutDev: Oh yeah since it's a UK/US thing both are used. Fine by me :)

anisometropie: I think it’s even everyone except the US is using metre :)

emersion: We use American English everywhere in the codebase (e.g. "color" instead of "colour").

anisometropie: I would say, in this case it’s not really the US/UK thing we should consider, but something like what the International System of Units does (they also use metre). 



Are we officially using american English ? If anything we’re more likely  to be used by UK users than US users.

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
anisometropie: it was `manchetteWithSpaceTimeChartContainer` in this hook, I think Ref would be better for both

anisometropie: it’s called manchetteWithSpaceTimeChartRef and manchetteWithSpaceTimeChartContainer
I think we can use ref everywhere

anisometropie: done

ui-charts/src/spaceTimeChart/index.ts
anisometropie: i need to import MouseTracker for my story. should it be moved to stories as well?

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
anisometropie: done

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
anisometropie: can’t reproduce.

ui-charts/src/manchette/utils/helpers.ts
anisometropie: done

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
anisometropie: done

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
SharglutDev: Same

SharglutDev (approved):
Lgtm and tested for my part, great job :)

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
anisometropie: renamed to meter

ui-charts/src/manchette/utils/helpers.ts
anisometropie: renamed to meter everywhere

ui-charts/src/spaceTimeChart/index.ts
anisometropie: can’t do much now. MouseTracker is in spaceTimeChart, and I need it in /storybook/stories

I propose we leave this as is for now, MouseTracker gets moved in the right location with https://github.com/OpenRailAssociation/osrd-ui/pull/965. 
we can update the import then and remove the `export * from './stories/lib/components';`

emersion: Could we `import '../../ui-charts/src/spaceTimeChart/stories/lib/components'` in the story instead? I'd prefer that hacky import over polluting our public API.

anisometropie: yes, that’s actually better, done !

kmer2016 (approved):
Great !

ui-charts/src/spaceTimeChart/index.ts
emersion: Actually this ended up resulting in two `createContext()` calls for the same context in the final bundle, so my suggestion doesn't work.

PR #926, opened 2025-02-24
ui-spacetimechart: remove vitest dependency

emersion:
vitest is not a runtime dependency: it's a devDependency. Drop it since the root package.json already pulls it in.


SharglutDev (approved):
Lgtm and tested

jacomyal (approved):

jacomyal:
Same here, all good to me :+1:

PR #928, opened 2025-02-24
Remove dist/ JS imports 

emersion:
Instead, import from the package itself. That way we can ensure
we're only importing public API and we're not messing with
internal types.

Drop the eslint-disable import/no-unresolved comments, because
these errors aren't triggered anymore.

_See individual commits._

### DIFF
--- a/ui-spacetimechart/src/index.ts
+++ b/ui-spacetimechart/src/index.ts
@@ -1,6 +1,15 @@
 import './styles/main.css';
 
-export type { HoveredItem, SpaceTimeChartProps } from './lib/types';
+export type {
+  HoveredItem,
+  SpaceTimeChartProps,
+  SpaceScale,
+  OperationalPoint,
+  SpaceTimeChartTheme,
+  Point,
+  DataPoint,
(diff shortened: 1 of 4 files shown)


ui-spacetimechart/src/index.ts
```diff
@@ -1,6 +1,16 @@
...
+export type {
+  HoveredItem,
+  SpaceTimeChartProps,
+  SpaceScale,
+  OperationalPoint,
+  SpaceTimeChartTheme,
+  Point,
+  DataPoint,
+  SpaceTimeChartContextType,
+  HoveredItem,
```
jacomyal: `HoveredItem` is exported twice.

jacomyal (changes requested):

ui-spacetimechart/src/index.ts
emersion: Yup, CI complained about this. Should be fixed already!

jacomyal (approved):

jacomyal:
LGTM :+1:

kmer2016 (approved):

PR #930, opened 2025-02-24
ui-manchette-with-spacetimechart: add ManchetteWithSpaceTimeChart

emersion:
This is a component designed for simple use-cases. It's easier to
use than the hook. Users can still decide to use the hook instead
of this component, especially when they need more customizability.

Other components can be passed in for the header and SpaceTimeChart,
and additional properties can be passed to Manchette/SpaceTimeChart.

This is a breaking API change: the hook is no longer the default
export for ui-manchette-with-spacetimechart. Users need to replace:

    import useManchetteWithSpaceTimeChart from 'ui-manchette-with-spacetimechart';

with:

    import { useManchetteWithSpaceTimeChart } from 'ui-manchette-with-spacetimechart';

_See individual commits._

### DIFF
--- a/ui-manchette-with-spacetimechart/src/components/ManchetteWithSpaceTimeChart.tsx
+++ b/ui-manchette-with-spacetimechart/src/components/ManchetteWithSpaceTimeChart.tsx
@@ -1,0 +1,93 @@
+import React, { useRef } from 'react';
+
+import Manchette, {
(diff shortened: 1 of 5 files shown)


theocrsb (approved):
LGTM and tested

clarani (approved):
LGTM, tested :white_check_mark:

PR #938, opened 2025-02-25
ui-spacetimechart: picking layer downscaling

### LINKED ISSUE #915: space time chart: picking only covers a portion of the stage when devicePixelRatio < 1
### What happened?

https://github.com/user-attachments/assets/54f111d3-36c3-4327-bc42-f0a6d9af373e

When `devicePixelRatio < 1`, the picking layer only covers a related fraction of the stage. The video shows for instance that paths hovering only works in the top left quadrant of the stage, and does not work on the right or on the bottom.

Interactions should work similarly on the whole stage all the time.

1. Go to [this page](https://openrailassociation.github.io/osrd-ui/?path=/story/spacetimechart-paths-interactions--default-args)
2. Zoom out in your browser (with `Ctrl` + `-` for instance)
3. Try interacting with the paths in the bottom right of the stage

Local Storybook

Firefox

jacomyal:
This commit fixes #915, and also improves picking performances, by downscaling the picking layer.

Details:
- Cleans unused and poorly named "number" argument from drawAliasedLine
- Adds new get getPickingScalingRatio helper, that returns the picking downscaling ratio depending on the current devicePixelRatio value
- Fixes picking layers resize handling
- Adds a new optional "scalingRatio" argument to each canvas aliased drawing function
- Adds new scalingRatio argument to PickingDrawingFunction, and updates useCanvas accordingly
- Gives the scaling ratio from PickingDrawingFunction to aliased functions in every places where usePicking is called

### DIFF
--- a/ui-charts/src/spaceTimeChart/components/ConflictLayer.tsx
+++ b/ui-charts/src/spaceTimeChart/components/ConflictLayer.tsx
@@ -61,7 +61,7 @@ export const ConflictLayer = ({ conflicts }: ConflictLayerProps) => {
   useDraw('paths', drawConflictLayer);
 
   const drawPicking = useCallback<PickingDrawingFunction>(
-    (imageData, { registerPickingElement, getTimePixel, getSpacePixel }) => {
+    (imageData, { registerPickingElement, getTimePixel, getSpacePixel }, scalingRatio) => {
       for (const [conflictIndex, conflict] of conflicts.entries()) {
         const x = getTimePixel(conflict.timeStart);
         const y = getSpacePixel(conflict.spaceStart);
@@ -78,7 +78,8 @@ export const ConflictLayer = ({ conflicts }: ConflictLayerProps) => {
           { x: x - border, y: y - border },
           width + 2 * border,
           height + 2 * border,
-          color
+          color,
+          scalingRatio
         );
       }
     },
--- a/ui-charts/src/spaceTimeChart/components/PathLayer.tsx
+++ b/ui-charts/src/spaceTimeChart/components/PathLayer.tsx
@@ -451,7 +451,7 @@ export const PathLayer = ({
   useDraw('paths', drawAll);
 
   const drawPicking = useCallback<PickingDrawingFunction>(
-    (imageData, stcContext) => {
+    (imageData, stcContext, scalingRatio) => {
       const { registerPickingElement } = stcContext;
 
       // Draw segments:
@@ -472,7 +472,8 @@ export const PathLayer = ({
             point,
             lineColor,
             STYLES[level].width + pickingTolerance,
-            true
+            true,
+            scalingRatio
           );
         }
       });
@@ -491,7 +492,8 @@ export const PathLayer = ({
           point,
           (STYLES[level].width + pickingTolerance) * 2,
           lineColor,
-          false
+          false,
+          scalingRatio
         );
       });
     },
--- a/ui-charts/src/spaceTimeChart/hooks/useCanvas.ts
+++ b/ui-charts/src/spaceTimeChart/hooks/useCanvas.ts
@@ -18,6 +18,7 @@ import {
   type Point,
   type SpaceTimeChartContextType,
 } from '../lib/types';
+import { getPickingScalingRatio } from '../utils/canvas';
 import { colorToIndex, rgbToHex } from '../utils/colors';
 import getPNGBlob from '../utils/png';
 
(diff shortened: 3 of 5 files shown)


ui-spacetimechart/src/utils/canvas.ts
```diff
@@ -139,8 +174,13 @@ export function drawAliasedDisc(
...
+  drawOnBottom: boolean,
+  scalingRatio: number = 1
 ): void {
+  centerX = Math.round(centerX * scalingRatio);
+  centerY = Math.round(centerY * scalingRatio);
+  radius = Math.round(radius * scalingRatio);
+
   centerX = Math.round(centerX);
   centerY = Math.round(centerY);
   radius = Math.ceil(radius);
```
emersion: These seem unnecessary now?

ui-spacetimechart/src/utils/canvas.ts
```diff
@@ -2,54 +2,89 @@ import { clamp, identity } from 'lodash';
...
-    return drawAliasedLine(imageData, to, from, [r, g, b], thickness, drawOnBottom);
+  if (fromX > toX)
+    return drawAliasedLine(
+      imageData,
+      { x: toX, y: toY },
+      { x: fromX, y: fromY },
+      [r, g, b],
+      thickness,
+      drawOnBottom
+    );
```
emersion: We should probably pass down the `scalingRatio` here?

ui-spacetimechart/src/utils/canvas.ts
```diff
@@ -139,8 +174,13 @@ export function drawAliasedDisc(
   { x: centerX, y: centerY }: Point,
   radius: number,
   [r, g, b]: RGBColor | RGBAColor,
-  drawOnBottom: boolean
+  drawOnBottom: boolean,
+  scalingRatio: number = 1
 ): void {
+  centerX = Math.round(centerX * scalingRatio);
+  centerY = Math.round(centerY * scalingRatio);
+  radius = Math.round(radius * scalingRatio);
```
emersion: Old code was using `Math.ceil` here, do we want to retain the old behavior?

I don't have the full context, but maybe it's to avoid ending up with `radius = 0`?

ui-spacetimechart/src/utils/canvas.ts
jacomyal: Oh good catch, thanks! Fixed :+1:

ui-spacetimechart/src/utils/canvas.ts
jacomyal: Yes, the `Math.ceil` was here for this reason, and I kind of brainfarted here I think. Anyway, it's cleaner now.

ui-spacetimechart/src/utils/canvas.ts
jacomyal: I merged the two parts and restored the `Math.ceil` call.

emersion (approved):
LGTM!

ui-spacetimechart/src/utils/canvas.ts
```diff
@@ -2,54 +2,90 @@ import { clamp, identity } from 'lodash';
...
+
+  // When devicePixelRatio is over 1 (like for Retina displays), we downscale based on the "HTML
+  // pixels":
+  if (dpr > 1) return PICKING_DOWNSCALING_RATIO;
+
+  // When devicePixelRatio is under 1 (like when the user zooms out for instance), we downscale
+  // based on the actual "screen pixels" (to avoid having a too large scene to fill):
+  if (dpr < 1) return PICKING_DOWNSCALING_RATIO * dpr;
+
+  return PICKING_DOWNSCALING_RATIO;
```
clarani:
```suggestion
  // When devicePixelRatio is over 1 (like for Retina displays), we downscale based on the "HTML
  // pixels":
  if (dpr >= 1) return PICKING_DOWNSCALING_RATIO;

  // When devicePixelRatio is under or equal to 1 (like when the user zooms out for instance), we downscale
  // based on the actual "screen pixels" (to avoid having a too large scene to fill):
  return PICKING_DOWNSCALING_RATIO * dpr;
```

clarani (commented):
LGTM :white_check_mark: 

Only left 1 small comment

clarani (approved):

ui-spacetimechart/src/utils/canvas.ts
jacomyal: I just integrated that modification, thanks!

PR #940, opened 2025-02-25
ui-spacetimechart: fix op graduation display when hidden

### LINKED ISSUE #9970: GET horizontal lines aren't updated when changing manchette visible OPs
### What happened?

Starting from here:
Image

Editing the OP in the manchette doesn't update the horizontal lines, for example if i remove all the intermediate OPs the lines remain:
Image

I'm expecting the lines to be in sync with the manchette OPs at all times.

.

Recette (SNCF)

Google Chrome

94d3a84

Yohh:
add operationnalPoints in fingerprint to trigger changes in useCanvas() then redraw charts if one or many operationnalPoints is hidden by machette settings panel

to test it, go to work-schedules `spacetimechart`story, remove one or several OPs

close https://github.com/OpenRailAssociation/osrd/issues/9970

### DIFF
--- a/ui-spacetimechart/src/components/SpaceTimeChart.tsx
+++ b/ui-spacetimechart/src/components/SpaceTimeChart.tsx
@@ -63,6 +63,7 @@ export const SpaceTimeChart = (props: SpaceTimeChartProps) => {
   const fingerprint = useMemo(
     () =>
       JSON.stringify({
+        operationalPoints,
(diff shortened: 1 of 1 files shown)


RomainValls (approved):
LGTM and tested 👍

clarani (approved):
LGTM :white_check_mark:

PR #944, opened 2025-02-27
ui-core: fix ui-core/input with small props

### LINKED ISSUE #943: weird height in input with prop small
### What happened?

This [PR](https://github.com/OpenRailAssociation/osrd-ui/pull/924) introduced a small bug, the `small` class now appears 2 times in the element : 

Image

Also (was probably there before but we missed because we don't use any `Input` with `small` prop yet) the height of the input is fixed and doesn't match the one of the component.

- [ ] No double `small` class
- [ ] Height should fit the content even with `small` prop

Inspect an Input with `small: true` in public storybook

Deployed Storybook

Firefox

SharglutDev:
fix #943


kmer2016 (approved):
LGTM !

theocrsb (approved):

Uriel-Sautron (approved):
LGTM

PR #946, opened 2025-02-28
ui-warped-map: move stories to storybook

emersion:
Plus some other loosely related cleanup.

_See individual commits._


SharglutDev (approved):
Lgtm and tested, thank you.

Probably doesn't come from your PR, but warped map stories are very long to load imo.

RomainValls (approved):
LGTM ✅

PR #957, opened 2025-03-05
ui-icons: move stories to storybook

emersion:
This way we can use proper imports from '@osrd-project/ui-icons' instead of accessing directly internal files.


theocrsb (approved):

Math-R (approved):
lgtm

PR #962, opened 2025-03-06
ui-chart: spacetimechart: bordered path layer

### LINKED ISSUE #11056: ui-charts: update spaceTimeChart to support train service style
Add the possibility to put border around a PathLayer, in order to handle PacedTrain's occurrences [vizualisation](https://www.sketch.com/s/2fb064cd-c483-46ac-bd30-72705fd63e75/p/4D0E36FD-BB1D-4D86-A18A-2CAC08EBBADB/canvas#Inspect) in the spaceTimeChart

Math-R:
This PR allow the user to add a border around a PathLayer.
 it closes [#11056](https://github.com/OpenRailAssociation/osrd/issues/11056)

### DIFF
--- a/ui-charts/src/spaceTimeChart/components/PathLayer.tsx
+++ b/ui-charts/src/spaceTimeChart/components/PathLayer.tsx
@@ -53,6 +53,7 @@ type PathStyle = {
   opacity?: number;
   lineCap?: CanvasLineCap;
 };
+
 export type PathLevel = 1 | 2 | 3 | 4;
 const STYLES: Record<PathLevel, PathStyle> = {
   1: {
@@ -84,6 +85,12 @@ export type PathLayerProps = {
   color: string;
   pickingTolerance?: number;
   level?: PathLevel;
+  border?: {
+    offset: number;
+    color: string;
+    width?: number;
+    backgroundColor?: string;
+  };
 };
 
 /**
@@ -97,6 +104,7 @@ export const PathLayer = ({
   color,
   level = DEFAULT_LEVEL,
   pickingTolerance = DEFAULT_PICKING_TOLERANCE,
+  border,
 }: PathLayerProps) => {
   /**
    * This function returns the list of points to join to draw the path. It will be both used to
@@ -354,8 +362,43 @@ export const PathLayer = ({
     [path]
   );
 
+  const drawBorder = useCallback<DrawingFunction>(
+    (ctx, stcContext) => {
+      if (!border) return;
+      const borderWidth = border.width || 1;
+      const mainPathStyle = STYLES[level];
+      const totalPathWidth = border.offset * 2 + mainPathStyle.width;
+      const backgroundColor = border.backgroundColor || '#fff';
+      const segments = getPathSegments(stcContext);
+      ctx.save();
+      ctx.beginPath();
+      const drawSegments = (lineWidth: number, borderColor = border.color) => {
+        ctx.strokeStyle = borderColor;
+        ctx.lineWidth = lineWidth;
+        ctx.lineCap = 'round';
+        segments.forEach(({ x, y }, i) => {
+          if (x === segments[i - 1]?.x && y === segments[i - 1]?.y) return;
+          if (i === 0) {
+            ctx.moveTo(x, y);
+          } else {
+            ctx.lineTo(x, y);
+          }
+        });
+        ctx.stroke();
+      };
+
+      drawSegments(totalPathWidth + borderWidth * 2);
+      drawSegments(totalPathWidth, backgroundColor);
+
+      ctx.restore();
+    },
+    [border, getPathSegments, level]
+  );
+
   const drawAll = useCallback<DrawingFunction>(
     (ctx, stcContext) => {
+      drawBorder(ctx, stcContext);
+
       // Draw stops:
       ctx.strokeStyle = color;
       ctx.lineWidth = PAUSE_THICKNESS;
@@ -401,6 +444,7 @@ export const PathLayer = ({
       drawExtremities,
       computePathLength,
       drawLabel,
+      drawBorder,
       path.label,
     ]
   );
--- a/ui-charts/src/spaceTimeChart/stories/additional-data.stories.tsx
+++ b/ui-charts/src/spaceTimeChart/stories/additional-data.stories.tsx
@@ -36,6 +36,8 @@ const CLOSED_DOORS = [
   },
 ];
 
+const DEFAULT_PATH_LEVEL = 2;
+
 type MonoTrackSpaceProps = {
   from: number;
   to: number;
@@ -257,7 +259,13 @@ const Wrapper = ({ swapAxis, spaceScaleType }: WrapperProps) => {
         }}
       >
         {PATHS.map((path) => (
-          <PathLayer key={path.id} path={path} color={path.color} />
+          <PathLayer
+            key={path.id}
+            path={path}
+            color={path.color}
+            border={path.border}
+            level={path.level || DEFAULT_PATH_LEVEL}
+          />
         ))}
         {MONO_TRACK_SPACES.map(({ from, to }, i) => (
           <MonoTrackSpace key={i} from={from} to={to} />
--- a/ui-charts/src/spaceTimeChart/stories/base-rendering.stories.tsx
+++ b/ui-charts/src/spaceTimeChart/stories/base-rendering.stories.tsx
@@ -62,7 +62,13 @@ const Wrapper = ({
         yOffset={yOffset}
       >
         {paths.map((path) => (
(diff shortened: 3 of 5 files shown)


ui-charts/src/spaceTimeChart/components/PathLayer.tsx
```diff
@@ -336,8 +344,55 @@ export const PathLayer = ({
...
+        ctx.beginPath();
+        segments.forEach(({ x, y }, i) => {
+          if (i === 0) {
+            ctx.moveTo(x, y + offsetY);
+          } else if (i === segments.length - 1) {
+            ctx.lineTo(x, y + offsetY);
+          } else {
+            ctx.lineTo(x + offsetX, y + offsetY);
+          }
+        });
```
Yohh:
```suggestion
        segments.forEach(({ x, y }, i) => {
          if (x === segments[i - 1]?.x && y === segments[i - 1]?.y) return;
          if (i === 0) {
            ctx.moveTo(x, y + offsetY);
          } else if (i === segments.length - 1) {
            ctx.lineTo(x, y + offsetY);
          } else {
            ctx.lineTo(x + offsetX, y + offsetY);
          }
        });
```

ui-charts/src/spaceTimeChart/components/PathLayer.tsx
```diff
@@ -336,8 +344,55 @@ export const PathLayer = ({
     [path]
   );
 
+  const drawBorder = useCallback<DrawingFunction>(
+    (ctx, stcContext) => {
+      if (!border) return;
```
Yohh, line 367: you can destructure `border` here

ui-charts/src/spaceTimeChart/stories/additional-data.stories.tsx
```diff
@@ -257,7 +257,13 @@ const Wrapper = ({ swapAxis, spaceScaleType }: WrapperProps) => {
         }}
       >
         {PATHS.map((path) => (
-          <PathLayer key={path.id} path={path} color={path.color} />
+          <PathLayer
+            key={path.id}
+            path={path}
+            color={path.color}
+            border={path.border}
+            level={path.level || 2}
```
Yohh: magic number

ui-charts/src/spaceTimeChart/components/PathLayer.tsx
Uriel-Sautron: Done

ui-charts/src/spaceTimeChart/stories/additional-data.stories.tsx
Uriel-Sautron: Done

ui-charts/src/spaceTimeChart/components/PathLayer.tsx
Uriel-Sautron: Following our discussion, we feel that destructuring could be detrimental to readability.

ui-charts/src/spaceTimeChart/components/PathLayer.tsx
```diff
@@ -66,6 +67,13 @@ export type PathLayerProps = {
   color: string;
   pickingTolerance?: number;
   level?: PathLevel;
+  border?: {
+    offset: number;
+    level: PathLevel;
```
emersion: Seems like this is unused

ui-charts/src/spaceTimeChart/components/PathLayer.tsx
```diff
@@ -336,8 +345,48 @@ export const PathLayer = ({
...
+      }
+
+      ctx.restore();
+    },
+    [border, getPathSegments, level]
+  );
+
   const drawAll = useCallback<DrawingFunction>(
     (ctx, stcContext) => {
+      if (border !== undefined) {
```
emersion: Nit: no need for an `if` here, since we already have one inside the function.

ui-charts/src/spaceTimeChart/components/PathLayer.tsx
```diff
@@ -336,8 +345,48 @@ export const PathLayer = ({
...
+            ctx.lineTo(x, y);
+          }
+        });
+        ctx.stroke();
+      };
+
+      drawSegments(totalPathWidth + borderWidth * 2);
+
+      if (border.backgroundColor) {
+        drawSegments(totalPathWidth, border.backgroundColor);
```
emersion: This is a bit confusing: when there is no background color, I would've expected a thin `width`-sized border `offset` away from the path. However a thick path is painted instead:

out

ui-charts/src/spaceTimeChart/components/PathLayer.tsx
```diff
@@ -336,8 +345,48 @@ export const PathLayer = ({
...
+      const drawSegments = (lineWidth: number, borderColor = border.color) => {
+        ctx.strokeStyle = borderColor;
+        ctx.lineWidth = lineWidth;
+        ctx.beginPath();
+        segments.forEach(({ x, y }, i) => {
+          if (x === segments[i - 1]?.x && y === segments[i - 1]?.y) return;
+          if (i === 0) {
+            ctx.moveTo(x, y);
+          } else if (i === segments.length - 1) {
+            ctx.lineTo(x - border.offset / 2, y);
```
emersion: Why the `- border.offset / 2` here? It seems like the spacing on the right is smaller than `offset` with this adjustment.

ui-charts/src/spaceTimeChart/components/PathLayer.tsx
Math-R: thanks, it was a leftover from a previous approach

ui-charts/src/spaceTimeChart/components/PathLayer.tsx
Math-R: It was unnecessary, i wasn't seeing the end of the line well because of the display device so i tried unusefully to see something at the end i guess... so => removed ! thanks

ui-charts/src/spaceTimeChart/components/PathLayer.tsx
Math-R: True, thanks !

ui-charts/src/spaceTimeChart/components/PathLayer.tsx
emersion: (I suppose the right way to fix this would be to leave the space between the path and the border transparent, which would be tricky to do? Perhaps a workaround would be to use white.)

Math-R: Sorry i forgot to answer, but i fix this using globalCompositeOperation

ui-charts/src/spaceTimeChart/components/PathLayer.tsx
```diff
@@ -336,8 +344,44 @@ export const PathLayer = ({
     [path]
   );
 
+  const drawBorder = useCallback<DrawingFunction>(
+    (ctx, stcContext) => {
+      if (!border) return;
+      const borderWidth = border.width || 1;
+      const mainPathStyle = STYLES[level];
+      const totalPathWidth = border.offset * 2 + mainPathStyle.width;
+      ctx.globalCompositeOperation = border.backgroundColor ? 'source-over' : 'xor';
```
emersion: This works only when the paced train is the first one painted. When I move `getPaths()` for "Paced" to the end of the list in `PATHS`, I see this:

out

Probably a simple way to workaround this would be to:

- Set `strokeStyle = 'transparent'`
- Set `globalCompositeOperation = 'copy'` for the second `drawSegments()` call

Math-R: Tried but still a problem with another behaviour. I also tried to drawImage from another context but this was not at all optimized in terms of performance, plus this occure blurry effect or aliasing depending of the device pixel ratio.

As the mockup by @thibautsailly  use a background color between the path and the border, the fact that a path  with border hide a path behind seems to be the normal behaviour.
So i put a default backgroundColor to white.

After a discussion with @jacomyal about these different behaviors, it seems that the only possibility for the user to have a transparent background (using xor for performance) would be to draw a bordered path in another layer.
But this seems heavy in terms of implementation and harder for everyone to use it in a simple way.

So, what you think ?

emersion: Yeah, defaulting to solid white sounds completely reasonable to me! I'd be fine with making the field required as well, so that everything is more explicit (e.g. if the user uses a dark theme, white might be confusing).

emersion (approved):
LGTM :) 

Seems like some ESLint warnings need fixing, but should be good to go apart from that.

Yohh (approved):

PR #963, opened 2025-03-07
ui-charts: make picking elements modular

### LINKED ISSUE #958: ui-charts: make picking layers generic
### Description and goal

The objects taken in the picking layers should not be hardcoded

### Acceptance criteria

- [ ] the picking layers can be used by external users

emersion:
Currently, `PickingElement` is a closed type: only 3 types explicitly listed inside osrd-ui are accepted. External layers can't define their own picking elements. For instance, an external user with a custom work schedule layer cannot define a work schedule element to display a tooltip on hover.

Fix this by making `PickingElement` an open-ended type: it only has a `type` field, rest of the fields are layer-specific. Each layer defines the more specific type and provides a helper to convert to that more specific type.

(I've contemplated using generics instead, but they spread all over the place and make everything more complicated.)

_See individual commits._

Closes #958

### DIFF
--- a/ui-charts/src/spaceTimeChart/components/ConflictLayer.tsx
+++ b/ui-charts/src/spaceTimeChart/components/ConflictLayer.tsx
@@ -1,7 +1,7 @@
 import { useCallback } from 'react';
 
 import { useDraw, usePicking } from '../hooks/useCanvas';
-import { type DrawingFunction, type PickingDrawingFunction } from '../lib/types';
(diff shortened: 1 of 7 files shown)


clarani (approved):
LGTM :white_check_mark:

Synar (commented):
I'm not familiar with the component so I don't quite understand the use case, but the code LGTM

emersion:
We've discussed out-of-band about the use-case, and Alice said this makes sense to her :)

PR #964, opened 2025-03-07
ui-charts: simplify fillRect() with getPoint()

emersion:
Instead of manually handling timeAxis, use the getPoint() helper.

### DIFF
--- a/ui-charts/src/spaceTimeChart/utils/canvas.ts
+++ b/ui-charts/src/spaceTimeChart/utils/canvas.ts
@@ -389,27 +389,19 @@ export type CanvasRect = {
 
 export function fillRect(
   ctx: CanvasRenderingContext2D,
-  rect: CanvasRect,
-  spaceTimeContext: SpaceTimeChartContextType
+  { timeStart, timeEnd, spaceStart, spaceEnd }: CanvasRect,
+  { getPoint }: SpaceTimeChartContextType
 ) {
-  const { getTimePixel, getSpacePixel, timeAxis } = spaceTimeContext;
-  const { timeStart, timeEnd, spaceStart, spaceEnd } = rect;
+  const startPoint = getPoint({ time: Number(timeStart), position: spaceStart });
+  const endPoint = getPoint({ time: Number(timeEnd), position: spaceEnd });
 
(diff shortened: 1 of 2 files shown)


ui-charts/src/spaceTimeChart/utils/canvas.ts
```diff
@@ -348,27 +348,19 @@ export type CanvasRect = {
...
   ctx: CanvasRenderingContext2D,
-  rect: CanvasRect,
-  spaceTimeContext: SpaceTimeChartContextType
+  { timeStart, timeEnd, spaceStart, spaceEnd }: CanvasRect,
+  { getPoint }: SpaceTimeChartContextType
 ) {
-  const { getTimePixel, getSpacePixel, timeAxis } = spaceTimeContext;
-  const { timeStart, timeEnd, spaceStart, spaceEnd } = rect;
+  const topLeft = getPoint({ time: Number(timeStart), position: spaceStart });
+  const bottomRight = getPoint({ time: Number(timeEnd), position: spaceEnd });
```
Math-R: Nit : Maybe it's about preferences but i'm feeling a bit weird with those const names as i expect topSomething to be greater than bottomAnotherStuff but you end up by soustracting top to bottom so in my mind it will at first read return something negative.

maybe we could use 

```
const startPoint = getPoint({ time: Number(timeStart), position: spaceStart });
const endPoint = getPoint({ time: Number(timeEnd), position: spaceEnd });
```

Math-R (approved):
Looks fine to me !

ui-charts/src/spaceTimeChart/utils/canvas.ts
anisometropie: The rectangle can be drawn upside down by the user,  I also think `startPoint`/`endPoint`  would make more sense here.

emersion: Right… TBH, I don't find `startPoint` and `endPoint` to be any more descriptive than `pointA` and `pointB`, but I don't really care that much either way. Changed!

anisometropie (approved):

PR #965, opened 2025-03-07
storybook: move SpaceTimeChart stories into /storybook

clarani:


ui-charts/src/spaceTimeChart/lib/types.ts
```diff
@@ -82,6 +82,13 @@ export type OperationalPoint = {
   importanceLevel?: number; // Lower is better. If null, the point won't be displayed.
 };
 
+export type WorkSchedule = {
```
emersion: Personally I think `WorkScheduleLayer.tsx` would be a better home for this type (but feel free to dismiss this comment).

ui-charts/src/spaceTimeChart/index.ts
emersion, line 1: Hm, this mixes up together a lot of types from different charts. For instance, `DEFAULT_THEME` is specific to the space time chart but nothing indicates that in the name. If we ever want to export a default theme for the speed time chart or the track occupancy diagram, we'd get collisions.

Not sure how we want to address this… We could rename exports so that it's more explicit what chart each type is useful for, or we could have one entrypoint per chart (e.g. `import { DEFAULT_THEME } from '@osrd-project/ui-charts/spaceTimeChart';`).

Seems like this discussion is outside the scope of this PR anyways.

storybook/stories/ui-charts/spaceTimeChart/scroll-navigation.stories.tsx
```diff
@@ -1,16 +1,21 @@
...
 import { keyBy } from 'lodash';
 
-import { SpaceTimeChart, PathLayer, isPointPickingElement, isSegmentPickingElement } from '..';
-import { OPERATIONAL_POINTS, PATHS } from './lib/paths';
-import { X_ZOOM_LEVEL, Y_ZOOM_LEVEL, zoom } from './lib/utils';
-import { type Point } from '../lib/types';
-import { isPathOnScreen } from '../utils/geometry';
-import { getSpaceAtTime } from '../utils/scales';
-import { getDiff } from '../utils/vectors';
+
```
emersion: Nit: double newline here

emersion (commented):
Overall looks good!

ui-charts/src/spaceTimeChart/index.ts
clarani: Yes I agree that this organization is not perfect. We will have to improve it later :+1:

ui-charts/src/spaceTimeChart/lib/types.ts
clarani: You're right, it's done :)

emersion (approved):
Very nice :)

PR #976, opened 2025-03-11
front: synchro split between manchette and get

### LINKED ISSUE #955: storybook: create a new story with splitted GET and manchette
### Description and goal

Create a new story with a synchronized GET and Manchette to prepare the integration of a GOV in these components

### Acceptance criteria

- [x] the scroll must be synchronized between the manchette and the GET
- [x] the vertical zoom should be synchronized between the manchette and the GET

Akctarus:
closes #955


emersion:
> Updates useManchetteWithSpaceTimeChart and styles, so that all elements in manchette are positioned absolutely

What is the motivation for this?

jacomyal:
> > Updates useManchetteWithSpaceTimeChart and styles, so that all elements in manchette are positioned absolutely
> 
> What is the motivation for this?

Good question, that's actually the main issue we've had with this feature. The most simple case is the following:
- Imagine you have two split points exactly on two consecutive waypoints (it's the case in the "Split points" story).
- With the old system, there would be no `div` to insert between the two split sections, and therefore they would touch each other, while we expect to be space between them (as there is in the GET).
- With the new system, the position of each split section and each waypoint is simply positionned relatively to the top of their container, based on the same helpers we use in the GET.

It's also quite more future proof, as inserting elements in the manchette does not rely on the previous elements. And finally, with the old method, we were kind of accumulating errors.

theocrsb (approved):
LGTM and tested !

storybook/stories/ui-charts/manchetteWithSpaceTimeChart/split-simple.stories.tsx
```diff
@@ -0,0 +1,168 @@
...
+import '@osrd-project/ui-core/dist/theme.css';
+import type { Meta } from '@storybook/react';
+import { clamp } from 'lodash';
+
+import { SAMPLE_WAYPOINTS, SAMPLE_PATHS_DATA } from './assets/sampleData';
+import type { ManchetteWithSpaceTimeChartProps } from '../../../../ui-charts/src/manchette/components/ManchetteWithSpaceTimeChart';
+import { positionMmToKm } from '../../../../ui-charts/src/manchette/utils';
+import { PathLayer, SpaceTimeChart } from '../../../../ui-charts/src/spaceTimeChart';
+import { useDraw } from '../../../../ui-charts/src/spaceTimeChart/hooks/useCanvas';
+import { type DrawingFunction } from '../../../../ui-charts/src/spaceTimeChart/lib/types';
```
emersion: We should be able to use `@osrd-project/ui-charts` imports here instead.

ui-charts/src/manchette/components/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -31,19 +34,22 @@ const ManchetteWithSpaceTimeChart = ({
...
-  const { manchetteProps, spaceTimeChartProps, handleScroll } = useManchetteWithSpaceTimeChart({
-    waypoints,
-    projectPathTrainResult,
-    manchetteWithSpaceTimeChartRef,
-    selectedTrain,
-    height,
-    spaceTimeChartRef,
-    defaultTimeOrigin: Math.min(...projectPathTrainResult.map((p) => +p.departureTime)),
-  });
+  const { manchetteProps, spaceTimeChartProps, handleScroll, paths } =
```
emersion: Probably need to pass down `spaceTimeChartChildren` here?

Either that, or just remove the `splitPoints` prop, since this is meant as a simple component anyways.

emersion:
The Manchette story is broken now, because the absolute positioning is performed in `useManchetteWithSpaceTimeChart` instead of `Manchette`:

out

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -58,6 +70,14 @@ type State = {
   scales: SpaceScale[];
 };
 
+export type SplitPoint<Payload = unknown> = {
+  position: number;
+  size: number;
+  payload: Payload;
+  spaceTimeChartComponent: ComponentType<{ position: number; size: number; payload: Payload }>;
+  manchetteComponent: ComponentType<{ position: number; size: number; payload: Payload }>;
```
emersion: What's the point in passing a payload and components, instead of two React nodes?

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -58,6 +70,14 @@ type State = {
   scales: SpaceScale[];
 };
 
+export type SplitPoint<Payload = unknown> = {
+  position: number;
+  size: number;
```
emersion, line 70: Nit: would be nice to rename this to `height` to make it clearer it's the vertical size, and add a `// in pixels` comment.

jacomyal:
> The Manchette story is broken now, because the absolute positioning is performed in `useManchetteWithSpaceTimeChart` instead of `Manchette`:
> 
> out

Good catch, thanks! I just fixed the Manchette stories, by moving the `position: absolute` to the inline styles added from `useManchetteWithSpaceTimeChart`, next to the `top` and `height`.

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -362,161 +334,361 @@ const useManchetteWithSpaceTimeChart = ({
...
-    () => zoomValueToSpaceScale(minZoomMillimeterPerPx, maxZoomMillimeterPerPx, yZoom),
-    [yZoom, minZoomMillimeterPerPx, maxZoomMillimeterPerPx]
-  );
-  return useMemo(
-    () => ({
-      manchetteProps,
-      spaceTimeChartProps,
+      spaceTimeChartChildren: splitPoints.map((sp, i) =>
+        createElement(sp.spaceTimeChartComponent, {
+          key: i,
```
emersion: Would be nicer to add an `id`/`key` prop to `SplitPoint` to avoid using indexes as keys.

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -362,161 +334,361 @@ const useManchetteWithSpaceTimeChart = ({
...
-  const timeScale = useMemo(() => zoomValueToTimeScale(xZoom), [xZoom]);
-  const spaceScale = useMemo(
-    () => zoomValueToSpaceScale(minZoomMillimeterPerPx, maxZoomMillimeterPerPx, yZoom),
-    [yZoom, minZoomMillimeterPerPx, maxZoomMillimeterPerPx]
-  );
-  return useMemo(
-    () => ({
-      manchetteProps,
-      spaceTimeChartProps,
+      spaceTimeChartChildren: splitPoints.map((sp, i) =>
```
emersion: It's a bit weird to have `paths` returned in a field, and some children returned in another field, and then have the user handle these separately. Also some stories just ignore `spaceTimeChartChildren`.

What is the public API we want to expose here?

- Why do we want to expose `paths` and `spaceTimeChartChildren` separately? We could build path components in the hook and return these in `spaceTimeChartChildren`. Is it to leave more freedom to the user to allow them to customize the paths if they want to, and the relative ordering of each component?
- Do we want to promise that `spaceTimeChartChildren` is only ever used when `splitPoints` is used? If so, better at least rename it to something less generic, e.g. `spaceTimeChartSplitPoints`.
- Do we want to leave it up to the user to build the split point components? ie, return a `{ position, size, payload }[]`, the user is responsible for doing the `.map((point) => <MyComponent {...point}/>)` just like paths?

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -362,161 +334,361 @@ const useManchetteWithSpaceTimeChart = ({
...
+        }
+        case 'splitSection': {
+          const {
+            split: { position, size, payload },
+          } = content;
+
+          const top = getSpacePixel(position);
+          finalContents.push(
+            createElement(
+              'div',
```
emersion: Do we need `createElement()` here? We can at least avoid it for `<div>`, and we should be able to avoid it for `manchetteComponent` as well:

```tsx
const ManchetteComponent = content.split.manchetteComponent;

<ManchetteComponent ... />
```

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
```diff
@@ -362,161 +334,361 @@ const useManchetteWithSpaceTimeChart = ({
...
+              {
+                style: {
+                  top: `${top + WAYPOINTS_OFFSET}px`,
```
emersion: Why do we need to apply an offset here, but not for "regular" waypoints?

ui-charts/src/manchette/styles/manchette.css
```diff
@@ -77,8 +93,6 @@
   }
 
   .space-time-chart-container {
-    height: 561px;
```
emersion, line 80: Nice to be able to get rid of this hardcoded height :)

emersion:
> I just fixed the Manchette stories, by moving the position: absolute to the inline styles added from useManchetteWithSpaceTimeChart, next to the top and height.

Sounds good!

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
jacomyal: I kept `size` rather than `height` as I imagined we will allow swapping time and space axis at some point in the future. It's already implemented [in the GET](https://openrailassociation.github.io/osrd-ui/?path=/story/spacetimechart-split--default&args=swapAxis:!true) because it was in the requirements, but I haven't seen that feature in any OSRD designs yet...

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
jacomyal: Hum, I don't see why it would be better to force users to add artificial IDs, only for some technical issue within `useManchetteWithSpaceTimeChart`. Especially since the `"react/no-array-index-key"` ESLint rule [is disabled](https://github.com/OpenRailAssociation/osrd-ui/blob/502987e0c2f596c2b34020d1309140123a760a24/.eslintrc#L79).

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
emersion: Makes sense to me.

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
jacomyal: The reason is:
- Waypoints are positioned so that the line at "the middle" is aligned with the space pixel
- Split points are positioned so that their top is aligned with the space pixel

ui-charts/src/manchette/styles/manchette.css
jacomyal: There are still many hanging around though :sweat_smile:

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
emersion: I don't think the IDs would be artificial. They would e.g. be OP IDs and such. We already require IDs for other types of array elements, e.g. waypoints.

The rule is disabled, but that doesn't mean that using indexes as keys is a good practice.
See https://react.dev/learn/rendering-lists#why-does-react-need-keys

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
emersion: Ah right, thanks for the explanation!

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
jacomyal: I'll rewrite them and rename the file as `.tsx`, it'll look surely cleaner indeed.

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
jacomyal: Agreed. Incorporating `Paths` in children from `useManchetteWithSpaceTimeChart` makes sense to me, and the API will look much more consistent and friendly.

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
jacomyal: Fair enough, I'll add IDs.

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
jacomyal: The API was initially with nodes, but it felt weird to me (you also made a comment about it when you saw it, actually). I decided to rewrite it with components, but I admit it still feels a bit weird.

I can change it back, I just don't know exactly if there is a best practice between those two patterns.

ui-charts/src/manchette/components/ManchetteWithSpaceTimeChart.tsx
jacomyal: Yes, good catch, thanks.

storybook/stories/ui-charts/manchetteWithSpaceTimeChart/split-simple.stories.tsx
jacomyal: I tried, but when I import `useDraw`, `PathLayer` or `SpaceTimeChart` from `@osrd-project/ui-charts`, StoryBook fails weirdly... I'll investigate some more.

jacomyal: I just pushed a fix for this :+1:

ui-charts/src/manchette/components/ManchetteWithSpaceTimeChart.tsx
jacomyal: I just pushed it

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
jacomyal: I just pushed it

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
jacomyal: I just pushed it

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
jacomyal: I just pushed it

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
emersion: Hm, it looks like `paths` is indeed used to customize the style in OSRD, so not sure this approach is a good idea after all:
https://github.com/OpenRailAssociation/osrd/blob/294581dae6cf5efcd230e014b04b9f5bbf85e664/front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx#L477

Moreover, I'm not really sure why the hook should be dealing with paths in the first place. There is no need to have any knowledge about paths to synchronize the space time chart and the manchette.

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
emersion: My main beef with the node version was that it was using callbacks. Usually props contain already-built elements (e.g. `children`). How would this look?

```ts
export type SplitPoint = {
  position: number;
  size: number;
  spaceTimeChartElement: ReactNode;
  manchetteElement: ReactNode;
};
```

emersion: But, maybe `spaceTimeChartElement` is unnecessary here, maybe it can be dropped and left for the user to pass to the space time chart on their own.

jacomyal: I don't see any issue with the `ReactNode` version, I think it is better indeed. About the `spaceTimeChartElement` though, I like that everything split-points related is in the same spot. It makes it way easier to manage, I think.

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
jacomyal: OK, moving the paths out of the hook makes a lot of sense indeed. The issue I have, then, is that `spaceTimeChartChildren` only contains the `SplitPoint.spaceTimeChartComponent`. And I still think it is comfortable for devs to only have to manage split points in single `useManchetteWithSpaceTimeChart` argument.

@emersion I have a quite dirty solution, and I'd like your validation on it first :sweat_smile: What if I add a new `additionalChildren?: ReactNode | ReactNode[] | null` prop to `SpaceTimeChart`, so that it becomes possible to give it two different sets of children (one as TSX tree children, and one as regular prop)? In which case, `useManchetteWithSpaceTimeChart` could return the rendered split-section nodes in `spaceTimeChartProps`, and users would just have to add their own children, such as the paths. The API would look like:

```tsx
const { manchetteProps, spaceTimeChartProps, handleScroll } = useManchetteWithSpaceTimeChart({
  waypoints,
  manchetteWithSpaceTimeChartRef,
  height,
  spaceTimeChartRef,
  splitPoints,
  defaultTimeOrigin,
});

return (
  <div className="manchette-space-time-chart-wrapper">
    <div
      className="manchette flex"
      style={{ height: `${height}px` }}
      onScroll={handleScroll}
    >
      <Manchette {...manchetteProps} />
      <div className="space-time-chart-container w-full sticky">
        <SpaceTimeChart className="inset-0 absolute h-full" {...spaceTimeChartProps}>
          {paths.map((path) => (
            <PathLayer key={path.id} path={path} color={path.color} level={path.level} />
          ))}
        </SpaceTimeChart>
      </div>
    </div>
  </div>
);
```

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.tsx
```diff
@@ -0,0 +1,700 @@
...
+    (newXZoom: number, xPosition = (spaceTimeChartRef?.current?.offsetWidth || 0) / 2) => {
+      setState((prev) => ({
+        ...prev,
+        ...zoomX(prev.xZoom, prev.xOffset, newXZoom, xPosition),
+      }));
+    },
+    [spaceTimeChartRef]
+  );
+
+  const spaceScales = useMemo(() => {
```
anisometropie, line 344: That’s a lot of code to squeeze into one useMemo. can’t this be broken in smaller, tested functions ?

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
jacomyal: I just pushed a fix for this 👍

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.ts
jacomyal: I just pushed this implementation, and I think it does improve the API. And it's still possible to handle split sections rendering manually in the children (as the paths), which is also nice I think.

storybook/stories/ui-charts/manchetteWithSpaceTimeChart/base.stories.tsx
```diff
@@ -66,7 +63,7 @@ const ManchetteWithSpaceTimeWrapper = ({
 };
 
 const meta: Meta<typeof ManchetteWithSpaceTimeWrapper> = {
-  title: 'Manchette with SpaceTimeChart/rendering',
+  title: 'Manchette with SpaceTimeChart/useManchetteWithSpaceTimeChart',
```
anisometropie: could probably have a better name. what does this story do ?

storybook/stories/ui-charts/manchetteWithSpaceTimeChart/simple.stories.tsx
```diff
@@ -6,7 +6,7 @@ import type { Meta } from '@storybook/react';
 import { SAMPLE_WAYPOINTS, SAMPLE_PATHS_DATA } from './assets/sampleData';
 
 const meta: Meta<typeof ManchetteWithSpaceTimeChart> = {
-  title: 'Manchette with SpaceTimeChart/simple',
+  title: 'Manchette with SpaceTimeChart/ManchetteWithSpaceTimeChart',
```
anisometropie: we have a story ManchetteWithSpaceTimeChart and useManchetteWithSpaceTimeChart
do we need both ?
can we have a more explicit name here too ?

storybook/stories/ui-charts/manchetteWithSpaceTimeChart/rectangle-zoom.stories.tsx
```diff
@@ -69,11 +70,7 @@ const ManchetteWithSpaceTimeWrapper = ({
         onScroll={handleScroll}
       >
         <Manchette {...manchetteProps} />
-        <div
-          className="space-time-chart-container w-full sticky"
-          ref={spaceTimeChartRef}
-          style={{ bottom: 0, left: 0, top: 2, height: `${DEFAULT_HEIGHT - 6}px` }}
```
anisometropie, line 74: really nice not to have to give all these specific dimensions anymore, and to have all this mess simplified!

ui-charts/src/manchette/utils/__tests__/helpers.spec.ts
```diff
@@ -18,98 +18,58 @@ const mockedWaypoints = [
...
-      maxZoomMillimeterPerPx
-    );
+    const result = selectWaypointsToDisplay(mockedWaypoints, {
+      height: 100,
+      isProportional: false,
+      yZoom: 1,
+    });
     expect(result).toHaveLength(mockedWaypoints.length);
-    expect(result[0].styles?.height).toBe(`${BASE_WAYPOINT_HEIGHT}px`);
-    expect(result[1].styles?.height).toBe(`${BASE_WAYPOINT_HEIGHT}px`);
```
anisometropie, line 51: Would be nice not to lose these tests. from what I see height computation was moved to the manchette + spacetime hook, but it’s no longer covered by unit tests.

storybook/stories/ui-charts/manchetteWithSpaceTimeChart/split-simple.stories.tsx
```diff
@@ -0,0 +1,198 @@
...
+        payload: { name: SAMPLE_WAYPOINTS[2].name },
+        spaceTimeChartNode: <FlatStep position={SAMPLE_WAYPOINTS[2].position} />,
+        manchetteNode: (
+          <SplitElement
+            size={100}
+            position={SAMPLE_WAYPOINTS[2].position}
+            name={SAMPLE_WAYPOINTS[2].name}
+          />
+        ),
+      },
```
anisometropie: could we have some document on how to use split points ?
to explain what the payload does for example,
and why do we need to give both spaceTimeChartNode and a manchetteNode

storybook/stories/ui-charts/manchetteWithSpaceTimeChart/base.stories.tsx
jacomyal: It shows how to use the `useManchetteWithSpaceTimeChart` hook.

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.tsx
jacomyal: I don't really see the point, actually. It was more spread before, and it was way harder to me to understand what is related to the internal mechanics of this hook, and what was related to what it returns.

ui-charts/src/manchette/utils/__tests__/helpers.spec.ts
jacomyal: The height is no more relevant, as the elements are now positioned absolutely. Also, their new `top` positions are computed with the GET helpers, that are derived from `spaceScalesToBinaryTree` and `getNormalizedScaleAtPosition` that have enough unit tests, in my opinion at least.

storybook/stories/ui-charts/manchetteWithSpaceTimeChart/simple.stories.tsx
jacomyal: The two stories show how to use the two different APIs `ManchetteWithSpaceTimeChart` (the component) and  `ManchetteWithSpaceTimeChart` (the hook). I don't see the issues with these stories being there, nor with there names, actually.

It's still the same debate I think: The name of these stories is based on their code content, not on what they show on screen.

storybook/stories/ui-charts/manchetteWithSpaceTimeChart/split-simple.stories.tsx
jacomyal: Good point - and recent changes made some of that data obsolete. I'll push something ASAP :+1:

jacomyal: I just pushed:
- an introductory comment at the beginning of the story, to clarify the intent
- some comments in the `SplitPoint` declaration
- the suppression of the `payload` property, that is no more used since earlier today

storybook/stories/ui-charts/manchetteWithSpaceTimeChart/rectangle-zoom.stories.tsx
jacomyal: Agreed - and I hope this will make it easier to fix the remaining alignment issues in the future :crossed_fingers:

storybook/stories/ui-charts/manchetteWithSpaceTimeChart/simple.stories.tsx
anisometropie: hm, okay makes sense to show these two ways to use it. it’s just that the names aren’t really helpful. we already know it’s manchette + space time since it’s the name of the section.

Screenshot_20250326_203839

anisometropie: perhaps just name it `component api` and `hook api`.

jacomyal: OK, I like these better as well, good solution. I just pushed a fix.

storybook/stories/ui-charts/manchetteWithSpaceTimeChart/split-simple.stories.tsx
anisometropie: thanks !

anisometropie (approved):
Well done, thanks for the quality refacto

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.tsx
anisometropie: I guess my main concern is the lack of tests. but that’s an issue we have with most of our components. we can leave it for now

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.tsx
```diff
@@ -362,161 +325,345 @@ const useManchetteWithSpaceTimeChart = ({
...
+    baseScales.forEach((baseScale) => {
+      const coefficient =
+        'coefficient' in baseScale && typeof baseScale.coefficient === 'number'
+          ? baseScale.coefficient
+          : (baseScale.to - baseScale.from) / baseScale.size;
+
+      // Search for split point strictly BETWEEN from and to:
+      const splitPoint = splitPoints.find(
+        (point) => point.position > baseScale.from && point.position < baseScale.to
+      );
```
clarani: There might be several splitPoints in a single scale

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.tsx
```diff
@@ -362,161 +325,345 @@ const useManchetteWithSpaceTimeChart = ({
...
+        allScales.push({
+          to: baseScale.to,
+          coefficient,
+        });
+      } else {
+        allScales.push({
+          to: baseScale.to,
+          coefficient,
+        });
+      }
```
clarani:
```suggestion
      }
      
      allScales.push({
        to: baseScale.to,
        coefficient,
      });
```

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.tsx
```diff
@@ -362,161 +325,345 @@ const useManchetteWithSpaceTimeChart = ({
...
+        case 'splitSection': {
+          const {
+            split: { position },
+          } = content;
+
+          const top = getSpacePixel(position);
+          finalContents.push(
+            <div
+              style={{
+                top: `${top + WAYPOINTS_OFFSET}px`,
```
clarani, line 495: Can you add a small comment about `WAYPOINTS_OFFSET` ? I asked myself the same question as Simon when I read the code.

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.tsx
```diff
@@ -362,161 +325,345 @@ const useManchetteWithSpaceTimeChart = ({
...
       zoomMode,
       rect,
-      spaceScale,
-      timeScale,
-      toggleZoomMode,
+      minZoomMillimeterPerPx,
+      maxZoomMillimeterPerPx,
       setTimeOrigin,
+      isShiftPressed,
+      state.zoomMode,
```
clarani: zoomMode is already in the list of dependencies

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.tsx
jacomyal: Good catch, thanks! I just pushed a fix.

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.tsx
jacomyal: The fix for the other discussion simplifies the code and fixes this as well.

clarani (approved):
LGTM ! You can close my last comments once you have fixed them :+1:

anisometropie:
Sorry I pushed to the wrong branch, I just removed my commit

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.tsx
jacomyal: I just pushed a comment for this

ui-charts/src/manchette/hooks/useManchetteWithSpaceTimeChart.tsx
jacomyal: Fixed, thanks!

PR #977, opened 2025-03-12
ui-charts: add quadrilateral layer and associated story

RostomBoudi:


ui-charts/src/spaceTimeChart/components/Parallelogram.tsx
```diff
@@ -0,0 +1,54 @@
+import { useCallback } from 'react';
+
+import { useDraw } from '../hooks/useCanvas';
+import { type DataPoint, type DrawingFunction } from '../lib/types';
+
+export type ParallelogramProps = {
+  spaceTimePoints: [DataPoint, DataPoint, DataPoint, DataPoint];
```
anisometropie: if we can use any arbitrary 4 points on the plane, we can technically draw any quadrilateral. 
I think `Quadrilateral` would be a better name for that component

Screenshot_20250314_143740

ui-charts/src/spaceTimeChart/components/Parallelogram.tsx
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Draws a parallelogram in the space time chart
+ * Position correspond to each point of the parallelogram
+ *
+ *      spaceTimePoints[0]   ________ spaceTimePoints[1]
+ *                         /        /
+ *     spaceTimePoints[3] /________/ spaceTimePoints[2]
+ *
+ * @param {ParallelogramProps} props
+ * @returns {null}
```
anisometropie: already given by TypeScript annotations
```suggestion
```

ui-charts/src/spaceTimeChart/components/Parallelogram.tsx
anisometropie:
```suggestion
  vertices: [DataPoint, DataPoint, DataPoint, DataPoint];
```

RostomBoudi: Indeed, renaming done (in the component, props and associated story)

ui-charts/src/spaceTimeChart/components/Parallelogram.tsx
RostomBoudi: Habits die hard ^^'
Done 👌

ui-charts/src/spaceTimeChart/components/Parallelogram.tsx
RostomBoudi: Renaming done

anisometropie: thanks!

ui-charts/src/spaceTimeChart/components/Quadrilateral.tsx
```diff
@@ -0,0 +1,52 @@
...
+      ctx.lineWidth = style.borderWidth ?? 1;
+
+      ctx.beginPath();
+      vertices.forEach((dataPoint, index) => {
+        if (index === 0) {
+          ctx.moveTo(getTimePixel(dataPoint.time), getSpacePixel(dataPoint.position));
+        } else {
+          ctx.lineTo(getTimePixel(dataPoint.time), getSpacePixel(dataPoint.position));
+        }
+      });
```
Yohh:
```suggestion
      vertices.forEach((dataPoint) => {
          ctx.lineTo(getTimePixel(dataPoint.time), getSpacePixel(dataPoint.position));
      });
```

nit: you draw only one shape, no need to seek for first index here

Yohh (commented):
nice! thanks for this great first PR

Yohh:
can you reword your commit?

image

anisometropie (approved):
Thanks! looking good, works well.

gratz on your first contribution, nice to see more people who want to bring the project to life!

ui-charts/src/spaceTimeChart/components/Quadrilateral.tsx
emersion: Have you tested this? I think we need to move the canvas' current point, before we can start drawing lines?

anisometropie: Can’t this cause unexected behavior ?

emersion:
It seems this needs to be rebased on top of the latest `dev` commit, to resolve conflicts in `.gitignore`.

ui-charts/src/spaceTimeChart/components/Quadrilateral.tsx
RostomBoudi: I also though it was mandatory 🤔, but I just tested it and it seems to work fine

Yohh: > Have you tested this? I think we need to move the canvas' current point, before we can start drawing lines?

indeed
it seems moveTo and lineTo are the same, expect that lineTo draws, if it's the first point nothing will be draw

emersion: Right, just read the spec and it seems like `lineTo()` works just like `moveTo()` when there is no sub-path: https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-lineto

Math-R (approved):
Looks good to me ! thanks for this first cool contribution !

PR #978, opened 2025-03-13
storybook: move speed space chart stories

clarani:


Math-R (approved):
Lgtm ! seemsc that @Uriel-Sautron got some eslint warning with the import from core but don't have them on my side

Uriel-Sautron (approved):
Thank you

PR #987, opened 2025-03-20
fix a few imports

anisometropie:


storybook/stories/ui-charts/manchetteWithSpaceTimeChart/rectangle-zoom.stories.tsx
```diff
@@ -19,6 +19,7 @@ import { MouseTracker } from '../spaceTimeChart/helpers/components';
 
 import '@osrd-project/ui-core/dist/theme.css';
 import '@osrd-project/ui-charts/dist/theme.css';
+import '../../../styles/ManchetteWithSpaceTimeChart/rectangle-zoom.css'
```
emersion: This breaks the package isolation (external users won't be able to perform this import), and I believe this style should be included in `@osrd-project/ui-charts/dist/theme.css`.

emersion: Or is this style specific to stories? If so, it should be moved somewhere in `storybook/stories/ui-charts/`.

ui-charts/src/speedSpaceChart/types.ts
```diff
@@ -1,21 +1,21 @@
-type ElectricalProfileValues = {
+export type ElectricalProfileValues = {
```
emersion, line 1: I'm not sure I understand why this is required. This type is only used from `types.ts`, and only `Data` is exported in our public API.

If we need this type in OSRD, we also need to add it in `index.ts` exports.

storybook/stories/ui-charts/manchetteWithSpaceTimeChart/rectangle-zoom.stories.tsx
anisometropie: this style is only used by the story, should it really belong to something we export to our users ?

anisometropie: yes, it’s specific to the stories. I moved it

ui-charts/src/speedSpaceChart/types.ts
anisometropie: yes for OSRD, added it in speedSpaceChart/index.ts

ui-charts/src/speedSpaceChart/index.ts
```diff
@@ -6,3 +6,4 @@ export {
   type SpeedSpaceChartProps,
 } from './components/SpeedSpaceChart';
 export type { Data as SpeedSpaceChartData } from './types';
+export * from './types';
```
emersion: I think this is a bit too wide of an export: this will export as public API a whole bunch of our internal types, such as `Store`.

storybook/stories/ui-charts/spaceTimeChart/rectangle-zoom.stories.tsx
```diff
@@ -15,6 +15,8 @@ import { clamp } from 'lodash';
 
 import '@osrd-project/ui-core/dist/theme.css';
 import '@osrd-project/ui-charts/dist/theme.css';
+import './styles/rectangle-zoom.css';
+
```
emersion, line 19: Nit: extra newline here

anisometropie: done

ui-charts/src/speedSpaceChart/index.ts
anisometropie: ok, I restricted to only those used by osrd front

emersion (approved):
LGTM, thank you!

PR #1001, opened 2025-03-27
ui-charts: modify usepath hook params type

Math-R:
-remove selectedtrain props as the behaviour of changing string color must be must be managed according to user needs

-remove level from use path as it also must be managed according to user needs

### DIFF
--- a/ui-charts/src/manchette/hooks/usePaths.ts
+++ b/ui-charts/src/manchette/hooks/usePaths.ts
@@ -1,6 +1,5 @@
 import { useMemo } from 'react';
 
-import { type PathLevel } from '../../spaceTimeChart';
 import { PATH_COLOR_DEFAULT } from '../consts';
 import { type ProjectPathTrainResult } from '../types';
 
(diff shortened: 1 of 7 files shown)


Yohh (approved):

ui-charts/src/manchette/hooks/usePaths.ts
```diff
@@ -10,25 +9,23 @@ const transformCurve = (curve: ProjectPathTrainResult['spaceTimeCurves'][0], dep
...
           id: string;
           label: string;
-          color: string;
-          level: PathLevel;
           points: { time: number; position: number }[];
-        }>((spaceTimeCurve, ind) => ({
-          id: `${path.id}-${ind}`,
+          color: string;
+        }>((spaceTimeCurve) => ({
+          id: `${path.id}`,
```
SharglutDev:
```suggestion
          id: path.id,
```

SharglutDev (approved):
Lgtm and tested, you can close the comment after fixing it :)

ui-charts/src/manchette/hooks/usePaths.ts
Math-R: done thx !

Issue #901, opened 2025-02-13
ui-core: remove input-wrapper in inputs


clarani:
### What happened?

Inputs have wrapper of 3px, which complicates their integration. The padding should be on the label, not on the input itself.

![Image](https://github.com/user-attachments/assets/433b9351-36f6-42d9-adc3-edcad0863d5a)

It creates some component mismatch (here the select is bigger than the combo box for instance)
![Image](https://github.com/user-attachments/assets/69e8dde7-a366-4b9e-8839-e263f6344b86)

### How can we reproduce it (as minimally and precisely as possible)?

1. Open STDCM view on osrd

### On which environments the bug occurs?

Local Storybook

### On which browser the bug occurs?

Firefox

anisometropie:
if the component really needs a wrapper it has to be 0px

PR #1013, opened 2025-04-01
ui-inputs: remove input  wrapper padding

ghost:
Resolves this [issue](https://github.com/OpenRailAssociation/osrd-ui/issues/901).

@clarani Why do is the css written in separate files? I see tailwind is installed in the project. May I open a pr where I apply the appropriate tailwind classes [here](https://github.com/OpenRailAssociation/osrd-ui/blob/dev/ui-core/src/components/inputs/FieldWrapper.tsx#L51)?

### DIFF
--- a/ui-core/src/styles/inputs/fieldWrapper.css
+++ b/ui-core/src/styles/inputs/fieldWrapper.css
@@ -1,5 +1,4 @@
 .feed-back {
-  --input-wrapper-padding: 3px;
   --field-wrapper-padding-bottom: 16px;
   --status-message-wrapper-tooltip--offset: 4px;
 


SharglutDev:
Thank you for contributing turnerian2004 ! 
I think the related issue is no longer present. A few month ago, we added the prop `narrow` for some ui-core components (allowing us to remove all padding and margins around) and this class seems to have been removed at that time because it's no where to be found. @clarani could you try to reproduce your issue ?

Moreover, the fix you're proposing removes a css variable which is not related to the mentioned class. It is used in the file `statusMessage.css` l.40 for a component which needs it.

Regarding your tailwind question, even though tailwind is a dependency in this repo, we decided to use it only where the element doesn't need additional styling (like only for a flex align-center for example) to reduce the size of tsx file and improve their readability.

ghost:
> Thank you for contributing turnerian2004 ! I think the related issue is no longer present. A few month ago, we added the prop `narrow` for some ui-core components (allowing us to remove all padding and margins around) and this class seems to have been removed at that time because it's no where to be found. @clarani could you try to reproduce your issue ?
> 
> Moreover, the fix you're proposing removes a css variable which is not related to the mentioned class. It is used in the file `statusMessage.css` l.40 for a component which needs it.
> 
> Regarding your tailwind question, even though tailwind is a dependency in this repo, we decided to use it only where the element doesn't need additional styling (like only for a flex align-center for example) to reduce the size of tsx file and improve their readability.

Thanks for the answers @SharglutDev :) Much appreciated!!! 

I"ll close the pr now, since this pr seems no longer needed :)

PR #1014, opened 2025-04-02
ui-core: add tailwind prettier auto sorting

ghost:
Install tailwind prettier auto sorting classes feature
https://tailwindcss.com/blog/automatic-class-sorting-with-prettier


SharglutDev:
Thank you for proposing this PR @turnerian2004. After some discussions, we decided that we won't need this plugin.

SharglutDev:
To be more precise, we feel like the sorting won't do such a big difference for classes readability. We were also afraid to have sometimes merge conflicts because of that. Thank you for your comprehension.

ghost:
> To be more precise, we feel like the sorting won't do such a big difference for classes readability. We were also afraid to have sometimes merge conflicts because of that. Thank you for your comprehension.

@SharglutDev Thank you for the feedback & explanation. Merge conflicts are no fun, so best to avoid them :)

PR #1016, opened 2025-04-03
front: disable tailwind preflight from ui-core and ui-chart

### LINKED ISSUE #988: osrd-ui stylesheets leak outside of osrd-ui components
Rules like the following are included in our dist CSS files:

```css
*,
::before,
::after {
  box-sizing: border-box; /* 1 */
  border-width: 0; /* 2 */
  border-style: solid; /* 2 */
  border-color: #e5e7eb; /* 2 */
}
```

@RostomBoudi suggested either nesting all rules under osrd-ui specific classes, or disabling [TailwindCSS preflight rules](https://tailwindcss.com/docs/preflight).

RostomBoudi:
Disable tailwind preflight to prevent css from spilling over.

A lot of preflight apply css to everything (*), which may break application using osrd-ui.

`@import tailwindcss/base` should also be removed as it contain preflight and is redondant (as tailwindcss/components and tailwindcss/utilities are already imported).

Closes: https://github.com/OpenRailAssociation/osrd-ui/issues/988

### DIFF
--- a/storybook/tailwind.config.js
+++ b/storybook/tailwind.config.js
@@ -3,4 +3,7 @@ import osrdUiPreset from '../tailwind-preset.js';
 export default {
   presets: [osrdUiPreset],
   content: ['./stories/**/*.{js,jsx,ts,tsx}'],
+  corePlugins: {
+    preflight: false,
+  },
 };
--- a/ui-charts/src/manchette/styles/main.css
+++ b/ui-charts/src/manchette/styles/main.css
@@ -1,4 +1,3 @@
-@import 'tailwindcss/base';
 @import 'tailwindcss/components';
 @import 'tailwindcss/utilities';
 @import 'manchette.css';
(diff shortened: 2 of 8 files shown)


emersion:
I can confirm that the leaking styles are gone now. Unfortunately it seems like various components are broken: the button no longer has a hand cursor on hover, the manchette is missing borders, and so on.

RostomBoudi:
> I can confirm that the leaking styles are gone now. Unfortunately it seems like various components are broken: the button no longer has a hand cursor on hover, the manchette is missing borders, and so on.

Darn ! Sorry, my bad. 
I will try to find something else 🥲

emersion:
Hi! osrd-ui has been migrated to the main OSRD repository, in the `front/ui/` directory. If you want to continue working on this PR, please migrate it with the following steps: https://github.com/OpenRailAssociation/osrd/pull/10711#migrate-pr

Sorry for the inconvenience!

PR #1017, opened 2025-04-04
storybook: move over ui-core stories 

emersion:
_See individual commits._

These are the last stories that need moving.


clarani (approved):
LGTM :white_check_mark:

PR #1029, opened 2025-04-09
build: drop terser

emersion:
terser minifies the final JavaScript file. This isn't super useful because users have their own build system and minifier, and makes debugging more difficult because other build systems are in general unable to read source maps as input (thus stack traces are obfuscated).


kmer2016 (approved):

anisometropie (approved):
big yes

PR #1048, opened 2025-04-23
ui-charts: fix spacetimechart crashing when width goes to 0

### LINKED ISSUE #11385: scenario page crashes when too small
### What happened?

We get the following error when changing window size : `Failed to execute 'getImageData' on 'CanvasRenderingContext2D': The source width is 0.`

Image

The scenario page should not crash

Go to scenario page
Add 1 train
Open the console and stretch it until the bug appears

Local

Firefox

development

Synar:
Close https://github.com/OpenRailAssociation/osrd/issues/11385

The issue occurs when the width naturally goes to 1 due to the width of the page being too small, but with a ratio lower than 1, leading to an invalid value for the final width of the canvas. (Perhaps instead we could stop rendering the component entirely in such a case? It's not a very relevant case though.)

### DIFF
--- a/ui-charts/src/spaceTimeChart/hooks/useCanvas.ts
+++ b/ui-charts/src/spaceTimeChart/hooks/useCanvas.ts
@@ -241,8 +241,8 @@ export function useCanvas(
 
         canvas.style.width = size.width + 'px';
         canvas.style.height = size.height + 'px';
-        canvas.setAttribute('width', size.width * ratio + 'px');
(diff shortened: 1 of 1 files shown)


SharglutDev:
nit for the commit : `get` is a french acronym :)

jacomyal (approved):
Good catch, LGTM :+1:

kmer2016 (approved):
LGTM !

Synar:
> nit for the commit : `get` is a french acronym :)

Fixed =p

